PR #25, opened 2016-04-11
.

zde:
.


zde:
Not exactly sure why are you ignoring this merge request. Perhaps breaking any python application that decides to import cmd2 package while there are >64 kilobytes in a clipboard is on purpose?

FedericoCeratto:
Thank you for the PR. A description of the bug and the proposed solution would be more helpful than ".". I'll try to get it merged.

tleonhardt:
@zde Thanks for brining this issue to our attention.  It has been fixed in the most recent commit.

PR #28, opened 2016-12-09
Loading in a script file now works in Python 3.5

tleonhardt:
This code fixes a bug where you weren't able to load in a script file on Python 3.  I tested it on both Python 3.5 and 2.7 and it worked fine on both.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,5 @@
+__pycache__
+build
+dist
+cmd2.egg-info
+
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,6 @@ __pycache__
 build
 dist
 cmd2.egg-info
+.idea
+.cache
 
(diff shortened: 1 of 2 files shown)


tleonhardt:
At some point a more extensive refactoring should occur to write the code in a fashion which is more intrinsically compatible between Python 2 and Python 3 (making use of the six module and other techniques).  I've I get some free time maybe I'll take a stab at it.  But this quick hack fixes a major bug that prevented me from using this with Python 3.

FedericoCeratto:
Thanks for the PR. Would it be safer to try: file     except: import io  \n file = io.TextIOWrapper   ?

tleonhardt:
@FedericoCeratto 
My pleasure with the PR.  Thanks for helping maintain a really excellent tool which makes it incredibly simple to create a feature-rich shell program.  I like your suggestion and believe it would be safer.  I pushed an update.

I have a much larger PR in the works where I have done some extensive refactoring and created a unified Python 2/3 codebase using the six module and have removed the setuptools dependency on 2to3 (but added a runtime dependency on six).  But I want to polish it up a little more before sending it your way.

PR #29, opened 2016-12-12
Refactored to create a unified Python 2/3 codebase based on six which doesn't require 2to3

tleonhardt:
I refactored the existing code to create a unified Python 2/3 codebase based on six which doesn't require usage of the 2to3 during build, but does add a runtime dependency on the six module.

I also removed a couple apparently unused or irrelevant files.  

In the process my IDE made a bunch of whitespace changes to match the PEP 8 style guide.


tleonhardt:
@FedericoCeratto Let me know if there are any changes you would like for me to make to this Pull Request. 

The version number didn't necessarily need to change, I just did that since it was a significant change.

PR #30, opened 2017-01-11
Fixed example.py and exampleSession.txt as well as README.rst

tleonhardt:
Fixed example.py and exampleSession.txt so that the example runs and the test passes when running "python example.py -t exampleSession.txt".

Also made the front-page documentation consistent so it references the actual code and text in this modified example.


tleonhardt:
PR #31 now effectively includes a newer version the code changes present here.  Should this one just be closed?

PR #32, opened 2017-01-26
Added testing on Mac OS X with Python 2.7 and for Python 3.6 on Linux and fixed Windows AppVeyor tests.

tleonhardt:
Attempting to get Mac OS X testing to work with Travis C for Python 2.7 and 3.6.  Also adding Python 3.6 to the Linux tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,4 @@
 language: python
-os:
-  - linux
-  - osx
 sudo: false  # enable container-based build
 install:
   - pip install tox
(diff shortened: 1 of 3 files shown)


tleonhardt:
This PR will likely be open for awhile.  I'm using it to experiment with getting the Mac OS X build working.  It isn't ready yet.

tleonhardt:
Testing should now be working fine on Mac OS X with Python 2.7 and also on Linux with Python 3.6 in additional to all of the previous tests on Linux.

Getting Python 3.x testing working on Mac OS X with Travis CI proved more difficult than I thought.  So I'll try to do that later with a separate pull request.

One "trick" to get Mac OS X testing working at all is to realize that Python isn't officially supported as a language for OS X on Travis CI.  So you need to set the language to either "generic" or "objective-c".  Then it will let you run whatever you want.

The default Mac OS X image comes with Python 2.7 out of the box.  To test with any other version of Python on Mac OS X, we probably need to manually create a virtualenv and manually install the version we want to test with during either the "install" or "before_install" stage.

tleonhardt:
The Windows tests on AppVeyor are also now passing, thanks to the previous unit fix to the base_show unit test which fixed a bug in the unit test itself when run on Windows.

PR #34, opened 2017-01-30
Unit test improvements

tleonhardt:
Added some more unit tests in order begin increasing code coverage


tleonhardt:
Testing on my system shows a code coverage of 64% now.

PR #38, opened 2017-02-06
Remove command aliases

tleonhardt:
I documented previously undocumented commands and removed redundant command aliases.

This cleans up the out-of-the-box help menu significantly.  But short-form versions of commands are still available by default as long as self.abbrev is True (which is the default).

All of the unit tests have been updated to reflect these changes and the GitHub front-page documentation has been as well.

Unrelated, but put it in here anyways:  Updated AppVeyor badge to display build status for master branch instead of the latest one.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -6,22 +6,18 @@ cmd2
    :target: https://travis-ci.org/python-cmd2/cmd2
    :alt: Build status
 
-.. image:: https://ci.appveyor.com/api/projects/status/github/python-cmd2/cmd2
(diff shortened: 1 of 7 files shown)


tleonhardt:
@FedericoCeratto This seems like a really good change to me.  But I believe you probably have much better insight into what would be considered breaking backwards compatibility or not.  Feature-wise it it doesn't break anything.  But I was just hoping to get your buy-in on this one to make sure it won't cause you or the OpenStack folks any problems.

FedericoCeratto:
Thanks!

PR #39, opened 2017-02-07
Parsing improvements for commands using the options decorator

tleonhardt:
Improved how commands using the @options decorator are parsed to handle some difficult edge cases related to quotes and escape characters.

Added an "autorun_on_edit" settable parameter which defaults to True to preserve current behavior, but can be set to False to disable the automatic running of an edited file as a script when the editor closes.

Improved how piping works so it can deal with commands which output a large quantity of data.

Fixed where preloop() and postloop() get called so that do_load() for loading a script doesn't trigger them.

### DIFF
--- a/cmd2.py
+++ b/cmd2.py
@@ -1055,11 +1055,8 @@ class Cmd(cmd.Cmd):
         off the received input, and dispatch to action methods, passing them
         the remainder of the line as argument.
         """
-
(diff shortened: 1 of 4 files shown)


tleonhardt:
@FedericoCeratto I've tested this and it works well.  I just wanted you to be aware in case you had any suggestions for improvement.

PR #108, opened 2017-05-30
Remove parseline

catherinedevlin:
Nothing is actually using this method!  I think it was a fossil from very early work.

### DIFF
--- a/cmd2.py
+++ b/cmd2.py
@@ -548,7 +548,8 @@ def replace_with_file_contents(fname):
     """
     if fname:
         try:
-            result = open(os.path.expanduser(fname[0])).read()
+            with open(os.path.expanduser(fname[0])) as source_file:
+                result = source_file.read()
         except IOError:
             result = '< %s' % fname[0]  # wasn't a file after all
     else:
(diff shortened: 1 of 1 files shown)


cmd2.py
```diff
@@ -880,33 +881,6 @@ def postparsing_postcmd(self, stop):
         """
         return stop
 
-    def parseline(self, line):
```
tleonhardt: @catherinedevlin 
This is an override of a a method defined in cmd.  It is actually used within the parent cmd module in the complete() method.

Believe it or not it wasn't around from before, I just added this override a couple weeks ago.  

cmd2 needed to override it because cmd2 has more command shortcuts than cmd.

Maybe we should add something to the docstring to better explain its purpose.

cmd2.py
```diff
@@ -1882,7 +1856,7 @@ def do_load(self, file_path=None):
             self.perror('Problem accessing script from %s: \n%s' % (targetname, e))
             return
         keepstate = Statekeeper(self, ('stdin', 'use_rawinput', 'prompt',
-                                       'continuation_prompt', 'current_script_dir'))
```
tleonhardt: Oh, good catch!

cmd2.py
```diff
@@ -548,7 +548,8 @@ def replace_with_file_contents(fname):
     """
     if fname:
         try:
-            result = open(os.path.expanduser(fname[0])).read()
+            with open(os.path.expanduser(fname[0])) as source_file:
```
tleonhardt: Oh, another great catch!  We were never closing that file.

tleonhardt:
I added the parseline() method back since it is actually being used in a roundabout way.  In the process I tried to add a better docstring comment explaining why it is there.

catherinedevlin:
Ugh!  Thank you for realizing that - I woke up in the night realizing I hadn't checked on whether parseline was overriding a core Cmd method.

tleonhardt:
Thank you for catching that nasty bug where my IDE didn't rename one reference to that member variable because it was referring to it indirectly by using its name in a string!

PR #110, opened 2017-05-30
Extract parser methods

catherinedevlin:
Created a ParserManager class to build and store parsers, insulating the main Cmd class from that (still confusing) code.

### DIFF
--- a/cmd2.py
+++ b/cmd2.py
@@ -576,15 +576,19 @@ class Cmd(cmd.Cmd):
     Line-oriented command interpreters are often useful for test harnesses, internal tools, and rapid prototypes.
     """
     # Attributes which are NOT dynamically settable at runtime
+
+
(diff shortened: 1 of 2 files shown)


cmd2.py
```diff
@@ -700,6 +708,10 @@ def __init__(self, completekey='tab', stdin=None, stdout=None, use_ipython=False
 
     # -----  Methods related to presenting output to the user -----
 
+    def _finalize_app_parameters(self):
+        self.commentGrammars.ignore(pyparsing.quotedString).setParseAction(lambda x: '')
+        self.shortcuts = sorted(self.shortcuts.items(), reverse=True)
```
tleonhardt: In the future, we might want to add a comment documenting that self.shortcuts starts out as a dictionary, but here we are turning it into a sorted list of tuples, sorted in reverse order by shortcut.

PR #141, opened 2017-06-25
Fix pastebuffer/clipboard issues on macOS and Linux

tleonhardt:
Various changes include:

- Enable macOS build for Python 2.7 and 3.6 on TravisCI
- Removed some unused code from cmd2.py
- Fixed a couple bugs with pastebuffer/clipboard interaction on macOS and Linux
- Ensure path and shell command completions are sorted alphabetically
- Added a number of unit tests

This closes #141

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,5 +1,5 @@
 install:
-  - python -m pip install pywin32 tox
+  - python -m pip install tox
 
 build: off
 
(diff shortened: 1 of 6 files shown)


tleonhardt:
Unfortunately the pastebuffer/clipboard code unit tests don't run with either of our Windows or Linux Continuous Integration (CI) setups at this time.  These tests do work and pass on these OSes when the correct prerequisites are installed and I've manually tested them on those OSes.

But the out-of-the-box Travis Linux setup doesn't include xclip and the Appveyor default setup doesn't include the pywin32 extensions.

PR #143, opened 2017-06-28
Refactored to remove a few things that felt out of place

### LINKED ISSUE #142: Remove features from cmd2 which probably don't belong there
There are a number of features currently present in ``cmd2`` which seem like they probably should not be present in a generic framework for creating interactive command line applications.  Instead these features should probably exist in wrapper code present in specific applications which lie on top of ``cmd2``. 

Features included in this ticket include:

- loading scripts from URLs (as opposed to files)
- Default file extension for loading text scripts
- Default file name for load/save

I suspect these features are unused and removing them will simplify code maintenance in the future.  If I am wrong and downstream users of ``cmd2`` are using these features, please get in touch and we can work something out.

tleonhardt:
Changes include:

- On TravisCI, comment out mac OS builds for now since heavily backlogged
- The ``load`` command now only loads files, it no longer attempts to load a script from a URL
    - As far as I can tell, this feature didn't work and there were no unit tests.  It also felt out of place
    - If any end user was actually using this feature and can explain to me how it was supposed to work so that I can fix it and add a unit test, I'd be happy to entertain the concept of putting it back in
- The ``defaultExtension`` member variable no longer exists
    - It was only used by load and felt really kludgey and unecessary
- ``autorun_on_edit`` defaults to False
    - This feels like a safer and more expected default
- The ``default_file_name`` member variable no longer exits
    - ``load`` and ``_relative_load`` now require  a file path
    - ``edit`` and ``save`` use a temporary file if a file path isn't provided
- The implementation of ``load`` and ``_relative_load`` has been significantly simplified
- A couple bugs related to calling ``edit`` or ``save`` in a certain way when the history list is empty have been fixed

This closes #142

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,25 +28,25 @@ matrix:
 #          python: pypy3
 #          env: TOXENV=pypy3
         # Stock OSX Python
-        - os: osx
-          language: generic
-          env: TOXENV=py27
-        # Latest Python 3.x from Homebrew
-        - os: osx
-          language: generic
-          env:
-            - TOXENV=py36
-            - BREW_INSTALL=python3
+#        - os: osx
+#          language: generic
+#          env: TOXENV=py27
+#        # Latest Python 3.x from Homebrew
+#        - os: osx
+#          language: generic
+#          env:
+#            - TOXENV=py36
+#            - BREW_INSTALL=python3
 install:
-#  - pip install tox
-  - |
-    if [[ $TRAVIS_OS_NAME == 'osx' ]]; then
-      if [[ -n "$BREW_INSTALL" ]]; then
-        brew update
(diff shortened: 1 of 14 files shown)


tleonhardt:
@FedericoCeratto You are familiar with what features of ``cmd2`` are used by the OpenStack folks right?  From reading the old help text of the ``load`` command for loading ASCII text scripts, it looks like once upon a time it supported loading scripts from URLs.  From what I can see this feature didn't work with the current code base.  So this PR removes that non-working feature.  

If there is some combination of OS and Python version on which it actually works and the OpenStack people are actually using it, please let me know and we can add it back in, fix it up so it works across the board, and add good unit tests.

PR #182, opened 2017-07-09
Use monkeypatch to ensure os.system is restored

felixonmars:
Without this patch, running tests like test_edit_number before
test_default_to_shell_good in the same process will trigger
this error:

```
_______________________________________________________________ test_default_to_shell_good ________________________________________________________________

capsys = <_pytest.capture.CaptureFixture object at 0x7f7e90103898>

    def test_default_to_shell_good(capsys):
        app = cmd2.Cmd()
        app.default_to_shell = True
        line = 'ls'
        statement = app.parser_manager.parsed(line)
        retval = app._default(statement)
        assert not retval
>       assert out == ''
E       AssertionError: assert '*** Unknown syntax: ls \n' == ''
E         - *** Unknown syntax: ls

tests/test_cmd2.py:840: AssertionError
```

Using monkeypatch fixture to restore os.system fixes the problem here.

### DIFF
--- a/tests/test_cmd2.py
+++ b/tests/test_cmd2.py
@@ -661,13 +661,13 @@ def test_edit_no_editor(base_app, capsys):
     expected = _expected_no_editor_error()
     assert normalize(str(err)) == expected
 
-def test_edit_file(base_app, request):
+def test_edit_file(base_app, request, monkeypatch):
     # Set a fake editor just to make sure we have one.  We aren't really going to call it due to the mock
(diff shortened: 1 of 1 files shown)


tleonhardt:
@felixonmars Thanks for the PR!

I'm relatively new to unit testing in Python and pytest.  I realized there was an occasional timing problem when running unit tests in parallel, but didn't know how to fix it.  So I learned something from this.

Thank you!

PR #184, opened 2017-07-09
Avoid conflict the `eg` tool from go-tools

felixonmars:
With go-tools installed there is an `eg` command in system search path.
Let's add an additional character to avoid the test failure.


tleonhardt:
Thank you for the PR, that is a nice little ruggedization.

PR #193, opened 2017-07-19
Corrected help text for shortcuts since they don't have to be single key

kmvanbrunt:


tleonhardt:
@kmvanbrunt You need to fix the "test_base_shortcuts" unit test

PR #211, opened 2017-08-19
.pfeedback() now honors feedback_to_output setting

### LINKED ISSUE #209: feedback_to_output setting doesn't work
Feedback always appears to be redirected into the captured output. Discovered on python3.6.

kotfu:
If feedback_to_output is false, .pfeedback() now sends messages to stderr instead of stdout.

Reluctantly had to modify an another existing test, test_cmd2.py::test_allow_redirection, as a result of this change.

Also fixes a TODO in test_cmd2.py::test_output_redirection to use tempfile.mkstemp to create a temporary file instead of using a hardcoded filename.

Tests pass on macOS running python3.6.

This closes #209

### DIFF
--- a/cmd2.py
+++ b/cmd2.py
@@ -627,7 +627,7 @@ class Cmd(cmd.Cmd):
             if self.feedback_to_output:
                 self.poutput(msg)
             else:
-                print(msg)
+                sys.stderr.write("{}\n".format(msg))
 
     def colorize(self, val, color):
         """Given a string (``val``), returns that string wrapped in UNIX-style
--- a/tests/test_cmd2.py
+++ b/tests/test_cmd2.py
@@ -7,6 +7,7 @@ Released under MIT license, see LICENSE file
 """
 import os
 import sys
+import tempfile
 
 import mock
 import pytest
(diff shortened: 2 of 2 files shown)


tleonhardt:
Thanks for the PR.  I'll review this as soon as I'm home and can get to my laptop (on a tablet now).

From taking a quick look, it seems that a unit test is failing on Windows.  Doesn't look like it fails for a good reason though.  Might need to add a short sleep to file system finish closing file before deleting?

If it is problematic, might be ok to skip that test on Windows as long as feature works well on Windows with manual testing.

cmd2.py
```diff
@@ -627,7 +627,7 @@ def pfeedback(self, msg):
             if self.feedback_to_output:
                 self.poutput(msg)
             else:
-                print(msg)
+                sys.stderr.write("{}\n".format(msg))
```
tleonhardt: I verified manually on my Mac that this change fixes the original issue.

tests/test_cmd2.py
```diff
@@ -504,25 +505,60 @@ def test_save_invalid_path(base_app, capsys):
...
+        # Verify that appending to a file also works
+        run_cmd(base_app, 'help history >> {}'.format(filename))
+        expected = normalize(BASE_HELP + '\n' + HELP_HISTORY)
+        with open(filename) as f:
+            content = normalize(f.read())
+        assert content == expected
+    except:
+        raise
+    finally:
+        os.remove(filename)
```
tleonhardt: Something is going wrong with these unit test changes on Windows in both Python 2.7 and Python 3.x.  The code looks OK to me.  I'll boot up my Windows VM and see if I can get some insight into what is happening.

tleonhardt:
I have reproduced the unit test failure which occurs on Windows only on my Windows 10 VM.  It is the same exception as occurred with the AppVeyor CI tests.

tests/test_cmd2.py
```diff
@@ -504,25 +505,60 @@ def test_save_invalid_path(base_app, capsys):
...
-    run_cmd(base_app, 'help history >> {}'.format(filename))
-    expected = normalize(BASE_HELP + '\n' + HELP_HISTORY)
-    with open(filename) as f:
-        content = normalize(f.read())
-    assert content == expected
-
-    # Delete file that was created
-    os.remove(filename)
+    fd, filename = tempfile.mkstemp(prefix='cmd2_test', suffix='.txt')
+
```
tleonhardt: @kotfu I've been bitten by this exact issue before!  As the Python documentation clearly states, but is non-intuitive to me at least, "mkstemp() returns a tuple containing an OS-level **handle to an open file** (as would be returned by os.open()) and the absolute pathname of that file":
https://docs.python.org/3.6/library/tempfile.html

So immediately after calling mkstemp, you need to call os.close(fd):

```Python
fd, filename = tempfile.mkstemp(prefix='cmd2_test', suffix='.txt')
os.close(fd)
```

Do the same thing in all of the unit tests you modified please and then things should work great ;-)

PR #220, opened 2017-08-24
Piped input now properly honors the echo setting

### LINKED ISSUE #218: Piping input from the shell behaves poorly
```$ echo help | python examples/example.py
(Cmd)
Documented commands (type help <topic>):
========================================
_relative_load  help     mumble  pyscript  save  shell      speak
cmdenvironment  history  orate   quit      say   shortcuts
edit            load     py      run       set   show

stty: stdin isn't a terminal
(Cmd) stty: stdin isn't a terminal
```

Notice how the command is not output by **Cmd2**, but the prompt is. This means that the output can not be used as a transcript. There are several options to improve the behavior:

1. Output the prompt and the command. This makes the output a testable transcript.
2. Output neither the prompt or the command. This makes the output easily parseable by other command line utilities that may want to consume only the output. This is the current behavior of the **load** command, which does a similar thing.
3. Make some setting which allows you to choose between option 1 or option 2. Subclasses of **Cmd2** could get this setting from a command line switch, or permanently set it one way or the other.

The other problem are the mysterious 'stty: stdin isn't a terminal' messages. A cursory glance at the code in **_cmdloop** seems to indicate that **Cmd2** always expects to read input from a tty. When we run the **load** command, the commands to run get added to **.cmdqueue** and nothing is read from stdin.

kotfu:
When input is piped into **cmd2** from something that isn't a tty, we now honor the **echo* setting to determine whether we should print the prompt and the command to the output stream.

Also fixed a bug where we tried to execute **stty** on the input stream, even when it wasn't a tty.

This closes #218

### DIFF
--- a/cmd2.py
+++ b/cmd2.py
@@ -998,6 +998,8 @@ class Cmd(cmd.Cmd):
                 self.poutput(safe_prompt, end='')
                 self.stdout.flush()
                 line = self.stdin.readline()
+                if len(line) == 0:
+                    line = 'eof'
             else:
                 # we are reading from a pipe, read the line to see if there is
                 # anything there, if so, then decide whether to print the
--- a/cmd2.py
+++ b/cmd2.py
@@ -972,7 +972,7 @@ class Cmd(cmd.Cmd):
     def pseudo_raw_input(self, prompt):
         """
         began life as a copy of cmd's cmdloop; like raw_input but
-        
+
         - accounts for changed stdin, stdout
         - if input is a pipe (instead of a tty), look at self.echo
           to decide whether to print the prompt and the input
@@ -984,8 +984,7 @@ class Cmd(cmd.Cmd):
         if self.use_rawinput:
             try:
                 if sys.stdin.isatty():
-                    sys.stdout.write(safe_prompt)
-                    line = sm.input()
+                    line = sm.input(safe_prompt)
                 else:
                     line = sm.input()
                     if self.echo:
--- a/cmd2.py
+++ b/cmd2.py
@@ -970,25 +970,45 @@ class Cmd(cmd.Cmd):
         return result
 
     def pseudo_raw_input(self, prompt):
-        """copied from cmd's cmdloop; like raw_input, but accounts for changed stdin, stdout"""
+        """
+        began life as a copy of cmd's cmdloop; like raw_input but
+        
+        - accounts for changed stdin, stdout
+        - if input is a pipe (instead of a tty), look at self.echo
+          to decide whether to print the prompt and the input
+        """
 
         # Deal with the vagaries of readline and ANSI escape codes
         safe_prompt = self._surround_ansi_escapes(prompt)
 
         if self.use_rawinput:
             try:
-                line = sm.input(safe_prompt)
+                if sys.stdin.isatty():
+                    sys.stdout.write(safe_prompt)
+                    line = sm.input()
+                else:
+                    line = sm.input()
+                    if self.echo:
+                        sys.stdout.write('{}{}\n'.format(safe_prompt,line))
             except EOFError:
                 line = 'eof'
         else:
-            self.poutput(safe_prompt, end='')
-            self.stdout.flush()
-            line = self.stdin.readline()
-            if not len(line):
-                line = 'eof'
+            if self.stdin.isatty():
+                # on a tty, print the prompt first, then read the line
+                self.poutput(safe_prompt, end='')
+                self.stdout.flush()
+                line = self.stdin.readline()
             else:
-                line = line.rstrip('\r\n')
-
+                # we are reading from a pipe, read the line to see if there is
+                # anything there, if so, then decide whether to print the
+                # prompt or not
+                line = self.stdin.readline()
+                if len(line):
+                    # we read something, output the prompt and the something
+                    if self.echo:
+                        self.poutput('{}{}'.format(safe_prompt, line))
+                else:
+                    line = 'eof'
         return line.strip()
 
     def _cmdloop(self):
--- a/cmd2.py
+++ b/cmd2.py
@@ -717,8 +717,9 @@ class Cmd(cmd.Cmd):
         """
         if not sys.platform.startswith('win'):
             # Fix those annoying problems that occur with terminal programs like "less" when you pipe to them
-            proc = subprocess.Popen(shlex.split('stty sane'))
-            proc.communicate()
+            if self.stdin.isatty():
+                proc = subprocess.Popen(shlex.split('stty sane'))
+                proc.communicate()
         return stop
 
     def parseline(self, line):
--- a/tests/test_cmd2.py
+++ b/tests/test_cmd2.py
@@ -1393,9 +1393,9 @@ def test_pseudo_raw_input_tty_rawinput_true():
     # use context managers so original functions get put back when we are done
     # we dont use decorators because we need m_input for the assertion
     with mock.patch('sys.stdin.isatty',
-                mock.MagicMock(name='isatty', return_value=True)):
+            mock.MagicMock(name='isatty', return_value=True)):
         with mock.patch('six.moves.input',
-                    mock.MagicMock(name='input', side_effect=['set', 'quit'])) as m_input:
+                mock.MagicMock(name='input', side_effect=['set', EOFError])) as m_input:
             # run the cmdloop, which should pull input from our mocks
             app = cmd2.Cmd()
             app.use_rawinput = True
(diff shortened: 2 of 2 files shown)


kotfu:
code coverage went down, I think it's because I don't currently test the stdin.isatty() logic branches. Lemme see if I can find a way to mock that and force the execution path over those lines.

kotfu:
I don't know how to write a test to cover all the code. There are two parts of **cmd2.pseudo_raw_input()** that I don't know how to cover. The first part executes if **use_raw_input** is set:
```
if sys.stdin.isatty():
    sys.stdout.write(safe_prompt)
    line = sm.input()
else:
```
I can mock up **.isatty()** to return true, and I can mock up **sm.input()** to return the expected value. However, a side effect of **sm.input()** is that the typed characters are output to **sys.stdout**. I dunno how to make that happen. The **mock(side_effect=func)** parameter lets you dynamically change the return value, but I need it to do other stuff too.

It's the same problem with this chunk, also in **cmd2.pseudo_raw_input()**:

```
if self.stdin.isatty():
    # on a tty, print the prompt first, then read the line
    self.poutput(safe_prompt, end='')
    self.stdout.flush()
    line = self.stdin.readline()
    if len(line) == 0:
        line = 'eof'
else:
```

except in this chunk it's **stdin.readline()** that needs to echo the input characters to **self.stdout**.

Help!

cmd2.py
```diff
@@ -969,25 +970,47 @@ def _surround_ansi_escapes(prompt, start="\x01", end="\x02"):
...
+        """
 
         # Deal with the vagaries of readline and ANSI escape codes
         safe_prompt = self._surround_ansi_escapes(prompt)
 
         if self.use_rawinput:
             try:
-                line = sm.input(safe_prompt)
+                if sys.stdin.isatty():
+                    sys.stdout.write(safe_prompt)
```
tleonhardt: Just curious ... why not just pass **safe_prompt** directly to ``sm.input(safe_prompt)`` like before?  Why call ``sys.stdout.write(safe_prompt)`` first and then ``sm.input()`` with no prompt?

Actually I did some manual testing, it needs to be the other way, very bad things can happen like the user can delete the prompt by backspacing far enough if you do it this way.

tests/test_cmd2.py
```diff
@@ -1388,7 +1389,72 @@ def test_echo(capsys):
     assert app._current_script_dir is None
     assert out.startswith('{}{}\n'.format(app.prompt, command) + 'history [arg]: lists past commands issued')
 
+# the next helper function and two tests check for piped
+# input when use_rawinput is True.
+#
+# the only way to make this testable is to mock the builtin input()
```
tleonhardt: Yeah, this is a pain to test, mocking it is the only way I know of.

cmd2.py
```diff
@@ -969,25 +970,46 @@ def _surround_ansi_escapes(prompt, start="\x01", end="\x02"):
...
 
         # Deal with the vagaries of readline and ANSI escape codes
         safe_prompt = self._surround_ansi_escapes(prompt)
 
         if self.use_rawinput:
             try:
-                line = sm.input(safe_prompt)
+                if sys.stdin.isatty():
+                    line = sm.input(safe_prompt)
+                else:
```
tleonhardt: This else seems to nicely handle the case where stdin is a pipe, both when echo is either True or False

cmd2.py
```diff
@@ -969,25 +970,46 @@ def _surround_ansi_escapes(prompt, start="\x01", end="\x02"):
...
+                    line = sm.input()
+                    if self.echo:
+                        sys.stdout.write('{}{}\n'.format(safe_prompt,line))
             except EOFError:
                 line = 'eof'
         else:
```
tleonhardt: This **else** case is there honestly just because it is there in ``cmd`` and because ``cmd`` defines the ``raw_input`` attribute.  It is never used explicitly in ``cmd2`` other than some unit testing "just in case".

cmd2.py
```diff
@@ -969,25 +970,46 @@ def _surround_ansi_escapes(prompt, start="\x01", end="\x02"):
...
+        """
 
         # Deal with the vagaries of readline and ANSI escape codes
         safe_prompt = self._surround_ansi_escapes(prompt)
 
         if self.use_rawinput:
             try:
-                line = sm.input(safe_prompt)
+                if sys.stdin.isatty():
+                    line = sm.input(safe_prompt)
```
tleonhardt: I committed a small change to restore this line to the way it was due to issues found during manual testing.

tleonhardt:
You can view visual code coverage analysis in the code coverage tool with covered lines highlighted in green and uncovered lines highlighted in red.  But finding the right link to click on can be non-intuitive.  For this PR it is here:
https://codecov.io/gh/python-cmd2/cmd2/pull/220/tree

As you suggested, it is the two isatty branches that start with:

```Python
if sys.stdin.isatty():
```

and

```Python
if self.stdin.isatty():
```

that are uncovered.

I don't think stdin is going to be a tty inside of a unit test.  So these are a pain to test.

cmd2.py
kotfu: My (apparently lame) thought was that I can patch or mock the **sm.input()** but since I don't know how to get the side effect of echoing the characters to stdout, I'd just print them out there first. But if it breaks real-life, who cares whether the tests pass or are easy?

tleonhardt:
Here is one possible unit test approach for the uncovered cases ...

Say we have the block which looks like:

```Python
                if sys.stdin.isatty():
                    line = sm.input(safe_prompt)
                else:
                    line = sm.input()
                    if self.echo:
                        sys.stdout.write('{}{}\n'.format(safe_prompt,line))          
```
Sometimes for difficult cases like this I may just shoot for "branch verification" instead of output or side-effect verification.  

For example, maybe there could be three unit tests for these cases:
- stdin is a tty
    - verify that input() is called with an argument
- stdin is not a tty and echo is True
    - verify that input is called without an argument and that stdout.write is is not called
- stdin is not a tty and echo is False
    - verify that input is called without an argument and that stdout.write is called with an argument

For all of these cases, the following functions could be mocked so that they don't do anything but we can see how many times they are called and with what argument(s):
- sys.stdin.isatty
- sm.input
- sys.stdout.write

This is a weaker form of unit testing than actually verifying output.  But in cases where manipulations are being done to stdin and stdout it is sometimes the best that can be reasonably done.

tleonhardt:
The unit tests I created previously for testing the ``pseudo_raw_input()`` function are these:
- **test_base_cmdloop_without_queue**
- **test_cmdloop_without_rawinput**

Though these tests weren't very focused and tested a number of different things at the same time.  It looks like your new unit tests you added do a better job of focusing in on testing the ``pseudo_raw_input()`` function.

We just might need a couple similar ones which mock ``isatty()`` to return true.  If we only verify branches taken instead of actual output in that case, do you think it would be a reasonable approach?

kotfu:
Yup. I'm working on it right now. I know I can mock **isatty()** to be true, and then I can at least check that **input()** got called with the correct prompt. Stand by.

kotfu:
Uh-oh. This code

https://github.com/python-cmd2/cmd2/blob/c92ac8c634bbf17bd91bd0612d1aae976b53151c/tests/test_cmd2.py#L1439-L1457

Is not deterministic when pytest runs with xdist. I expect because we are patching sm.input() which could be simultaneously used by other tests that are running.

tests/test_cmd2.py
```diff
@@ -1388,7 +1389,106 @@ def test_echo(capsys):
     assert app._current_script_dir is None
     assert out.startswith('{}{}\n'.format(app.prompt, command) + 'history [arg]: lists past commands issued')
 
-
+def test_pseudo_raw_input_tty_rawinput_true():
+    # use context managers so original functions get put back when we are done
+    # we dont use decorators because we need m_input for the assertion
+    with mock.patch('sys.stdin.isatty',
```
tleonhardt: I wasn't aware you could use mocks as context managers.  This is a convenient feature.

tests/test_cmd2.py
```diff
@@ -1388,7 +1389,106 @@ def test_echo(capsys):
...
+    app.use_rawinput = True
+    app.echo = echo
+    # run the cmdloop, which should pull input from our mock
+    app._cmdloop()
+    out, err = capsys.readouterr()
+    return (app, out)
+
+# using the decorator puts the original function at six.moves.input
+# back when this method returns
+@mock.patch('six.moves.input',
```
tleonhardt: I wasn't aware you could use mocks as decorators either.  This is another great feature.

tleonhardt:
@kotfu I always delete the branch after merging into master to keep things clean.  If you ever want me to not delete it, just say something in the PR.

tests/test_cmd2.py
kotfu: Until yesterday, I wasn't either. It's super convenient.

PR #227, opened 2017-09-22
Provide method to run multiple commands w/o a cmdloop.

### LINKED ISSUE #224: load no longer works when used inside onecmd_plus_hooks
As of 0.7.5 load uses cmdqueue, which means running load inside onecmd_plus_hooks no longer works as nothing processes the cmdqueue in that scenario. Ditto for any of the functions which call onecmd_plus_hooks like run().

If do_eos were to return self._STOP_AND_EXIT, and a self._cmdloop() were added back to the end of do_load(), it would work as it used to.

Not sure if such a change is acceptable and compatible with the intent of switching to use cmdqueue to begin with.

alevy03:
runcmds_plus_hooks can accept multiple commands process the command queue to
 deal with subsequent commands loaded from scripts without requiring a command
 loop. This better supports a one-off batch processing scenario.

Also fixed the insertion order of commands placed in the command queue by load
and _relative_load so that script commands are run in the expected order.

Minor tweak to setup instructions in CONTRIBUTING.md to include pyperclip in
prerequisites.

closes #224

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -192,7 +192,7 @@ Once you have cmd2 cloned, before you start any cmd2 application, you first need
 
 ```bash
 # Install cmd2 prerequisites
-pip install -U six pyparsing
+pip install -U six pyparsing pyperclip
 
 # Install prerequisites for running cmd2 unit tests
 pip install -U pytest mock
--- a/cmd2.py
+++ b/cmd2.py
@@ -789,6 +789,44 @@ class Cmd(cmd.Cmd):
         finally:
             return self.postparsing_postcmd(stop)
 
+    def runcmds_plus_hooks(self, cmds):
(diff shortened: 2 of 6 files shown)


CONTRIBUTING.md
```diff
@@ -192,7 +192,7 @@ Once you have cmd2 cloned, before you start any cmd2 application, you first need
 
 ```bash
 # Install cmd2 prerequisites
-pip install -U six pyparsing
+pip install -U six pyparsing pyperclip
```
tleonhardt: Nice catch!  I forgot to update this after adding a dependency on pyperclip

tleonhardt:
@alevy03 I forced a merge of the latest stuff from master into this PR.  Once all of the unit tests pass, I will merge this PR in.

PR #245, opened 2018-01-07
Fix for #244

### LINKED ISSUE #244: Unit test failures in Python 3.7.0a3
When running the ``pytest`` unit tests for ``cmd2`` with the alpha 3 development version of Python 3.7, there are 7 failures, all within the **test_parse_transcript_expected** test suite.

They all appear to be similar in nature, like the following:

```
AssertionError: assert 'use\\ 2/3\\ cup' == 'use\\ 2\\/3\\ cup'
E         - use\ 2/3\ cup
E         + use\ 2\/3\ cup
```

I have no idea if something changed in Python 3.7 which is causing these failures or if it is a change in pytest which is causing them.

kotfu:
In python 3.7, the behavior of re.escape() changed. Rework the test cases in test_transcript.py to ensure they run properly on both 3.4-3.6 as well as 3.7.
 
This closes #244

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,5 @@ cmd2.egg-info
 .cache
 *.pyc
 .coverage
+.tox
 htmlcov
--- a/cmd2.py
+++ b/cmd2.py
@@ -2398,7 +2398,22 @@ class Cmd2TestCase(unittest.TestCase):
             self.assertTrue(re.match(expected, result, re.MULTILINE | re.DOTALL), message)
 
     def _transform_transcript_expected(self, s):
-        """parse the string with slashed regexes into a valid regex"""
+        """parse the string with slashed regexes into a valid regex
+        
+        Given a string like:
+        
+            Match a 10 digit phone number: /\d{3}-\d{3}-\d{4}/
+        
+        Turn it into a valid regular expression which matches the literal text
+        of the string and the regular expression. We have to remove the slashes
+        because they differentiate between plain text and a regular expression.
+        Unless the slashes are escaped, in which case they are interpreted as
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -6,4 +6,5 @@ cmd2.egg-info
 .cache
 *.pyc
 .coverage
+.tox
```
tleonhardt: Good catch, that should have always been there ;-)

cmd2.py
```diff
@@ -2398,7 +2398,22 @@ def _test_transcript(self, fname, transcript):
             self.assertTrue(re.match(expected, result, re.MULTILINE | re.DOTALL), message)
 
     def _transform_transcript_expected(self, s):
-        """parse the string with slashed regexes into a valid regex"""
+        """parse the string with slashed regexes into a valid regex
```
tleonhardt: Nice description

tests/test_transcript.py
```diff
@@ -309,21 +310,25 @@ def test_transcript(request, capsys, filename, feedback_to_output):
 
 
 @pytest.mark.parametrize('expected, transformed', [
-    ( 'text with no slashes', 'text\ with\ no\ slashes' ),
-    # stuff with just one slash
-    ( 'use 2/3 cup', 'use\ 2\/3\ cup' ),
-    ( '/tmp is nice', '\/tmp\ is\ nice'),
-    ( 'slash at end/', 'slash\ at\ end\/'),
+    # strings with zero or one slash or with escaped slashes means no regular
```
tleonhardt: Looks good.

PR #247, opened 2018-01-08
Argparse support

### LINKED ISSUE #36: Investigate replacing optparse with argparse
optparse has been deprecated since Python 3.2 and all development on it has ceased:
https://docs.python.org/3.6/library/optparse.html

We should investigate replacing the usage of optparse with argparse or a similar library such as docopt or click.

Since the APIs aren't identical, it won't be a trivial drop-in replacement.

kotfu:
This PR is for issue #36.

A few questions for discussion:

1. I have not removed the optparse code, although the documentation now says it is deprecated. I think we should leave the optparse code, which will give users of cmd2 time to migrate to the new argparse decorator.

2. Do you think I should write a section of documentation the briefly shows how to migrate from the optparse decorator to the argparse decorator? I don't think cmd2 is the right place for a optparse to argparse migration guide, but what do you think?

3. [EDIT] We need to resolve the parse_args() vs parse_known_args() question.
  
This closes #36


kotfu:
If we used `parse_known_args`, the method signature for the command would look like:

```
def do_sspeak(self, argv, args, unused_args)
end
```

This would give you ready access to the unused arguments, but it takes away the capability of argparse to enforce valid options. If I type `speak --bogusargument hello` I won't get a usage message, I'll just get `hello`. `nargs='*'` provides a way to get the rest of the arguments if you need them. If we decide we need a `parse_known_args` capability, I suggest we create two decorators, one that calls `parse_args` and another which calls `parse_known_args`.

tleonhardt:
Oh man, I don't like how GitHub added all of my comments to specific commits without the context of the lines I was commenting on!

You make an excellent point about the capability of argparse to enforce valid options - that is nice.  I like the two decorator approach.  I think the ``parse_known_args`` approach might make it easier to port existing ``cmd2`` optparse code since existing decorated methods accept two arguments, opts and args. But the ``parse_args`` approach may be a better approach for new "from scratch" code.

I think the best way to document for users how to migrate existing ``cmd2`` optparse decorator code to new ``cmd2`` argparse decorators code would be in an example where we show a couple essentially identical methods, one which uses each decorator.  Then we can update the existing argument parsing documentation within this PR, to say see that example for an illustration of how to port existing code.

I fully expect this solution may slightly evolve after it merges into master and some early adopters give feedback regarding migration.

tleonhardt:
@kmvanbrunt 
Kevin, if you have time to review and comment, I'd appreciate another opinion since this is a really significant feature change.

tleonhardt:
I deleted a few comments which were not relevant to this PR since they were from very early discussions.

tleonhardt:
So now I'm leaning more towards a single decorator approach.  I don't want to give end users too many ways of doing something similar - that could risk confusing them.  I think the existing approach embodied in this PR seems ok.

@kotfu Is there anything else you would like to add, or should we merge it in?

PR #248, opened 2018-01-11
Argparse bugfix

kotfu:
- Fixed a nasty bug I introduced in the last PR (oops)
- Update README.md to reference argparse and new argparse decorator instead of old optparse stuff
- Update some examples to use argparse instead of optparse


tleonhardt:
Hmm dependency-ci appears to be temporarily off-line.

PR #249, opened 2018-01-13
decorator changes for replacing optparse with argparse

### LINKED ISSUE #47: All commands should go through same shlex parsing
Currently arguments for commands using the options decorator are split using shlex.split(), however arguments for commands which do not use the options decorator are not split in this fashion.

Ideally, arguments for all commands would go through a more similar processing chain.

The real question, is where is the best place in the code to do that?

### LINKED ISSUE #250: Use @with_argument_parser for all methods which require argument parsing
Many methods in cmd2 utilize the now deprecated **@options** decorator, ex. **do_pyscript()** and **do_history()**. Other methods like **do_save()** use `pyparsing`, which works, but it too should probably be migrated to **@with_argument_parser**. **do_set()** does custom argument parsing not utilizing any library. Still other methods do no argument parsing, but probably should, like **do_load()**. All of these should be moved to `argparse` based argument processing.

Methods which implement single word commands, like **do_quit()** and **do_shortcuts()** could be left to ignore all passed arguments, which is what they do now.

### LINKED ISSUE #251: Move read-only settings from cmdenvironment to set
The **cmdenvironment** command outputs settings that the end user can not modify. This command is generally only useful to developers, and may be confusing to users.

Suggest modifying the set command to accept a **-a/--all** option. If you do **set --all** then you get the standard settings, and then a blank line, and then a line that says "Read only settings", and then the output of cmdenvironment. That way this information would still be available, but would be less confusing for end users because there would be no cmdenvironment command in help.

### LINKED ISSUE #252: Consolidate functionality of editing, saving, and showing the history of commands entered
The **do_history()** command allows you to search and slice history before displaying it. The **do_save()** command doesn't, but it will save history to a file. The **do_edit()** command lets you edit one history item and then run it. We should consolidate all the history functionality into a single command.

- [x] Add a **-o/--output-file** argument to **do_history()** which writes the selected history item(s) to the given file
- [x] Remove the **do_save()** method and it's associated tests
- [x] Add a **-r/--run** argument to **do_history()** to run the selected command(s)
- [x] Remove the **do_run()** method and it's associated tests
- [x] Add a **-e/--edit** argument to **do_history()** to let you edit the selected history item(s) in an editor. Run the contents of the edited file when the editor exits.
- [x] Modify **do_edit()** to not edit commands from history, just to edit files
- [x] Remove **_last_matching()** method
- [x] Remove **autorun_on_edit** attribute and setting

### LINKED ISSUE #253: Hide __relative_load command from the help menu by default
The **__relative_load** command is not a command which an end user would ever want to invoke directly from the interactive command line because it is intended to be used within a script that then loads another script.

As such, it would probably be best to hide it from the help menu to prevent any unnecessary confusion to end users.

kotfu:
This PR adds a new **@use_argument_list** decorator which converts **do_*** methods to get passed a list of strings instead of a string.  This is in lieu of using the **@with_argument_parser** decorator for full-blown argparse parsing.

This closes #47
This closes #250
This closes #251 
This closes #253
This closes #252


cmd2.py
```diff
@@ -500,6 +518,7 @@ class Cmd(cmd.Cmd):
     excludeFromHistory = '''run ru r history histor histo hist his hi h edit edi ed e eof eo eos'''.split()
     exclude_from_help = ['do_eof', 'do_eos']  # Commands to exclude from the help menu
     reserved_words = []
+    use_argument_list = False
```
tleonhardt: This class member shouldn't be added.  It creates massive backwards compatibility problems with ``cmd``.

cmd2.py
```diff
@@ -1007,7 +1026,18 @@ def onecmd(self, line):
             func = getattr(self, funcname)
         except AttributeError:
             return self.default(statement)
-        stop = func(statement)
+
+        if self.use_argument_list:
```
tleonhardt: We can't do this in **onecmd**.   This breaks any underlying calls which are actually implemented by Python's built-in ``cmd`` module which ``cmd2`` extends - most importantly **help**.  With this in place doing a "help history" will  throw a "TypeError" exception because the underlying ``cmd`` help commands expects a string.

Additionally, if someone tries to upgrade their ``cmd`` app to use ``cmd2``, at first none of their commands would work.

cmd2.py
```diff
@@ -276,6 +276,24 @@ def cmd_wrapper(instance, cmdline):
     return arg_decorator
 
 
+def with_argument_list(func):
```
tleonhardt: I like this decorator-based approach for getting ``cmd2`` to pass the do_* methods an argument list instead of an argument string.  It is an easy way to alter the default behavior on a per-function basis.

docs/argument_processing.rst
```diff
@@ -139,6 +142,57 @@ Which yields:
    This command can not generate tags with no content, like <br/>
 
 
+Receiving an argument list
+==========================
+
+The default behavior of ``cmd2`` is to pass the user input directly to your
+``do_*`` methods as a string. If you don't want to use the full argument parser support outlined above, you can still have ``cmd2`` apply shell parsing rules to the user input and pass you a list of arguments instead of a string. There are two methods to effect this change.
+
+The ``use_argument_list`` attribute of a ``cmd2`` class or subclass defaults to ``False``. Set it to ``True`` and all of your ``do_*`` methods will receive a list of arguments parsed by ``shlex.split()`` instead of receiving a string::
```
tleonhardt: Update this documentation once the **use_argument_list** attribute has been removed.

tleonhardt:
On second thought, including the ``use_argument_list`` attribute and modifications to **onecmd()** can be ok.  But if we keep them, we need to put in a modification to **do_help()** so that the underlying call to **cmd.Cmd.do_help()** always gets a ``str`` even when ``use_argument_list`` is True.  So basically we would need to put in a "''.join()'' to do the conversion.

kotfu:
So to keep the `use_argument_list` attribute, we need to go find all occurrences of do_* methods in both `cmd.cmd` and `Cmd2.cmd` and make sure they still work properly. After looking at `do_help()`,  that one will be easy. But to continue on this path we need to find them all. I think I'll investigate using introspection in `onecmd()` to check if the method we are about to call is part of `cmd.cmd` or `Cmd2.cmd`, and if so, pass the string instead of the argument list. That way if in the future we add a new `do_*` method to `Cmd2.cmd` we won't create an unexpected bug.

If this still turns out to be too problematic, perhaps we can try creating a new subclass of str which has a arglist() method. Pass an instance of this new class to the do_ methods where we are currently passing a plain string. Existing code works fine, if you want a list of arguments, call the arglist() method.

tleonhardt:
Yes, to keep the ``use_argument_list`` attribute, we need to find all occurrences of **do_*** methods in ``cmd2.Cmd`` and make sure they work properly both when that argument is **True** and **False**.  Additionally, we would need something like you mention with introspection within **onecmd()** so that if the underlying method call is to something within ``cmd.Cmd``, that it always gets passed a string instead of a list.  And we would need a tweak within the ``cmd2`` override of **do_help**.

Many of the **do_*** methods in ``cmd2`` currently use the **@options** decorator and have a snippet of code within them such as the following to behave one way when a str is passed and another way when a list is passed:

```Python
# If arguments are being passed as a list instead of as a string
        if USE_ARG_LIST:
            if arg:
                arg = arg[0]
            else:
                arg = ''
```

kotfu:
@tleonhardt After looking at some of the **do_*** methods in `cmd2.py`, I thought of another potential solution. Most of the **do_*** methods in `cmd2.py` (i.e. **do_set()**, **do_pyscript()**) have some way of parsing arguments. Some use **@options*, others use `pyparsing`, others just do custom parsing. If we migrate all of that argument parsing to our new `argparse` based decorator, that avoids any potential issues with handling plain string vs arguments as a list. Issue #250 contains all the details.

Since I think we have to do issue #250 anyway, at this point I'm inclined to try and implement the changes we have discussed to make `use_argument_list` work, and see how ugly it is. If it's horrible, then I think we just abandon `use_argument_list`. If it's not horrible, and if I can do it in a way that's not too brittle, then I think we should use it.

kotfu:
@tleonhardt, I don't understand your prior comment: 'Additionally, if someone tries to upgrade their cmd app to use cmd2, at first none of their commands would work.' Can you better explain your concern?

tleonhardt:
@kotfu 
Thanks for creating #250.  We do need to replace the use of the **@options** decorator since we are calling that deprecated.

Yeah, definitely go forward with the use of ``use_argument_list`` and implement the changes as discussed.  Hopefully there won't be any gotchas we didn't think about.

As long as the default value of ``use_argument_list`` is **False**, I think my previous concern is unfounded.  We just need the default behavior for **do_*** methods without any decorator to be consistent with ``cmd`` where they get passed a single argument which is a ``str``.

kotfu:
While working on Issue #250, I discovered that we will have to discard the `use_argument_list` attribute. It breaks **do_shell()** and **do_py()**. I'm gonna think a bit more about this and see if I can come up with an alternative solution, but as of right now, I think we will have to stick with the decorators only.

tleonhardt:
Now that we have a 2nd decorator which simply converts a **do_*** command so that it gets passed an argument list instead, I'm pretty happy with the decorator approach.  That feels to me like a relatively painless, explicit, and self-documenting API.

If you can think of a smart and safe alternative to the ``use_argument_list`` attribute which gives similar functionality, I'd be curious to see the approach.  But I wouldn't recommend spending too much time thinking about it.

tleonhardt:
@kotfu I hope me pushing a few minor fixes doesn't get in the way of what you are doing

cmd2.py
```diff
@@ -1266,17 +1292,14 @@ def select(self, opts, prompt='Your choice? '):
...
-                arg = arg[0]
-            else:
-                arg = ''
-
-        param = arg.strip().lower()
+    argparser = argparse.ArgumentParser(description='show value of a parameter')
+    argparser.add_argument('-l', '--long', action='store_true', help='describe function of parameter')
+    argparser.add_argument('param', nargs='?', help='name of parameter, if not supplied show all parameters')
+    @with_argument_parser(argparser)
+    def do_show(self, arglist, args):
```
tleonhardt: If all of the arguments are being parsed by ``argparse`` when the **with_argument_parser** decorator is used, then why are we also passing the full arglist argument?  This seems redundant and unnecessary.

cmd2.py
```diff
@@ -242,23 +242,33 @@ def strip_quotes(arg):
     return arg
 
 
+def parse_quoted_string(cmdline):
```
tleonhardt: Good job breaking this functionality out into a function.  That reduces duplicated code.

cmd2.py
```diff
@@ -1148,7 +1173,7 @@ def _cmdloop(self):
             return stop
 
     # noinspection PyUnusedLocal
-    def do_cmdenvironment(self, args):
+    def do_cmdenvironment(self, _):
```
tleonhardt: I like how using **_** for the argument name makes it nice and explicit and self-documented that the argument is unused.

cmd2.py
```diff
@@ -1589,32 +1612,28 @@ def quit():
             self._in_py = False
         return self._should_quit
 
-    # noinspection PyUnusedLocal
-    @options([], arg_desc='<script_path> [script_arguments]')
-    def do_pyscript(self, arg, opts=None):
+    @with_argument_list
+    def do_pyscript(self, arglist):
```
tleonhardt: For some of these commands that got converted from an empty **options** decorator with a non-empty ``arg_desc`` argument to a **with_argument_list** decorator, we may want to add an explicit "usage" statement to the help text.

cmd2.py
```diff
@@ -1817,7 +1833,8 @@ def _current_script_dir(self):
         else:
             return None
 
-    def do__relative_load(self, file_path):
+    @with_argument_list
+    def do__relative_load(self, arglist):
```
tleonhardt: Random train of thought, I've been thinking about adding **do__relative_load** to the ``exclude_from_help`` list (which **do_eof** and **do_eos** are already in).  Adding a **do_*** method to this list simply excludes it from appearing in the help menu when an empty **help** command is issued.  Do you have an opinion on this?

Basically, is it useful for most end users to see this in the help menu, or would its presence be more likely to confuse them?

cmd2.py
kotfu: Thanks for fixing the unit tests that I broke 😳

cmd2.py
kotfu: I think we should exclude **do__relative_load** from the help by default. Developers can easily add it back in for their use case if they really want it.

cmd2.py
kotfu: I didn't really think about it, I just did it the same way that the **@options** one worked. But as I think about it, there is no reason to pass the argils too. You can't go backwards to the original input from the arglist, so it really is redundant and not very useful. After I finish migrating **do_save()** to use **@with_argument_list** I'll make this change.

tleonhardt:
@kotfu I just pushed a change where I changed the **with_argument_parser** decorator to only pass a single argument to **do_*** commands (no more arglist).

I also added a new **with_argparser_and_list** decorator which uses **argparse.parse_known_args** to pass two arguments to commands - the argparse args and a list of unknown args.

If you don't like the changes, we can undo them.

I also finished moving a few of the examples to the new decorators.

kotfu:
I like it. The **_and_list** feels like it could be more descriptive. What do you think about changing the name of the new decorator from **with_argparser_and_list** to **with_argparser_and_unknown_args**?

tleonhardt:
I like your name better ;-)  It is much more descriptive and self-documenting.

tleonhardt:
Ok, I'm done working on #251 and #253, last least for now.  

Let me know when you push changes for #252 and I'll take a look at them.  That one is a little more involved.  But once we wrap that up, I think we should finally merge this PR.

tleonhardt:
I like how the new and improved history command is behaving.  FYI, I think it is "cowardly", not "cowerdly".

So now to complete this ticket we just need to remove the deprecated **run** and **save** commands and remove the command editing capability from the **edit** command.

Random train of thought for another ticket.  Now that **history** command has the ability both to execute previous commands and to output previous commands to a file, it might be cool to add the capability for it to automatically generate a transcript.  I don't know how hard that would be, but if it was easy, that would be a really cool feature to create an Issue for.

kotfu:
@tleonhardt I think we've got everything done. I just updated the changelog, and I think this PR is ready to merge.

CHANGELOG.md
```diff
@@ -2,11 +2,32 @@
 * Bug Fixes
     * Fixed unit tests on Python 3.7 due to changes in how re.escape() behaves in Python 3.7
 * Enhancements
-    * Added new **with_argument_parser** decorator for argparse-based argument parsing of command arguments
-        * This replaces the old **options** decorator for optparse-based argument parsing
-            * The old decorator is still present for now, but should be considered *deprecated* and will eventually be removed
-        * See the **Argument Processing** section of the documentation for more information
-            * Alternatively, see the **argparse_example.py** example
+    * Three new decorators for **do_*** commands to make argument parsing easier
```
tleonhardt: Thanks for updating the Changelog

README.md
```diff
@@ -236,7 +236,6 @@ example/transcript_regex.txt:
 # regexes on prompts just make the trailing space obvious
 (Cmd) set
 abbrev: True
-autorun_on_edit: False
```
tleonhardt: My recent commit was just a minor cleanup of documentation and such

PR #255, opened 2018-01-17
Improved how new argparse-based decorators provide help

tleonhardt:
Now "help command" and "command -h" provide exactly the same help text.

The function docstring for the "do_*" command sets and overrides the ArgumentParser "description" if the docstring is not empty.

### DIFF
--- a/cmd2.py
+++ b/cmd2.py
@@ -288,13 +288,10 @@ def with_argparser_and_unknown_args(argparser):
         argparser.prog = func.__name__[3:]
 
         # put the help message in the method docstring
-        funcdoc = func.__doc__
-        if funcdoc:
-            funcdoc += '\n'
-        else:
-            # if it's None, make it an empty string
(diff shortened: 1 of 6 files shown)


tleonhardt:
@kotfu 
I started porting an existing app to use the new argparse-based decorators and didn't like how the output of "help command_name" wasn't the same as "command_name -h".  So this PR fixes that.

tleonhardt:
I wanted to make "help command" consistent with "command -h" and there are three different ways I can see doing this.  I chose one.  If you prefer one of thee others we could do that.

The three ways I could see are:

* Have docstring overwrite the ArgumentParser ``description``, so it appears near the beginning of help text immediately after the usage
* Have docstring prepend the ArgumentParser ``usage``, so it appears at the very beginning of the help text
* Have docstring overwrite the ArgumentParer ``epilog``, so it appears at the very end of the help text

PR #257, opened 2018-01-20
Sub-commands and automatic transcript generation

### LINKED ISSUE #71: Consider adding support for nested commands or subcommands
Many REPL frameworks have support for multi-level commands, where the command has sub-commands.

While this can be done with cmd2, there isn't any built-in support to make it easy.

Consider adding some built-in support to make it easy to have nested commands or subcommands.

### LINKED ISSUE #254: Explore having history command automatically generate a transcript
Now that the **history** command is capable of both executing previous commands and saving previous commands to a file, it would be really great if it could essentially combine these capabilities so that it could automatically generate and save a transcript file which could then be used with the transcript testing built-in regression testing feature.

### LINKED ISSUE #258: Unknown commands should not end up in the history
Currently if an unknown command is executed, that command ends up in the history and is displayed by the **history** command.

Unknown commands should not be saved in the history.

tleonhardt:
Added support for argparse sub-commands when using ``cmd2`` decorators

Modified the ``do_help()`` method to behave differently for commands which have been decorated with an argparse ArgumentParser.  This is so that the **help** command will properly deal with sub-command help.

Suppose you have a base command **base** which has two sub-commands, **foo** and **bar**.  Then "help base" will provide very different help text than "help base foo".

Slightly tweaked the two argparse decorators to set an attribute in the decorated function's dictionary so that the do_help method can know which functions have an ArgumentParser and which do not.

Added a "subcommands.py" example for demonstrating how to create and use subcommands based on argparse and the cmd2 @with_argument_parser decorator.

Tab-completion of subcommand names is supported out-of-the box - it just requires the user to pass a 2nd argument to the decorators.

Also bumped version from 0.8.0a to 0.8.0

Unless someone has any bright ideas regarding how to improve support for sub-commands, this takes care of it.

Edit:
- Fixed a bug where unknown commands were getting saved in the history
- Added a -t flag to the history command which automatically generates a transcript file based on selected contents within the history

This closes #71
This closes #254 
This closes #258


tleonhardt:
@kotfu If you could review the general ``argparse`` usage of sub-commands and modifications to the decorators and help it would be appreciated.

@kmvanbrunt If you could review the tab completion of sub-commands with an eye towards edge cases I might have missed, I would appreciate that as well.

tleonhardt:
I plan to add a few unit tests, but other than that I'm done with this PR now (pending feedback).

tleonhardt:
I added a number of unit tests for sub-commands.

The overall code coverage will inevitably have dropped due to the fact that I had to override the cmd.complete() method which is a relatively low-level function which provides tab-completion based on readline.

In order to add unit tests of that function, we will have to mock out all of the readline calls.

tleonhardt:
I added a number of unit tests to to mostly cover the newly overridden ``complete()`` method.  This should bring the code coverage back up.

tleonhardt:
The one asymmetry I don't like about how help is being handled here is that getting **help** on a command decorated with an ``argparse.ArgumentParser`` sends the help text to ``sys.stdout`` instead of to ``self.stdout``.

I suppose we could live with that minor inconsistency, but if there is a way to fix it, it would be welcome.

tleonhardt:
**history** command now has a -t flag which automatically generates a transcript file for regression testing based on the chosen contents of the history.

Also fixed a bug where unknown commands were being saved in the history.

tleonhardt:
I did try to find a way to interrogate the ``argparse`` **ArgumentParser** object to find out what sub-commands it had.  I was able to determine if there were any subcommands, but not what their names were.

I think your suggestion about changing the sub-command names in **subcommand.py** to **base_foo()** and **base_bar()** to set a good example/convention is a great idea.  I'll do that.

In regards to renaming the decorator, we haven't created a release to PyPI yet, so I view that as no big deal.  Better to make the improvement now.  So I will do that as well.

tleonhardt:
Ok, I renamed that stuff you suggested.

While we are at renaming the decorators, can you think of a better slightly shorter name for **@with_argparser_and_unknown_args**?  It just feels a little long.

Also, can you of a good way to make it so that **help** for commands which use one of the argparse decorators goes to **self.stdout** instead of **sys.stdout**?  It would be nice to have that consistent.  If for no other reason, then doing a **history -t transcript.txt** won't save **help command** text to the transcript for commands which use these decorators because their output is coming directly from ``argparse``.

tleonhardt:
TravisCI is saying that the build works on Python 2.7, 3.5, 3.6, and 3.7; but is broken on Python 3.4.  But the error messages don't make any sense to me.  So I'm not sure if something is messed up with TravisCI's infrastructure for Python 3.4, if one of the required test dependencies pushed an update that broke on Python 3.4, or if I actually made a change which broke ``cmd2`` on Python 3.4.

I'll try to get a local virtualenv setup for testing.

tleonhardt:
I'm going to merge this to master before it gets any bigger.  If we find any problems, we can open an Issue and we will wait a little while before creating a release to PyPI so we have some time to use the new and modified features.

PR #261, opened 2018-01-30
Fixed issue with automated transcript generation via "history -t"

### LINKED ISSUE #259: Explore issue with automated transcript generation
Currently the ability to automatically generate a transcript with **history -t transcript_file.txt** works well, but not perfectly.  There are a minority of cases where it generates a transcript which nominally looks correct but which fails actual transcript testing.

The automated transcript generation feature is awesome if users can count on it 100% of the time (or at least very close to it).  So we need to figure out what is going wrong in the minority of cases.  Perhaps we need to do something slightly different when generating the transcript.  Or perhaps, transcript testing could be a little bit more permissive, particularly towards differences in trailing whitespace at the end of a line, end of a command, and end of the file.

The one case I am certain of that creates a bad transcript is:

```Bash
help history
history -t transcript.txt
```

Edit:  The title and text changed to reflect that we need to figure out what is going on and to try to avoid jumping to a conclusion in regards to a potential solution.

tleonhardt:
There is now a post-processing step which escapes all "/" characters which transcript testing treats as a regex escape if there isn't a "\" to escape it.

This closes #259.


tleonhardt:
@kotfu Since you wrote or modified most of the regex processing code within the transcript testing code, you may have a better idea of how to fix this corner case.  But this PR does appear to fix it in a relatively simple and straightforward manner.

kotfu:
Just missed adding this note before you merged.

Looked at the post processing in the history transcript code. I'd like to pose a question for consideration:  Do we think the transcript output would ever be too big for memory? Or to state it in another way, should we optimize for reduced memory usage or reduced i/o usage?

If we think the answer to that question is no (or reduced i/o usage), then we could perhaps be more efficient in the post processing if instead of opening the transcript file and setting it to self.stdout, we created a StringIO file (a file in memory) and set that to self.stdout. We then do the replacement on the contents of the StringIO file, and then write it all out to the specified transcript file. This would save us a read and a write of the entire transcript. Let's call this alternative 1.

If we think the answer to the question is yes (or reduced memory usage), then maybe we modify the post processing to open a temporary file and set it to self.stdout. We then open the temporary file for read, and the real transcript file for write, and read the lines one at a time, do the replacement, and write them to the real transcript file. Thereby never having to have the entire transcript in memory. Let's call this alternative 2.

Reality is likely that nobody would notice whether we leave it as is, or do one of these other two options. One benefit I like of the current implementation is that there is no reliance on a temporary file, which I think is a drawback of alternative 2. Alternative 1 depends on the `six` library for compatibility between python 2 and 3, because the StringIO library is different.

tleonhardt:
Personally I have never envisioned a situation where a transcript would be too big to fit in memory.  But you bring up a good point.  If that situation were ever possible, then we would have a problem.  Can you envision a situation where you think that would be likely?

If we call the current implementation "alternative 0", then do you have any preference as to alternatives 0, 1, or 2?  They all seem like viable options with their own distinct advantages and disadvantages.  I would be happy with any of them.

I wouldn't be concerned about using the ``six`` library for StringIO compatibility between Python 2 and 3 since we are using it for so many other compatibility issues.

tleonhardt:
@kotfu Are there any outstanding issues other than the possible resolution here between alternatives 0, 1, and 2 that you would like to see addressed prior to putting out a **0.8.0** release?

I was thinking of putting out a release within the next few days since we have made a lot of significant changes recently.

PR #270, opened 2018-02-09
Added optional persistent readline history feature

### LINKED ISSUE #265: Consider adding optional persistent history
It would be nice if there was an easy way to add persistent history to ``cmd2`` applications.

We would need to decide if this is done for readline history, ``cmd2`` **history** command history, or both.  Perhaps there would be two separate options for making each of these persistent?

We could probably add an optional argument or two (with default values) to the cmd2.Cmd.__init__() initializer.

[This](https://kushaldas.in/posts/developing-command-line-interpreters-using-python-cmd2.html) blog post demonstrates the basics of how to make the readline history persistent:

```Python
import os
import atexit
import readline

history_file = os.path.expanduser('~/.mycli_history')
if not os.path.exists(history_file):
    with open(history_file, "w") as fobj:
        fobj.write("")
readline.read_history_file(history_file)
atexit.register(readline.write_history_file, history_file)
```

For ``cmd2`` **history** command serialization, we could probably just serialize/deserialize the contents of self.history using the ``pickle`` module.

tleonhardt:
Added two additional optional parameters to the ``cmd2.Cmd`` class **__init__()**:
- **persistent_history_file** - file path to load a persistent readline history from
- **persistent_history_length** -  max number of lines which will be written to the history file

If **persistent_history_file** is supplied and is non-empty, then at the beginning of its initialization, the ``cmd2.Cmd`` class attempts to expand this path and load readline history from this file.  If the file isn't found, it just moves on.  It also uses **atexit** to register saving the readline history to this file when the ``cmd2`` application exits.

The **persistent_history_file** argument defaults to an empty string and thus by default the ``cmd2.Cmd`` application does NOT use a persistent readline history.

Included an example and info in the Sphinx docs.

Unrelated changes added to this PR:
- Created CHANGELOG entry for 0.8.1 release
- Added info to README about new sub-menu feature
- Bumped version to 0.8.1

This closes #265.


tleonhardt:
@kotfu I can't figure out a good way to unit test readline history behavior to unit test this new persistent history feature.  I manually tested it and verified it works.  But I would prefer to have a unit test.  Can you think of any way to unit test this?

tleonhardt:
So the **pexpect**-based unit test works fine on my computer and I tested on both macOS and Linux.  However, it doesn't work on TravisCI.

Issue #271, opened 2018-02-12
Multiple decorators breaks help on subcommands


kotfu:
Make the attached modifications to [examples/subcommands.py](https://github.com/python-cmd2/cmd2/files/1715319/subcommands.txt). I added a simple decorator, which I then added to `do_base()`. With the additional decorator, the `help` command does not behave as expected.

`help base` yields:

```
usage: nimal_decorator [-h] {foo,bar} ...

Base command help

optional arguments:
  -h, --help  show this help message and exit

subcommands:
  {foo,bar}   subcommand help
    foo       foo help
    bar       bar help
```

Notice the incorrect and mangled command name on the first line.

If you switch the order of the decorators on `do_base()`, i.e. so `@minimal_decorator` is first, instead of second, `help base` now behaves properly, but `help base foo` and `help base bar` are not recognized, they return the same thing as `help base`.

tleonhardt:
Hmm .. this one seems like it might be tricky to fix.  Do you have any good ideas?  

Getting the help to work correctly for subcommands at all was a pain, in part because I couldn't find a way to query an ``argparse.ArgumentParser`` to get access to its subparsers - it was easy to find out if it hand any subparsers, but not to get access to them.

kotfu:
I haven't even looked at the help code yet, I'll try and take a peek tomorrow.

kotfu:
Yikes. I'm not sure the current implementation can be modified to solve this problem. I can't think of a way to make the `__dict__['has_parser']` bits work when there could be multiple decorators and those decorators could be in any order. We may have to try and think of an alternative design pattern, like some class method which you call to register your arg parser where you pass the arg parser and the method the arg parser should be applied to.

Still thinking about this one... but totally agree that this might be tricky to fix.

tleonhardt:
I'm not sure, but I think that using the [functools.wraps](https://docs.python.org/3.6/library/functools.html) helper in our decorators to carry over the metadata from the undecorated method may solve a lot of problems for us.

kotfu:
Didn't know about `functools`, looks super useful. `functools.wraps` could solve the problem, but it won't work if you want to use a decorator from another 3rd party package.

tleonhardt:
#283 closes this

PR #283, opened 2018-02-24
Added use of @functools.wraps() in our decorators

tleonhardt:
This updates our decorator wrapper functions to look like the wrapped function.

This partially addresses Issue #271 where using multiple decorators can break help on subcommands, depending on the order in which the decorators are applied.

This PR doesn't actually fix anything in and of itself.  But it encourages the usage of @functools.wraps().  If other decorators use this, then there shouldn't be any problem.  Of course, we can't control how 3rd-party libraries implement decorators.  But we can at least be part of the solution instead of part of the problem.

### DIFF
--- a/cmd2.py
+++ b/cmd2.py
@@ -30,6 +30,7 @@ import cmd
 import codecs
 import collections
 import datetime
+import functools
 import glob
 import io
 import optparse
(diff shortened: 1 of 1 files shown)


tleonhardt:
@kotfu Is there more you would like to do for #271?  Or should this PR close it?

PR #284, opened 2018-02-24
Converted a few class variables into instance variables

### LINKED ISSUE #273: Consider which class variables should be made into instance variables
PR #272 brought up the fact that now that users can nest instances of ``cmd.Cmd2`` to support creating sub-menus, we should revisit which class variables should be made into instance variables to prevent potential problems.

tleonhardt:
Now that users can nest instances of cmd.Cmd2 to support creating sub-menus, we should need to be more careful about class vs instance variables to prevent potential problems.

This converts the following former class variables into instance variables:
- multiline_commands
- shortcuts
- exclude_from_help
- exclude_from_history

In the process, a couple camelCase variable names got converted to pep8_compliant names.

There may be a few other class variables which should be converted to instance variables.  But at the very least, this is a good start.


tleonhardt:
You make a good point.  On second thought, maybe changing the two attribute names isn't such a hot idea.  It would even break our own examples.

I think the change would be dramatically less breaking if I didn't rename those two.  All of our examples reference the class variables

I don't view the act of making these attributes instance members instead of class members to be particularly breaking, especially since I think all of our examples reference them via ``self`` anyways.

I think I'll change those back to their original names.

PR #291, opened 2018-03-02
Tab completion

### LINKED ISSUE #274: Fix tab completion for subcommand help
Tab completion works for subcommands and subcommand help works, but tab completion of the subcommand name does not work when using the **help** command.

### LINKED ISSUE #288: Add tab-completion helpers for positional and flag-based completion
Add helper functions in ``cmd2`` to assist with tab-completion in the following scenarios:
- Completion following an ``argparse`` flag
- Completion based on argument index/position

Include an example of how to use these and unit tests.

kmvanbrunt:
This adds flag-based and index-based tab-completion helper functions.

This closes #288.

Tab completion for subcommand help was also added.

This closes #274


tleonhardt:
@kmvanbrunt Before we merge this, I would like to see the following added:
- Example added which demonstrates the usage of the following new helper functions:
    - ``flag_based_complete``
    - ``index_based_complete``
- Multiple unit tests to cover these same two new helper functions
    - See **test_completion.py** for analogous unit test examples

tests/test_completion.py
```diff
@@ -467,52 +465,6 @@ def sc_app():
     return app
 
 
-def test_cmd2_subcommand_completion_single_end(sc_app):
```
tleonhardt: These unit tests for subcommand completion should not be deleted.  They should simply be refactored to use ``index_based_complete()`` or ``complete()`` instead of the old ``complete_subcommand()``.

PR #292, opened 2018-03-02
Removed abbrev attribute

### LINKED ISSUE #289: Remove support for abbreviated commands
The ability to abbreviate commands has always seemed like an anti-feature.  The presence of good tab-completion of command names makes it completely unnecessary to allow for accepting abbreviated command names.  While accepting abbreviated command names can result in users accidentally doing something quite unintentional.  Supporting abbreviated commands significantly complicates other support for several other features and it has outright never worked in combination with multi line commands, which we have documented.

The ``cmd2.Cmd.abbrev`` attribute which supports abbreviated commands should be removed entirely.

Associated documentation, unit tests, and examples will all need to be updated and the CHANGELOG will need to adequately communicate this potentially breaking change.

kmvanbrunt:
Removed _abbrev_ attribute and support for abbreviated commands

This closes #289

### DIFF
--- a/tests/test_cmd2.py
+++ b/tests/test_cmd2.py
@@ -792,6 +792,30 @@ def test_base_py_interactive(base_app):
     m.assert_called_once()
 
 
+def test_exclude_from_history(base_app, monkeypatch):
+    # Mock out the os.system call so we don't actually open an editor
+    m = mock.MagicMock(name='system')
+    monkeypatch.setattr("os.system", m)
+
+    # Run edit command
+    run_cmd(base_app, 'edit')
+
+    # Run history command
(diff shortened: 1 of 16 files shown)


tests/test_cmd2.py
```diff
@@ -1275,49 +1271,6 @@ def test_cmdresult(cmdresult_app):
...
 
 
-@pytest.fixture
-def abbrev_app():
-    app = cmd2.Cmd()
-    app.abbrev = True
-    app.stdout = StdOut()
-    return app
-
-def test_exclude_from_history(abbrev_app, monkeypatch):
```
tleonhardt: We may want to add back a much simpler version of this test

PR #295, opened 2018-03-03
Removed support for case-insensitive command parsing

### LINKED ISSUE #290: Consider removing support for case-insensitive commands
Supporting case-insensitive command names via the ``cmd2.Cmd.case_insensitive`` attribute gets a little complicated.  How should the case-insensitiveness behave for shortucts?  What about sub-command names?

Does anyone actually use and and like support for case-insensitive commands?  If not, I would prefer to remove this "feature".  But if people do use it and like it then we won't touch it.

tleonhardt:
This PR completely removes the ``case_insensitive`` attribute and all support for case-insensitive command-name parsing.  

Thus, from here on out, ``cmd2`` is case sensitive in its parsing of command names.

Since Python is a case-sensitive language and ``cmd2`` is a Python module, it just feels right to keep it simple and make it case sensitive.

An alternative would have been to do this in stages, first making it case sensitive by default but leaving the ``case_insensitive`` attribute around for a release or two, but changing the default value to ``False``.  But I think it is probably better to just rip this band-aid off.

This closes #290


tleonhardt:
@kmvanbrunt Good call!  That call to ``.lower()`` was not a good thing.  It has been removed.

PR #299, opened 2018-03-09
Tweaked complete function to handle cases where a flag appears before…

kmvanbrunt:
… subcommand name

### DIFF
--- a/cmd2.py
+++ b/cmd2.py
@@ -254,34 +254,34 @@ def index_based_complete(text, line, begidx, endidx, index_dict, default_complet
         # Invalid syntax for shlex (Probably due to missing closing quote)
         return []
 
+    if len(tokens) == 0:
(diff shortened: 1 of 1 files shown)


cmd2.py
```diff
@@ -193,7 +193,7 @@ def flag_based_complete(text, line, begidx, endidx, flag_dict, default_completer
         return []
 
     completions = []
-    flag_processed = False
+    flag_present = False
```
tleonhardt: As far as I can tell, these changes look good.  You might consider modifying either the tab_completion.py or subcommands.py example to illustrate the case where this PR would matter.

kmvanbrunt:
I don't yet have an example for those since I've been to sick to do
anything. I will be using this at work though, so an example will exist
sometime next week.

Issue #268, opened 2018-02-09
Improve code coverage of unit tests


tleonhardt:
Over time our code coverage has slowly declined from a high watermark of 97% to a current value around 87%.  This has primarily been do to new features getting added which have incomplete code coverage.

We should all make an effort to increase code coverage for the 0.9.1 release.  This will likely be a tricky and time consuming effort because a lot of the uncovered code is particularly difficult to unit test.

But on the positive side, our unit tests now run much, much more quickly after the migration from ``pyparsing`` to ``shlex``.  So adding more unit/integration tests should be fine from a performance perspective.

tleonhardt:
PR #301 partially addressed this issue, in particular in relation to tab-completion of submenu commands.

tleonhardt:
Large uncovered chunks of code include all of the following:

- **argcomplete_bridge.py** (@anselor)
    - ``CompletionFinder`` class (lines 143-246)
    - ``tokens_for_completion`` function (lines 45-98)
- **argparse_completer.py** (@anselor)
    - ``_format_usage`` method within ``ACHelpFormatter`` class (lines 670-689)
- **cmd2.py** (@tleonhardt)
    - ``ppaged`` function (lines 841-892)
    - ``history -t`` command (lines 3006-3031)

Covering some of this code will likely be difficult, but should be do-able for most of it.

tleonhardt:
@kotfu Thanks for working on the ``history -t`` tests.

tleonhardt:
I'm going to close this issue, we are back up over 90% code coverage.  Lets just try to keep it over 90% and if it dips below again, then we should create a new issue and improve it.

PR #301, opened 2018-03-12
submenu autocomplete

albertored:
For submenu commands autocompletion does not work correctly.

The problem is in calling `completenames` method, this method is overridden from `cmd.Cmd` class:

```python
# cmd
def completenames(self, text, *ignored):
    pass
```

```python
# cmd2
def completenames(self, text, line, begidx, endidx):
    pass
```
But in submenus it is called with only text argument (`cmd_obj.completenames(text)`) generating a `TypeError`.

I also added some completion tests for submenus.

Partially related to #268

### DIFF
--- a/cmd2.py
+++ b/cmd2.py
@@ -790,7 +790,9 @@ def _complete_from_cmd(cmd_obj, text, line, begidx, endidx):
     command_subcommand_params = line.split(None, 3)
 
     if len(command_subcommand_params) < (3 if text else 2):
-        return cmd_obj.completenames(text)
+        n = len(command_subcommand_params[0])
+        n += sum(1 for _ in takewhile(str.isspace, line[n:]))
+        return cmd_obj.completenames(text, line[n:], begidx - n, endidx - n)
 
     command, subcommand = command_subcommand_params[:2]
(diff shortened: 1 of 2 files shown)


cmd2.py
```diff
@@ -790,7 +790,9 @@ def _complete_from_cmd(cmd_obj, text, line, begidx, endidx):
     command_subcommand_params = line.split(None, 3)
 
     if len(command_subcommand_params) < (3 if text else 2):
-        return cmd_obj.completenames(text)
+        n = len(command_subcommand_params[0])
+        n += sum(1 for _ in takewhile(str.isspace, line[n:]))
+        return cmd_obj.completenames(text, line[n:], begidx - n, endidx - n)
```
tleonhardt: Nice catch!

tests/test_completion.py
```diff
@@ -756,9 +756,81 @@ def get_endidx():
 
     assert first_match is None
 
-def test_complete_subcommand_single_end(sc_app):
+
+def test_cmd2_help_subcommand_completion_single_end(sc_app):
```
tleonhardt: Your improvements to these unit tests are sincerely appreciated.  Thanks you!

albertored:
your welcome!

PR #304, opened 2018-03-13
Add option to cancel the current line on keyboard interrupt.

MarkLalor:
When receiving the INT signal (e.g. from Ctrl+C), the behavior of most shells is to cancel the current line of input and provide a fresh prompt. This change would provide an option to specify this behavior by setting a flag (attribute). The default behavior would of course respect backwards compatibility.

As someone new to the code base, I was wondering a few things while exploring.
1. Is the attribute defined in a good spot and is this attribute the best way to add the option of this functionality.
2. Is there another entry point besides _cmdloop where a change might be necessary?
3. Is there a cleaner way to test without the little SayApp that I made?

Also whether or not to print out the "^C" would be a matter of preference. It seems like Bash does this, but Zsh doesn't, for example (with my configurations, at least).

### DIFF
--- a/tests/test_cmd2.py
+++ b/tests/test_cmd2.py
@@ -900,6 +900,47 @@ def test_precmd_hook_failure(hook_failure):
     assert out == True
 
 
+class SayApp(cmd2.Cmd):
+    def __init__(self, *args, **kwargs):
+        # Need to use this older form of invoking super class constructor to support Python 2.x and Python 3.x
+        cmd2.Cmd.__init__(self, *args, **kwargs)
+
+    def do_say(self, arg):
+        self.poutput(arg)
+
+@pytest.fixture
+def say_app():
+    app = SayApp()
+    app.stdout = StdOut()
+    return app
+
+def test_interrupt_quit(say_app):
+    # Mock out the input call so we don't actually wait for a user's response on stdin
+    m = mock.MagicMock(name='input')
+    m.side_effect = ['say hello', KeyboardInterrupt(), 'say goodbye', 'eof']
+    sm.input = m
+
+    say_app.cmdloop()
+
+    # And verify the expected output to stdout
+    out = say_app.stdout.buffer
+    assert out == 'hello\n'
+
+def test_interrupt_noquit(say_app):
+    say_app.quit_on_sigint = False
+
+    # Mock out the input call so we don't actually wait for a user's response on stdin
+    m = mock.MagicMock(name='input')
+    m.side_effect = ['say hello', KeyboardInterrupt(), 'say goodbye', 'eof']
+    sm.input = m
+
+    say_app.cmdloop()
+
+    # And verify the expected output to stdout
+    out = say_app.stdout.buffer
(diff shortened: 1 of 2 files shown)


tleonhardt:
@MarkLalor 
Hi Mark, thanks for the Pull Request!  

First let me attempt to provide a little high level context.  The reason I haven't worried about the current default behavior when hitting Ctrl+C is that I have always implemented a **sigint** handler in all of my ``cmd2`` applications.  But I never wanted to force a default **sigint** handler on end users.  That being said, I think the default behavior of quitting the ``cmd2`` application when Ctrl+C is hit would rarely be desired.

Let me try to answer some of your questions ...

1. It might be slightly better to define the attribute as an instance member inside the ``__init__()`` method, but I think it is fine as it is now as a class attribute.  The only time it would really make a difference is when people are using the SubMenu feature and in that case it is hard to say which alternative would be preferable.  I would even be OK with the attribute having a default value which did not preserve backwards compatibility if @kotfu agrees.

2. As it is, your change can prevent quitting when a ``cmd2`` application is blocking waiting for input from the console.  However, it will NOT prevent quitting while a ``cmd2`` application is executing a command.  In order to do that, a similar **except KeyboardError** statement would needed to be added to the existing exceptions within the ``onecmd_plus_hooks`` method.

3. Your little SayApp seems pretty clear to me.  I think you did a nice job with how the mocks work for that.  Unit testing ``cmd2`` behavior, particularly when it relates to things like ``readline`` command-line input can be challenging.

tests/test_cmd2.py
```diff
@@ -900,6 +900,47 @@ def test_precmd_hook_failure(hook_failure):
...
+@pytest.fixture
+def say_app():
+    app = SayApp()
+    app.stdout = StdOut()
+    return app
+
+def test_interrupt_quit(say_app):
+    # Mock out the input call so we don't actually wait for a user's response on stdin
+    m = mock.MagicMock(name='input')
+    m.side_effect = ['say hello', KeyboardInterrupt(), 'say goodbye', 'eof']
```
tleonhardt: We appreciate that you spent the time to create good unit tests of the expected behavior for both cases.  Thank you!

MarkLalor:
@tleonhardt 
Thanks for answering my questions!

The high-level context makes sense! For me at a high level, the place that made the most sense to implement a **sigint** handler would be inside the loop in the `cmd2` code, since logically, you are not breaking out of the loop in these other shells when you signal an interrupt, so this seemed to make more sense than say, breaking out of the `cmdloop` and then starting the `cmdloop` again. But I also agree that others could benefit from this default behavior since it is more expected and quitting entirely is rarely desired.

As for 2.,

The point about exiting while blocking waiting for a command makes sense. Looking at the code, I don't _see_ this being handled. However, when I actually tested this out by running a long-running command (added a sleep to one of the examples) and then pressing Ctrl+C, the command stopped executing without actually quitting the entire program.

kotfu:
@MarkLalor This is a great add, thanks for the PR.

PR #305, opened 2018-03-13
AddSubMenu should work with default kwargs

albertored:
The `AddSubMenu` class decorator should work also with default kwargs.

Before this PR if one tries to use it without `shared_attributes` and without setting `require_predefined_shares` to `False` the code crashes because it tries to access the keys of a `None` object (`shared_attributes` default kwarg)

### DIFF
--- a/cmd2.py
+++ b/cmd2.py
@@ -867,14 +867,14 @@ class AddSubmenu(object):
             raise ValueError("reformat_prompt should be either a format string or None")
         self.reformat_prompt = reformat_prompt
 
+        self.shared_attributes = {} if shared_attributes is None else shared_attributes
         if require_predefined_shares:
-            for attr in shared_attributes.keys():
+            for attr in self.shared_attributes.keys():
                 if not hasattr(submenu, attr):
                     raise AttributeError("The shared attribute '{attr}' is not defined in {cmd}. Either define {attr} "
                                          "in {cmd} or set require_predefined_shares=False."
                                          .format(cmd=submenu.__class__.__name__, attr=attr))
 
-        self.shared_attributes = {} if shared_attributes is None else shared_attributes
         self.create_subclass = create_subclass
 
     def __call__(self, cmd_obj):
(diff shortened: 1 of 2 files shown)


cmd2.py
```diff
@@ -830,7 +830,7 @@ def __init__(self,
                  command,
                  aliases=(),
                  reformat_prompt="{super_prompt}>> {sub_prompt}",
-                 shared_attributes=None,
+                 shared_attributes={},
```
tleonhardt: Mutable default arguments are almost always a bad thing in common, they behave in a rather odd manner analogous to static function-scope variables in C:
http://docs.python-guide.org/en/latest/writing/gotchas/

The previous code was attempting to set ``self.shared_attributes`` to an empty dict if the argument was None or to whatever it was supplied as if it isn't None.

Good catch on finding this bug!  But I think a different fix is likely in order.  

What would you think about simply moving the line which sets ``self.shared_attributes`` earlier to prevent attempting to get keys on a None, like so:
```Python
self.shared_attributes = {} if shared_attributes is None else shared_attributes
        if require_predefined_shares:
            for attr in shared_attributes.keys():
                if not hasattr(submenu, attr):
                    raise AttributeError("The shared attribute '{attr}' is not defined in {cmd}. Either define {attr} "
                                         "in {cmd} or set require_predefined_shares=False."
                                         .format(cmd=submenu.__class__.__name__, attr=attr))
```

albertored: yes, you're right, I'll push the change :+1:

cmd2.py
```diff
@@ -867,14 +867,14 @@ def __init__(self,
             raise ValueError("reformat_prompt should be either a format string or None")
         self.reformat_prompt = reformat_prompt
 
+        self.shared_attributes = {} if shared_attributes is None else shared_attributes
         if require_predefined_shares:
-            for attr in shared_attributes.keys():
+            for attr in self.shared_attributes.keys():
```
tleonhardt: Nice.  I think this fix looks good.

Issue #303, opened 2018-03-12
SubMenu `shared_attributes` behaviour


albertored:
The current behaviour for `shared_attributes` in submenus is to copy attributes from parent on entering command loop and then reset original child attributes on exit loop.

Is there a reason for this?

I was expecting to have shared attributes modified on child to be copied back to parent on child loop exit in order to preserve the change.

What do you think? Maybe this behaviour can be conditioned to some arguments of init method of `AddSubMenu`?

tleonhardt:
Honestly I haven't used the submenu feature yet and I'm not sure if there is a good reason for why the ``shared_attributes`` behave like they do.  My guess would be that it met the needs of the people who designed this feature.

@lobocv Is one of the folks who designed this feature.  Calvin, can you speak to why it is implemented like that?

Regardless, I think it would be reasonable to condition the behavior on an argument to the init method of ``AddSubMenu``, particularly if the default behavior is like the current one.  Please feel free to submit a PR to add this if it would be useful for you.  

And if you do that and feel up to the task, if you wanted to add some more unit tests related to the submenu feature that would be awesome ;-)  Issue #268 was previously created for that purpose.

albertored:
Ok, it sounds good to me.

Just to add more info I explain the use case I'm working on.

Suppose you have a main `Cmd` class with a submenu. I would like to be able to set some options (like `debug` and `colors`) globally in my application so I put them in the `shared_attributes` dict argument of `AddSubMenu`. In this way if a user changes one option in the main app and then it enters the submenu the option is correctly set also there. The problem arises in doing the opposite, i.e. setting the option in the submenu and then going back to main app.

lobocv:
I'm fine with changing the behaviour to make it apply the changes to the parent. That actually seems logical otherwise it's basically read-only. Care will have to be made for anything non-mutable such as lists.

tleonhardt:
I believe this issue was closed by PR #306.

@albertored Feel free to re-open this issue if there was more work you want to do with it.

PR #306, opened 2018-03-13
WIP: add `preserve_shares` arg to AddSubMenu

albertored:
**Work in progress PR**

This is a first implementation of what discussed in #303 

At the moment I simply added the `preserve_shares` arg to `AddSubMenu` constructor to copy back to the parent the updated shared attributes of child (default `False`).

I would like to have some feedback from you before adding comments and tests.

Do you think is worth changing this flag to an array of attributes for allow copying back only some of them or it's better to leave it as a global thing?

I'm not sure on the naming of the new arg, suggestion are appreciated

@tleonhardt @lobocv

### DIFF
--- a/cmd2.py
+++ b/cmd2.py
@@ -832,6 +832,7 @@ class AddSubmenu(object):
                  reformat_prompt="{super_prompt}>> {sub_prompt}",
                  shared_attributes=None,
                  require_predefined_shares=True,
+                 preserve_shares=False,
                  create_subclass=False
                  ):
         """Set up the class decorator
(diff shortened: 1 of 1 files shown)


tleonhardt:
@albertored 
Hi Alberto.  Thanks for the WIP PR!

I think that leaving the flag as global should suffice in the vast majority of cases.  If you wanted to you could make it True/False by default but also allow a user to set it to an Iterable of attributes, but then you would need to detect the type and behave differently in the two cases.  I would say keep it simple unless you yourself have a good reason to make it otherwise for your intended use case.

The naming of the new arg seems fine to me.

@kotfu or @lobocv Do you have any opinions in regards to this?

tleonhardt:
@albertored I merged the latest stuff from master into your PR branch

lobocv:
Looks good. Thanks for the PR. I agree to leave it global for now. I see having it read-write being the default scenario and if people would like read-only attributes they could limit access via properties.

albertored:
Tests added

PR #307, opened 2018-03-14
Added documentation for quit_on_sigint

MarkLalor:
Documentation addition for the change in #304.

Tried to keep it short and sweet for a simple option like this!


tleonhardt:
Thank you for adding documentation to the feature you added in the previous PR!

PR #308, opened 2018-03-14
allow submenu history preservation

albertored:
I added the possibility of having history preservation in a submenu.

When `AddSubMenu` is initialized with the `persistent_history_file` arg the logic is the following:
1. on entering the submenu the current history (from main app) is cleared and saved in `history` variable
2. history is read from the given  `persistent_history_file`
3. on exit the submenu history is written on  `persistent_history_file`
4. the main app saved history is pushed back in main app

Points 1 and 4 are exactly the same of before, I just changed their position in the code.
Points 2 and 3 are not performed if  `persistent_history_file` is not given.

**Open points / caveats**
* `persistent_history_file` for submenus  is an argument of `AddSubMenu` and not of `Cmd`, it may lead to confusion
* if  `persistent_history_file` is the same of main app then history on submenu is the one of main app, maybe we can add a check but still we will not be able to check that is different from all others submenus. Maybe is better to leave it as user responsibility
* I don't know how to test it programatically but I tried it in person and it seems to work well

### DIFF
--- a/cmd2.py
+++ b/cmd2.py
@@ -832,8 +832,9 @@ class AddSubmenu(object):
                  reformat_prompt="{super_prompt}>> {sub_prompt}",
                  shared_attributes=None,
                  require_predefined_shares=True,
+                 create_subclass=False,
                  preserve_shares=False,
-                 create_subclass=False
+                 persistent_history_file=None
                  ):
         """Set up the class decorator
 
@@ -863,6 +864,10 @@ class AddSubmenu(object):
         self.submenu = submenu
         self.command = command
         self.aliases = aliases
+        if persistent_history_file:
+            self.persistent_history_file = os.path.expanduser(persistent_history_file)
+        else:
+            self.persistent_history_file = None
 
         if reformat_prompt is not None and not isinstance(reformat_prompt, str):
             raise ValueError("reformat_prompt should be either a format string or None")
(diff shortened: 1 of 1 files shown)


tleonhardt:
@albertored 
Thanks for the PR!  I like the possibility of preserving history separately for the different submenus.

Your open points and caveats don't concern me.  I'm fine with making it user responsibility to make sure things don't conflict in their application.

Could you merge in the latest changes from master.  This PR conflicts slightly with your previous PR #306 .  I could resolve the conflicts, but I would prefer you to do it in a way you are happy with.

albertored:
rebased :+1:

cmd2.py
```diff
@@ -832,8 +832,9 @@ def __init__(self,
                  reformat_prompt="{super_prompt}>> {sub_prompt}",
                  shared_attributes=None,
                  require_predefined_shares=True,
+                 create_subclass=False,
```
tleonhardt: Looks good.  Thanks for resolving merge conflicts!

tleonhardt:
@albertored I invited you to join the cmd2 contributor's team.  That will allow you to create branches directly within the main repository so you don't need to work out of a fork.  This makes it easier to use the integrated CI tools and to collaborate with others on the project.

PR #309, opened 2018-03-14
Fix/add completion for shell commands when default_to_shell is True

MarkLalor:
Currently, path completion works for shell commands when they are run in either the `shell <name>` form or in the `!<name>` form, but not in the plain `<name>` form when `default_to_shell` is set to `True`

This change would allow for path completion for a command if `default_to_shell` is `True` and the currently-typed command is in the users's path (as determined by the _get_exes_in_path method)

One thing I was wondering was if there was an effective difference between:
`compfunc = functools.partial(path_complete)` and `compfunc = path_complete`

I saw the former in other parts of the source, so I stuck to that choice for this change, but I was curious if there was a reason for this.

### DIFF
--- a/tests/test_completion.py
+++ b/tests/test_completion.py
@@ -209,6 +209,45 @@ def test_shell_command_completion(cmd2_app):
     begidx = endidx - len(text)
     assert cmd2_app.complete_shell(text, line, begidx, endidx) == expected
 
+def test_default_to_shell_command_completion(request):
+    app = cmd2.Cmd()
+    app.default_to_shell = True
+
+    test_dir = os.path.dirname(request.module.__file__)
+
+    text = 'c'
+    state = 0
+    path = os.path.join(test_dir, text)
+
+    if sys.platform == "win32":
+        command = 'calc.exe'
+    else:
+        command = 'egrep'
+
+    # Make sure the command is on the testing system
+    assert command in cmd2.Cmd._get_exes_in_path(command, False)
+    line = '{} {}'.format(command, path)
+
+    endidx = len(line)
+    begidx = endidx - len(text)
+
+    def get_line():
+        return line
+
+    def get_begidx():
+        return begidx
+
+    def get_endidx():
+        return endidx
+
+    with mock.patch.object(readline, 'get_line_buffer', get_line):
+        with mock.patch.object(readline, 'get_begidx', get_begidx):
+            with mock.patch.object(readline, 'get_endidx', get_endidx):
+                # Run the readline tab-completion function with readline mocks in place
+                first_match = app.complete(text, state)
+
+    assert first_match is not None and app.completion_matches == ['conftest.py ']
+
 def test_shell_command_completion_doesnt_match_wildcards(cmd2_app):
     if sys.platform == "win32":
         text = 'c*'
--- a/cmd2.py
+++ b/cmd2.py
@@ -1357,12 +1357,14 @@ class Cmd(cmd.Cmd):
                 if command == '':
                     compfunc = self.completedefault
                 else:
-
                     # Get the completion function for this command
                     try:
                         compfunc = getattr(self, 'complete_' + command)
                     except AttributeError:
-                        compfunc = self.completedefault
+                        if self.default_to_shell and command in self._get_exes_in_path(command, False):
+                            compfunc = functools.partial(path_complete)
+                        else:
+                            compfunc = self.completedefault
 
                     # If there are subcommands, then try completing those if the cursor is in
                     # index 1 of the command tokens


tests/test_completion.py
```diff
@@ -209,6 +209,40 @@ def test_shell_command_completion(cmd2_app):
     begidx = endidx - len(text)
     assert cmd2_app.complete_shell(text, line, begidx, endidx) == expected
 
+def test_default_to_shell_command_completion(request):
```
tleonhardt: This test is failing for Python 2.7 on Linux and for all versions of Python on Windows.

tleonhardt:
@MarkLalor Thanks for the PR!  This change makes a lot of sense.

I believe the reason we are using ``functools`` to do the binding to the ``path_complete`` function is that we are binding to a function but are effectively creating a method.  

Thank you for adding a new unit test to test the enhancement you added.  It currently has the following failures:

- Python 2.7 on all OSes:
https://travis-ci.org/python-cmd2/cmd2/jobs/353457302
- Python 3 on Windows:
https://ci.appveyor.com/project/FedericoCeratto/cmd2/build/1.0.984

Once you fix these unit test failures, this PR should be good to go.

Thanks again for another good PR.

tleonhardt:
I don't think it should effect you, but I merged master into your branch just to be safe due to recent PRs merging in.

MarkLalor:
Fixed the Python 3 problems, but the problem with Windows is still rather unclear. I added an extra check to double-check that the program is on the path, and it doesn't seem that was the problem. I won't have access to a Windows machine for debugging until next week so I'm not sure I'll be able to realistically debug until then since I can't really tell the details from the failures.

I have a _feeling_ it may have to do with my begidx value though, I'm still getting familiar with the details of readline and tab completion.

tleonhardt:
Mark, thanks for fixing the issue on Python 2.  I will try to take a look at the issue on Windows later.  Or maybe if @kmvanbrunt has time he will since he is our resident readline tab-completion expert.

If you ever just need to setup a Windows VM for free you can use Microsoft Azure totally free for 30 days and some services stay free indefinitely:
https://azure.microsoft.com/en-us/

Also students can get long-term free access to Microsoft Azure:
https://imagine.microsoft.com/en-us/Catalog/Product/99

kmvanbrunt:
In an effort to further simplify cmd2, I actually recommend removing the **default_to_shell** flag.

Due to recent additions to cmd2, a user is able to tab complete both shell commands and their paths by using either `shell <name>` or `!<name>`.

Setting **default_to_shell** to _True_, offers the user a savings of 1 character (!) and does not tab complete shell commands anyway.

Keeping an outdated flag around results in pull requests like this due to maintenance complexity, since developers have to be aware of yet another setting whenever a change is made.

@tleonhardt @kotfu Thoughts?

MarkLalor:
@kmvanbrunt I thought I should share what I really like about the default_to_shell flag. The cmd2 app I've been working on involves interacting with a remote API as well as the local filesystem a lot. I wrote a `do_cd` command to call `os.chdir`, and added a call to `os.getcwd()` when my prompt is generated. Users can type cd, ls, cat, tail, etc, just as they can in a normal shell.

I think it's nice that users do not have to retrain their muscle memory to add an `!` to these extremely common commands. I understand that one can reimplement this behavior in their own application instead of inside of cmd2, but just how *easy* it was to get this behavior is one of the reasons I chose cmd2 over others. The setup it took to get a shell with the features I wanted was so minimal.

kmvanbrunt:
@MarkLalor Look for an upcoming **alias** command that will make your job easier.
With aliases, you won't have to create a `do_<command>` for each shell command you want to wrap.
Plus aliases can be loaded from a file when the CLI starts.

Basically you'll do something like:
alias cd='!cd'
alias pwd='!pwd'

Aliases will also be tab completable, which **default_to_shell** doesn't offer.

kmvanbrunt:
@MarkLalor Will aliases be OK for what you need or would you like the **default_to_shell** flag preserved?
If so, then there are a few tweaks I would like to make. I will probably just take your changes and add mine on another branch.

On tweak in particular, is the `default()` method is calling any unrecognized commands as a shell commands if this flag is true. I would rather confirm its in the path before just blindly calling out to the shell.

I also have a tweak for `_get_exes_in_path()`, since it was only meant to be a worker function for `complete_shell()`.

tleonhardt:
@kmvanbrunt @MarkLalor 
I think you both have a valid perspective and points in regards to the ``default_to_shell`` flag.  In summary:

- Advantage:  It makes it extremely easy for developers to create an application based on ``cmd2`` which contains shell-like functionality
- Disadvantage:  It complicates the maintenance of ``cmd2``, perhaps unnecessarily

My opinion is that it is probably best to move forward with this PR containing improvements to the existing feature and preserve backward compatibility at least for now. 

Once the ``alias`` feature is complete and we have all had time to experiment with it, then maybe it would be good to re-open the discussion about the ``default_to_shell`` flag.

MarkLalor:
@kmvanbrunt I think it makes sense to actually check the path before blinding calling out to the shell, and yes I did notice that `_get_exes_in_path` was definitely targeted towards completion, but a subset of it's behavior was good for the job so I figured reusing it was fine. A way to make this usage cleaner would be nice though.

I'm not sure if aliases would feel much like a full replacement, unless I wrote at the start up of my program to patch aliases for every shell command in the path. If I just created aliases for a few of the common shell commands, then I feel like a user would be confused why they have to use an '!' suddenly for more obscure commands.

I think @tleonhardt summed it up pretty well though, and that the topic should definitely be revisited later.

tleonhardt:
There would be one disadvantage to checking the path before calling a command in the shell.  At least for certain operating systems and shells, some commands are built into the shell itself and don't exist anywhere on the path.  For example, in the Bash shell on Linux, commands like **cd** and **pwd** are built into the shell and don't exist anywhere on the path.

tleonhardt:
@MarkLalor I invited you to become a ``cmd2`` contributor.  This will give you the ability to create branches within the main cmd2 repo which will make it dramatically easier for other contributors to collaborate with you in the future.

MarkLalor:
@tleonhardt thanks for that, sounds good!

I've been trying to think about how to handle the shell builtins, and I'm not entirely sure. I noticed while looking around that it seems most of the really commonly-used Bash builtins happen to also have executables in the path macOS and Ubuntu (that are presumably shadowed by the shell implementation). But still a few other potentially useful commands like "wait" seem to be built in to the shell (not that you would want to tab-complete wait for a file anyways...)

Also, shell built-ins that do not have corresponding executables are often be tied to the state of the shell itself, which cmd2 doesn't seem to maintain (right?). For example jobs, fg, bg can't really be used.

It seems like there's a limited number of concrete options on how to handle this:

1. Use `type` on macOS/Linux and investigate a Windows solution to cover all the bases.
2. Create a set of common builtins that should be tab-completable and check if the command is in the path *or* this set.
3. Ignore the shell builtins because it should be "good enough" since there are overlapping executables in the path.
(then of course things like abandoning default_to_shell completion altogether)

I think 1 would be a bit too hacky and could blow up if everything was attempted to be covered (I think using shell=True uses the shell from the `SHELL` environment variable, so if they had a shell that doesn't support the `type` command... well...

Actually, the above statement made me realize, that trying to detect shell built-ins would pretty much be an undecidable problem since we presumably have no idea what shell is going parse the commands.

I think I'll stop typing this essay because I'm not sure exactly where I'm going with this. TL;DR I think it becomes a headache when thinking about builtins and that the consequences of ignoring them would be minimal.

MarkLalor:
By the way, since this is reimplemented basically at this point on the default_on_shell_unit_tests  branch does it seem fine to just close this PR and continue this discussion elsewhere if needed?

tleonhardt:
Ok, I'm going to close this since it got re-implented on another branch. 

@MarkLalor I apologize that your good work essentially got re-implemented in another branch.  We appreciate your contribution.

On the plus side, I do believe @kmvanbrunt figured out how to make a small change in ``cmd2`` that enabled the feature to work better on Windows.

PR #311, opened 2018-03-15
remove history line in select when user insert choice

albertored:
Without this change the history get polluted of user select choiches

### DIFF
--- a/cmd2.py
+++ b/cmd2.py
@@ -2046,6 +2046,9 @@ class Cmd(cmd.Cmd):
             self.poutput('  %2d. %s\n' % (idx + 1, text))
         while True:
             response = sm.input(prompt)
+            hlen = readline.get_current_history_length()
+            if hlen >= 1 and response != '':
+                readline.remove_history_item(hlen - 1)
             try:
                 response = int(response)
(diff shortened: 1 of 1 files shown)


tleonhardt:
Weird, it doesn't seem like the code you added should have had any effect on the existing unit tests one way or the other!

albertored:
Yes, it's strange.. I'll take a look later

albertored:
Now it should work, in tests history is always empty so I was trying to remove an element at index -1 and this raise a `ValueError` that is caught by the except statement in `onecmd_plus_hooks` method.

cmd2.py
```diff
@@ -2046,6 +2046,9 @@ def select(self, opts, prompt='Your choice? '):
             self.poutput('  %2d. %s\n' % (idx + 1, text))
         while True:
             response = sm.input(prompt)
+            hlen = readline.get_current_history_length()
+            if hlen >= 1 and response != '':
+                readline.remove_history_item(hlen - 1)
```
tleonhardt: Looks good.  Thanks for this fix.  Not polluting the readline history when using ``select()`` is a nice fix.

PR #312, opened 2018-03-15
Moresub

kmvanbrunt:
Added code to support tab-completion in subcommand routines.

### DIFF
--- a/cmd2.py
+++ b/cmd2.py
@@ -1369,7 +1369,6 @@ class Cmd(cmd.Cmd):
                 # Overwrite line to pass into completers
                 line = expanded_line
 
-                got_matches = False
                 if command == '':
                     compfunc = self.completedefault
                 else:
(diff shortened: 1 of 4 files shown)


cmd2.py
```diff
@@ -512,32 +499,34 @@ def cmd_wrapper(instance, cmdline):
...
+            subcommand_names = argparser._subparsers._group_actions[0]._name_parser_map.keys()
             cmd_wrapper.__dict__['subcommand_names'] = subcommand_names
 
         return cmd_wrapper
 
     return arg_decorator
 
 
-def with_argparser(argparser, subcommand_names=None):
+def with_argparser(argparser):
```
tleonhardt: Not requiring the 2nd argument to these decorators is a very nice improvement.

cmd2.py
```diff
@@ -549,17 +538,18 @@ def cmd_wrapper(instance, cmdline):
         # we want it to be the name of our command
         argparser.prog = func.__name__[3:]
 
-        # put the help message in the method docstring
-        if func.__doc__:
+        # If the description has not been set, then use the method docstring if one exists
```
tleonhardt: So an argparse description will now override the docstring?  That sounds OK with me.  You might want to update the Sphinx documentation and/or examples to show this behavior.

cmd2.py
```diff
@@ -549,17 +538,18 @@ def cmd_wrapper(instance, cmdline):
...
 
         cmd_wrapper.__doc__ = argparser.format_help()
 
         # Mark this function as having an argparse ArgumentParser (used by do_help)
         cmd_wrapper.__dict__['has_parser'] = True
 
-        # If there are subcommands, store their names to support tab-completion of subcommand names
-        if subcommand_names is not None:
+        # If there are subcommands, store their names in a list to support tab-completion of subcommand names
+        if argparser._subparsers is not None:
```
tleonhardt: It would be nice if ``argparse`` made this easier, but neither @kotfu or I could find a good way.  So thanks for this.

cmd2.py
```diff
@@ -2236,6 +2218,90 @@ def complete_shell(self, text, line, begidx, endidx):
         else:
             return path_complete(text, line, begidx, endidx)
 
+    def cmd_with_subs_completer(self, text, line, begidx, endidx, base):
```
tleonhardt: This is a very useful and convenient bit of code.  Thanks for the good work.

Now what I would love like to see in a follow-on PR is the following:
- unit tests covering this function
- update of [subcommands.py](https://github.com/python-cmd2/cmd2/blob/master/examples/subcommands.py) and/or [tab_completion.py](https://github.com/python-cmd2/cmd2/blob/master/examples/tab_completion.py) example to make use of this
- update of Sphinx documentation [sub-commands section](http://cmd2.readthedocs.io/en/latest/argument_processing.html?#sub-commands)

PR #314, opened 2018-03-16
Alias

### LINKED ISSUE #277: Consider adding synonym capabilities
It would be nice if you could teach cmd2 about synonyms for commands.

Say for example I create a command called **config** which shows the contents of a configuration file. I would like to make that same functionality available in a command called **settings**. I can make that happen by creating a **do_settings()** method which calls **do_config()**, but I have to remember to handle the help messages. This gets more complicated if you are using either of the argparser decorators because you may have to create two parsers, identical except for the **prog** parameter. The other sub-optimal thing is how the **help** command behaves. It shows both commands, but there is no indication that they are synonyms.

It would be great if you could do something like:
```
cmd2.synonyms = { 'config': ['settings', 'my_other_synonym']}
```
and have cmd2 do all the work: make the synonym command names work, with necessary changes to argparser **prog** parameters, and then display the help in such a way that it's easy to see which commands are synonyms for what.

kmvanbrunt:
Added `alias` and `unalias` commands.

Closes #277


cmd2.py
```diff
@@ -1922,6 +1946,92 @@ def _cmdloop(self):
 
             return stop
 
+    @with_argument_list
+    def do_alias(self, arglist):
```
tleonhardt: This looks pretty clean and similar to Bash, so it should probably be self-explanatory to many people.

docs/settingchanges.rst
```diff
@@ -41,12 +41,28 @@ To define more shortcuts, update the dict ``App.shortcuts`` with the
...
 .. warning::
 
-  Command aliases needed to be created by updating the ``shortcuts`` dictionary attribute prior to calling the
+  Shortcuts need to be created by updating the ``shortcuts`` dictionary attribute prior to calling the
   ``cmd2.Cmd`` super class ``__init__()`` method.  Moreover, that super class init method needs to be called after
   updating the ``shortcuts`` attribute  This warning applies in general to many other attributes which are not
   settable at runtime such as ``commentGrammars``, ``multilineCommands``, etc.
 
 
+Aliases
```
tleonhardt: Thanks for updating the documentation.

tests/test_cmd2.py
```diff
@@ -1541,3 +1542,46 @@ def test_poutput_none(base_app):
     out = base_app.stdout.buffer
     expected = ''
     assert out == expected
+
+
+def test_alias(base_app, capsys):
```
tleonhardt: Nice job adding unit tests for new features.

tleonhardt:
This PR looks good.  I think it is a nice feature which can allow end users to customize their usage of a given ``cmd2`` application, particularly if they load a script at startup.

Random train of thought, sometime soonish we may want to put some effort into getting our code coverage back up.

PR #315, opened 2018-03-16
Added tab completion for shell commands when default_to_shell is True

kmvanbrunt:


cmd2.py
```diff
@@ -1399,7 +1399,10 @@ def complete(self, text, state):
                     try:
                         compfunc = getattr(self, 'complete_' + command)
                     except AttributeError:
-                        compfunc = self.completedefault
+                        if self.default_to_shell and command in self._get_exes_in_path(command):
```
tleonhardt: This is similar to PR #309.

cmd2.py
```diff
@@ -2241,13 +2244,11 @@ def do_shell(self, command):
         proc.communicate()
 
     @staticmethod
-    def _get_exes_in_path(starts_with, at_eol):
+    def _get_exes_in_path(starts_with):
```
tleonhardt: But it looks like you are making some changes which are intended to help with that PR.  So I guess we can merge this in and then merge master into that PR to pick up the unit tests in that PR (once they are fixed).

PR #318, opened 2018-03-17
Default on shell unit tests

kmvanbrunt:


cmd2.py
```diff
@@ -1578,14 +1578,18 @@ def parseline(self, line):
...
-        while i < n and line[i] in self.identchars:
-            i += 1
-        command, arg = line[:i], line[i:].strip()
 
-        # Make sure there is a space between the command and args
-        # This can occur when a character not in self.identchars bumps against the command (ex: help@)
-        if len(command) > 0 and len(arg) > 0 and line[len(command)] != ' ':
-            line = line.replace(command, command + ' ', 1)
+        # If we are allowing shell commands, then allow any character in the command
+        if self.default_to_shell:
```
tleonhardt: Are you comfortable that there won't be negative unforeseen consequences due to this change?

kmvanbrunt: I restored the behavior of cmd's version so it should be fine.
I only added this code recently and really it doesn't make sense for a user to type a symbol in the middle of a command anyway.

tleonhardt:
@MarkLalor Thanks for creating the good unit test which exposed an underlying issue in ``cmd2`` on Windows!

PR #319, opened 2018-03-17
Added ability to specify a startup_script in cmd2.Cmd.__init__()

### LINKED ISSUE #317: Add option to specify a startup script in the cmd2.Cmd.__init__()
One thing which would make the new **alias** command dramatically more useful would be the ability to specify a script to automatically **load** at startup.

This script could run any arbitrary set of commands contained within the given ``cmd2`` application.  But one thing it would be particularly suitable for is allowing different end users to personalize their environment with custom aliases.

tleonhardt:
This fixes #317.

This is intended primarily to make using the new **alias** command more powerful, but has a much broader use case in that it allows individual users to customize their initialization of a ``cmd2`` application and to always run a set of commands at startup.


tleonhardt:
@kmvanbrunt @MarkLalor @kotfu 
I feel like this is a good feature, but if anyone has any recommendations on how to better document it, it would be appreciated.

kotfu:
I just tweaked the language in README, but the rest of this looks good to me.

MarkLalor:
Oh dang I was a few minutes late, but I noticed a few things, I'll open a separate PR.

PR #321, opened 2018-03-18
Fixed indexes in unit tests

kmvanbrunt:


tests/test_completion.py
```diff
@@ -247,7 +247,7 @@ def test_shell_command_completion_doesnt_complete_when_just_shell(cmd2_app):
     line = 'shell'
 
     endidx = len(line)
-    begidx = endidx - len(text)
+    begidx = 0
```
tleonhardt: Nice catch for a few unit tests which were passing but for the wrong reasons!

PR #323, opened 2018-03-19
Added default sigint handler

tleonhardt:
This is an attempt to improve the out-of-the-box way in which Ctrl-C is handled.

Added a default signal handler for SIGINT that does the following:
- Terminates a pipe process if one exists
- Raises a KeyboardInterrupt for other parts of the code to catch

Also:
- Changed the default value for quit_on_sigint to False
- Modified the way the subcommand functions are called in subcommand.py (unrelated to rest of PR)

### DIFF
--- a/cmd2.py
+++ b/cmd2.py
@@ -36,6 +36,7 @@ import os
 import platform
 import re
 import shlex
+import signal
 import six
 import sys
 import tempfile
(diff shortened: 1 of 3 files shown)


cmd2.py
```diff
@@ -1480,6 +1481,28 @@ def complete_help(self, text, line, begidx, endidx):
...
+        :param signum: int - signal number
+        :param frame
+        """
+        # Save copy of pipe_proc since it could theoretically change while this is running
+        pipe_proc = self.pipe_proc
+        if pipe_proc is not None:
+            pipe_proc.terminate()
+
+        # Re-raise a KeyboardInterrupt so other parts of the code can catch it
+        raise KeyboardInterrupt("Got a keyboard interrupt within a Python script")
```
kmvanbrunt: Is this supposed to check the self._in_py flag before raising the exception?

tleonhardt: Initially I made this method much more complicated where it had a couple checks, one for the flag you mention and onother for the case where we were blocking waiting for input from the user.  But then I thought about it some more and I think it is safe to always re-raise the KeyboardInterrupt.

I manually treated on both macOS and Windows and was happy with the behavior in all cases that I thought of testing.

PR #328, opened 2018-03-24
Allow quoted file paths when redirecting with < and >

kmvanbrunt:
< and > did not handle quoted file paths
@tleonhardt 
@kotfu Make sure to handle this case in the parsing improvements you are making

### DIFF
--- a/cmd2.py
+++ b/cmd2.py
@@ -748,7 +748,9 @@ def replace_with_file_contents(fname):
     :return: str - contents of file "fname"
     """
     try:
-        with open(os.path.expanduser(fname[0])) as source_file:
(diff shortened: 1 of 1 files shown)


cmd2.py
```diff
@@ -748,7 +748,9 @@ def replace_with_file_contents(fname):
     :return: str - contents of file "fname"
     """
     try:
-        with open(os.path.expanduser(fname[0])) as source_file:
+        # Any outer quotes are not part of the filename
```
tleonhardt: Nice little fix.  It would be undesirable to create a file name with quotes in it.  

The new behavior is much better.

PR #329, opened 2018-03-26
New quoted completion

kmvanbrunt:
Tab completing tokens with spaces now works.
If a completion has a space and no opening quote, then an opening quote will be written to the screen.
Closing quotes are added if completion type allows it.
Added more interaction with the readline library that Python wrapper does not offer.
Added tab completion after redirection strings (|, >, >>, <) for every command.
Moved all tab completion functions into the cmd2 class since they now depend on allow_redirection flag.

### DIFF
--- a/cmd2.py
+++ b/cmd2.py
@@ -584,7 +584,7 @@ def path_complete(text, line, begidx, endidx, dir_exe_only=False, dir_only=False
             # This is a directory, so don't add a space or quote
             set_allow_appended_space(False)
             set_allow_closing_quote(False)
-            return ["~" + os.path.sep]
+            return [completion_token + os.path.sep]
 
         elif completion_token.startswith('~'):
             # Tilde without separator between path is invalid
--- a/cmd2.py
+++ b/cmd2.py
@@ -1930,15 +1930,7 @@ class Cmd(cmd.Cmd):
                             return self.completion_matches[state]
 
                 # Check if a valid command was entered
-                if command not in self.get_all_commands():
-                    # Check if this command should be run as a shell command
-                    if self.default_to_shell and command in get_exes_in_path(command):
-                        compfunc = functools.partial(path_complete)
-                    else:
-                        compfunc = self.completedefault
-
-                # A valid command was entered
-                else:
+                if command in self.get_all_commands():
                     # Get the completer function for this command
                     try:
                         compfunc = getattr(self, 'complete_' + command)
@@ -1953,6 +1945,14 @@ class Cmd(cmd.Cmd):
                         compfunc = functools.partial(index_based_complete,
                                                      index_dict=index_dict,
                                                      all_else=compfunc)
+
+                # A valid command was not entered
+                else:
+                    # Check if this command should be run as a shell command
+                    if self.default_to_shell and command in get_exes_in_path(command):
+                        compfunc = functools.partial(path_complete)
+                    else:
+                        compfunc = self.completedefault
 
                 # Call the completer function
                 self.completion_matches = compfunc(text, line, begidx, endidx)
--- a/cmd2.py
+++ b/cmd2.py
@@ -134,14 +134,13 @@ except ImportError:
         pass
 
 
-# Tells what implementation of readline we are using
+# Check what implementation of readline we are using
 class RlType(Enum):
     GNU = 1
     PYREADLINE = 2
     NONE = 3
 
 
-# Check what implementation of readline we are using
 rl_type = RlType.NONE
 
 if 'pyreadline' in sys.modules:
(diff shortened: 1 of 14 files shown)


tleonhardt:
I found a segfault doing a little manual testing.  I did the following on macOS with gnureadline installed:

```Bash
./cmd2.py
(Cmd) !l<Tab><Tab>
```

The result was:  Segmentation fault: 11

The same behavior occurred with both Python 3.6 and Python 2.7.

tleonhardt:
When I got rid of the **gnureadline** module to force use of readline (as installed by Anaconda's ``conda`` package manager), then the segfault didn't occur and it worked as expected.

kmvanbrunt:
@tleonhardt What do we do about the gnureadline segfault?
Do you know what version of readline that is statically linked against?
I am clueless when it comes to the Mac side of this.

tleonhardt:
The current version of gnureadline links against version 6.3 of GNU Readline.  

If the bug is on our end then we should fix it.  If the bug is in gnureadline, then we should remove the use of gnureadline.

We might need to run with gdb/lldb on a Mac to figure out the cause.

tleonhardt:
@kmvanbrunt Hey Kevin, nice job fixing the problems with gnureadline!

cmd2.py
```diff
@@ -43,10 +44,38 @@
 import traceback
 import unittest
 from code import InteractiveConsole
+from enum import Enum
```
tleonhardt: The ``enum`` module was added in Python 3.4.  For Python 2.7, we either need to not use ``enum`` or add a conditional dependency on the ``enum34`` backport like so:

```Python
try:
    from enum34 import Enum
except ImportError:
    from enum import Enum
```

cmd2.py
```diff
@@ -100,6 +129,44 @@
     except ImportError:
         pass
 
+
+# Tells what implementation of readline we are using
+class RlType(Enum):
```
tleonhardt: Need to make some sort of change so that this works on Python 2.7.  See the comment above in the imports regarding the ``enum`` module.

cmd2.py
```diff
@@ -1410,90 +2015,202 @@ def complete(self, text, state):
...
+                # Add a closing quote if needed and allowed
+                if self.allow_closing_quote and unclosed_quote:
+                    str_to_append += unclosed_quote
 
-                self.completion_matches = completions
+                # If we are at the end of the line, then add a space if allowed
+                if self.allow_appended_space and endidx == len(line):
+                    str_to_append += ' '
+
+                self.completion_matches[0] = self.completion_matches[0] + str_to_append
```
tleonhardt: Recommend using **+=**

cmd2.py
```diff
@@ -100,6 +129,44 @@
...
+    rl_type = RlType.PYREADLINE
+
+    # Save the original pyreadline display completion function since we need to override it and restore it
+    # noinspection PyProtectedMember
+    orig_pyreadline_display = readline.rl.mode._display_completions
+
+else:
+    readline_lib_path = ''
+
+    if 'gnureadline' in sys.modules:
```
tleonhardt: Good work on figuring out how to dynamically find the library path.

cmd2.py
```diff
@@ -1127,9 +979,11 @@ def __init__(self, completekey='tab', stdin=None, stdout=None, persistent_histor
         # Call super class constructor.  Need to do it in this way for Python 2 and 3 compatibility
         cmd.Cmd.__init__(self, completekey=completekey, stdin=stdin, stdout=stdout)
 
-        # Commands to exclude from the help menu or history command
-        self.exclude_from_help = ['do_eof', 'do_eos', 'do__relative_load']
-        self.excludeFromHistory = '''history edit eof eos'''.split()
+        # Commands to exclude from the help menu and tab completion
+        self.hidden_commands = ['eof', 'eos', '_relative_load']
```
tleonhardt: I like how you just provide the command name now and don't need to include the **do_***

cmd2.py
```diff
@@ -1368,6 +1227,729 @@ def get_subcommands(self, command):
...
+            orig_pyreadline_display(matches)
+
+    def tokens_for_completion(self, line, begidx, endidx):
+        """
+        Used by tab completion functions to get all tokens through the one being completed
+        :param line: str - the current input line with leading whitespace removed
+        :param begidx: int - the beginning index of the prefix text
+        :param endidx: int - the ending index of the prefix text
+        :return: A 2 item tuple where the items are
+                 On Success
```
tleonhardt: Nice comments throughout!

setup.py
```diff
@@ -71,6 +71,8 @@
     ":sys_platform=='win32'": ['pyreadline'],
     # Python 3.4 and earlier require contextlib2 for temporarily redirecting stderr and stdout
     ":python_version<'3.5'": ['contextlib2'],
+    # Python 3.3 and earlier require enum34 backport of enum module from Python 3.4
+    ":python_version<'3.4'": ['enum34'],
```
tleonhardt: I added a conditional dependency on the enum34 module for Python 2.7.  This change here (and the associated one in the **tox.ini** file should fix the unit test failures on Python 2.7.

tleonhardt:
@kmvanbrunt The unit tests are all passing now.  Let me know when you are happy with this PR and I will do some more manual testing before merging it in.

If you want to increase code coverage that would be great.  But you have already put a tremendous amount of work into this PR, so we should probably wrap it up soonish and let you take a well deserved rest ;-)

PR #330, opened 2018-03-27
Wheel-compatible conditional requirements

nsoranzo:
See https://hynek.me/articles/conditional-python-dependencies/ for a background explanation.

The fallback for setuptools < 18 is inspired by https://gitlab.com/pycqa/flake8/blob/master/setup.py

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,6 +4,8 @@
 Setuptools setup file, used to install or test 'cmd2'
 """
 import sys
+
+import setuptools
 from setuptools import setup
 
 VERSION = '0.8.2'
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -64,17 +66,23 @@
...
 INSTALL_REQUIRES = ['pyparsing >= 2.0.1', 'pyperclip', 'six']
 
-# Windows also requires pyreadline to ensure tab completion works
-if sys.platform.startswith('win'):
-    INSTALL_REQUIRES += ['pyreadline']
-
-# Python 3.4 and earlier require contextlib2 for temporarily redirecting stderr and stdout
-if sys.version_info < (3, 5):
-    INSTALL_REQUIRES += ['contextlib2']
+EXTRAS_REQUIRE = {
```
tleonhardt: I wasn't familiar with this syntax within setuptools for conditional dependencies.

nsoranzo:
@tleonhardt Thanks for the merge!

> I don't think we normally release as a wheel, though I'm honestly not sure ...

Unfortunately not yet: https://pypi.python.org/pypi/cmd2

It is very simply and you can easily automate it with a Makefile:
```
$ python setup.py sdist
$ python setup.py bdist_wheel
$ twine upload dist/*
```
You just need recent versions of `pip`, `setuptools`, `twine` and `wheel` installed.

PR #332, opened 2018-03-30
Display length fix

kmvanbrunt:
Fixed bug where display width was not being calculated for display_matches

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -81,6 +81,8 @@ if int(setuptools.__version__.split('.')[0]) < 18:
     EXTRAS_REQUIRE = {}
     if sys.platform.startswith('win'):
         INSTALL_REQUIRES.append('pyreadline')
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -81,6 +81,8 @@
     EXTRAS_REQUIRE = {}
     if sys.platform.startswith('win'):
         INSTALL_REQUIRES.append('pyreadline')
+    else:
+        INSTALL_REQUIRES.append('wcwidth')
```
tleonhardt: Do you only need the **wcwidth** module on non-Windows platforms?

tleonhardt:
You should probably update the Changelog with a comment about this bug fix

setup.py
kmvanbrunt: Yes

PR #343, opened 2018-04-09
Fix tab_completion example

mrinalraghupathi:
Change the self to cmd2 since the completion helpers are function in the module cmd2 and not methods on the class/instance created.


kmvanbrunt:
@mrinalraghupathi 
The tab completion functions are instance methods and not module methods. During one release they were module methods, but that is no longer the case. The current examples are correct.

PR #344, opened 2018-04-10
Delete optparse

### LINKED ISSUE #340: Delete deprecated optparse code
For the 0.8.0 release of ``cmd2`` on February 1, 2018 we officially deprecated all usage of the **optparse** module within ``cmd2`` and replaced it with **argparse**-based decorators.

We should plan to delete the deprecated **optparse** code in either the 0.8.3 or 0.8.4 release.

Any opinions regarding which of these releases it should be deprecated in?

tleonhardt:
Deleted optparse code which had been previously deprecated in 0.8.0 release

Also:
- Bumped version to 0.9.0

This ended up being a larger change than I thought it would be at first mostly due to the fact that I forgot we had a bunch of unit tests that we had never ported to use the new decorators.

This closes #340


tests/test_transcript.py
```diff
@@ -266,15 +236,15 @@ def test_invalid_syntax(_cmdline_app, capsys):
     ('characterclass.txt', False),
     ('dotstar.txt', False),
     ('extension_notation.txt', False),
-    ('from_cmdloop.txt', True),
+    # ('from_cmdloop.txt', True),
```
tleonhardt: @kotfu I couldn't quickly figure out what was going wrong with these two unit tests here so I commented them out.  If you can figure it out without too much effort, then please go ahead and fix them.

tleonhardt:
The 0.8.5 changes have been merged in from **master** and the version in this PR has been bumped to 0.9.0.

Issue #346, opened 2018-04-10
version 0.8.3 dependency handling breaks cliff


dhellmann:
I had a bug report against cliff that cmd2 0.8.3 breaks the interactive mode because of a missing package. Looking at setup.py, I see it doing a bunch of logic in Python that could be handled using environment markers passed to setuptools instead that would let it be more reliable and let it work with wheels.

The original bug report is at https://bugs.launchpad.net/python-cliff/+bug/1762715

tleonhardt:
@dhellmann 
Hi Doug,

We apologize for breaking ``cliff``.  Thanks for creating the issue.

PR #330 was created by @nsoranzo to add "wheel-compatible conditional requirements" which sould now allow us to work with wheels as well as source distributions.  

I am relatively ignorant when it comes to ``setuptools``.  I just published a wheel of the 0.8.3 release to PyPI.  But it appears to be specific to Python 3.6.  Could you perhaps explain how we can publish a more generic wheel to PyPI?

Thanks,

Todd

tleonhardt:
Oh and for whatever it is worth, the new dependency that was added for POSIX (non-Windows) platforms is [wcwidth](https://pypi.python.org/pypi/wcwidth).

nsoranzo:
The issue was introduced in commit 2f4198ef94262dc0c747c34541a72128164dc340 where the new `wcwidth` requirement was added only for setuptools < 18 . I'll open a PR to fix this and allow generic wheels (if possible).

tleonhardt:
We will have to try releasing a new Wheel distribution to PyPI and see how it goes.

@dhellmann Do you think that the changes @nsoranzo made in PR #347 should address the problems the recent release of ``cmd2`` created for ``cliff``?

PR #347, opened 2018-04-10
 Add wcwidth for setuptools>=18 and build universal wheels

nsoranzo:
This at least partially addresses #346

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,2 @@
+[bdist_wheel]
+universal = 1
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -0,0 +1,2 @@
+[bdist_wheel]
+universal = 1
```
tleonhardt: Hmm, is that all you need to do for a universal wheel?  I hope it is that easy, that would be great.

nsoranzo: Yes, as long as:
- runs on Python 2 and 3 with no changes (i.e. it does not require 2to3)
- there are no C extensions.

See https://packaging.python.org/tutorials/distributing-packages/#universal-wheels

Issue #278, opened 2018-02-20
Consider categorizing output of 'help' command


kotfu:
Right now the help command just gives you an alphabetical columnar list of commands. It would be nice if there was a way to create groups or categories of commands, so you could have related commands show up together.

Instead of this:

```
Documented commands (type help <topic>):
========================================
config     expire       load      resources   shell                stop
connect    findleakers  py        restart     shortcuts            threaddump
deploy     help         pyscript  serverinfo  show                 undeploy
edit       history      quit      sessions    sslconnectorciphers  version
exit       license      redeploy  set         start                vminfo
exit_code  list         reload    settings    status               which
```
It would be nice if you could create something like:
```
Documented commands (for more info type help <command>)

Connecting
=========================================
connect     which

Application Management
=========================================
list
deploy       redeploy      undeploy
start        restart       stop
sessions     expire        findleakers

Server Information
==========================================
resources         serverinfo      sslconnectorciphers
status            thread dump     vminfo

Settings
==========================================
config       set           show

Other
=============================
edit      help        license
exit      history     load
version
```

It would be great if you could not only control the sections of the help, but the order and formatting of the commands within each section. Commands not specified for inclusion in a specific section would be featured in a catch all section like **Other**.

It might also be interesting to have a way to specify a short phrase of help for each command, with some command like **help --long** to show:

```
Documented commands (for more info type help <command>)

Connecting
=========================================
connect    connect to a tomcat server
which      show which server you are currently connected to
...
```

tleonhardt:
I really like the concept of having categories of commands which get displayed together in the help menu, particularly if this is an opt-in capability which defaults to the current behavior by default.  That would be a nice addition.

Potentially of relevance, the existing behavior actually divides the help menu into three categories:
- Documented commands
    - A **do_foo** method exists which has a docstring and/or an associated **help_foo** method
- Miscellaneous
    - A **help_foo** method exists, but there is no corresponding **do_foo**
- Undocumented commands
    - A **do_foo** method exists which does not have a docstring or an associated **help_foo** method

anselor:
I just added a low-impact implementation of help categories in the help_categories branch. 
Let me know what you guys think of this approach.

For the **help --long** (or maybe **help --verbose**), I was thinking just take the first non-empty line of function docstring used in the help for the command. Any thoughts?

tleonhardt:
[...]

PR #348, opened 2018-04-10
Help categories

anselor:
Adds ability to optionally group commands into categories:
screenshot_2018-04-10_15-04-53

As well as a way to see more verbose help output:
help -v / help --verbose
screenshot_2018-04-10_14-46-05

This addresses #278


cmd2.py
```diff
@@ -2913,21 +2931,80 @@ def _help_menu(self):
...
+        """Customized version of print_topics that can switch between verbose or traditional output"""
+        if cmds:
+            if not verbose:
+                self.print_topics(header, cmds, 15, 80)
+            else:
+                self.stdout.write('{}\n'.format(str(header)))
+                widest = 0
+                # measure the commands
+                for command in cmds:
+                    width = wcswidth(command)
```
tleonhardt: You need to put this in a conditional check since **wcswidth()** is not available on Windows.  You would so something like the following:

```Python
if 'gnureadline' in sys.modules or 'readline' in sys.modules:
    width = wcswidth(command) 
else:
    width = len(command)
```

examples/help_categories.py
```diff
@@ -0,0 +1,138 @@
+#!/usr/bin/env python
```
tleonhardt: The example is nice and helps explain to end users how to use it.

A couple unit tests covering the new features added would be extremely beneficial from a maintenance perspective.

tleonhardt:
Thanks for the PR @anselor! 

Please fix the dependency issue on Windows.

A few unit tests covering the new features would be greatly appreciated if you don't mind taking the time to add them.

tleonhardt:
If you update the Sphinx docs a little to mention the new features, you get bonus points ;-)

cmd2.py
anselor: It turns out we don't allow arbitrary unicode characters in the command names so I can actually just remove this.

kmvanbrunt:
I will need to tweak complete_help() since it isn't expecting a flag.

kmvanbrunt:
@anselor @tleonhardt 
Nevermind, the complete_help() function is fine with this change.

I do think we need to change what displays in verbose for some commands.
For instance, ``help`` says "usage: history [-h] [-r | -e | -s | -o FILE | -t TRANSCRIPT] [arg]"
I would assume it would say, "View, run, edit, and save previously entered commands" from the docstring.

anselor:
@tleonhardt I don't see anything in the current Sphinx docs covering how to document your commands.

tleonhardt:
Wow, we really don't.  The closes we have is this section:
http://cmd2.readthedocs.io/en/latest/argument_processing.html#help-messages

Feel free to add a section earlier ;-)

tests/conftest.py
```diff
@@ -19,6 +19,24 @@
 edit   history  py    quit      shell  unalias
 """
 
+BASE_HELP_VERBOSE = """
```
tleonhardt: Thanks for adding the unit tests.  They should really help with maintainability.

anselor:
@tleonhardt I'm looking forward to retiring Python 2 support. The gymnastics to keep it working are annoying.

tleonhardt:
@anselor Yeah, I now wish that @kotfu and I had agreed on an earlier date for ending Python 2 support than August 31, 2018.  The burden of supporting Python 2.7 is getting worse by the month.

Tell you what, if you can convince @kotfu that moving the date earlier (perhaps May 31, 2018 or even April 30, 2018) is desirable then I am happy to buy-in to that decision.  I know moving it earlier would make @kmvanbrunt very happy.

cmd2.py
```diff
@@ -113,6 +112,12 @@ def __subclasshook__(cls, C):
 else:
     from contextlib import redirect_stdout, redirect_stderr
 
+if sys.version_info > (3, 0):
```
tleonhardt: FYI, using the ``six`` module can help avoid Python 2 vs 3 compatibility gymnastics such as this.

``six`` is a Python 2 vs 3 compatibility module (6 == 2 * 3 == 3 * 2).

cmd2.py
```diff
@@ -340,6 +365,14 @@ def parse_quoted_string(cmdline):
     return lexed_arglist
 
 
+def with_category(category):
```
tleonhardt: I like that you added a decorator, that should make the code clean and easy to read.

I strongly encourage you to make one simple change to your decorator to help avoid order-of-operations dependencies when multiple decorators are involved.

You should apply the "@functools.wraps(func)" decorator to your internal **cat_decorator** function - see how we do it in other decorators below such as **with_argument_list**:
```Python
def with_category(category):
    """A decorator to apply a category to a command function"""
    @functools.wraps(func)
    def cat_decorator(func):
        categorize(func, category)
        return func
    return cat_decorator
```

anselor: I don't think that's necessary in this situation because this decorator function is truly a decorator function rather than a wrapping function. This adds an attribute the the passed-in function and returns it.

tleonhardt: You make a good point, your decorator is mutating and returning the original function.  I think you are fine as-is.

PR #351, opened 2018-04-12
Pass the return value back through for wrapped functions.

### LINKED ISSUE #350: command decorator wrapper functions don't return the function's return value
The wrapper functions are not returning the wrapped function's result. When trying to override and add arguments to do_quit(), this causes do_quit() to not actually quit.

anselor:
Resolves #350

### DIFF
--- a/cmd2.py
+++ b/cmd2.py
@@ -380,7 +380,7 @@ def with_argument_list(func):
     @functools.wraps(func)
     def cmd_wrapper(self, cmdline):
         lexed_arglist = parse_quoted_string(cmdline)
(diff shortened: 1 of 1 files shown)


cmd2.py
```diff
@@ -380,7 +380,7 @@ def with_argument_list(func):
     @functools.wraps(func)
     def cmd_wrapper(self, cmdline):
         lexed_arglist = parse_quoted_string(cmdline)
-        func(self, lexed_arglist)
+        return func(self, lexed_arglist)
```
tleonhardt: This looks good to me.

PR #352, opened 2018-04-13
User expansion

kmvanbrunt:
Allowing the use of ~user tab completion in paths

### DIFF
--- a/cmd2.py
+++ b/cmd2.py
@@ -1600,10 +1600,11 @@ class Cmd(cmd.Cmd):
         :return: List[str] - a list of possible tab completions
         """
 
-        # Used to complete ~ and ~user strings with a list of users that have existing home dirs
+        # Used to complete ~ and ~user strings
         def complete_users():
 
-            # We are returning ~user strings that resolve to directories, so don't append a space or quote
+            # We are returning ~user strings that resolve to directories,
+            # so don't append a space or quote in the case of a single result.
             self.allow_appended_space = False
             self.allow_closing_quote = False
 
@@ -1692,7 +1693,7 @@ class Cmd(cmd.Cmd):
             matches = [c for c in matches if os.path.isdir(c)]
 
         # Don't append a space or closing quote to directory
-        if len(matches) == 1 and not os.path.isfile(matches[0]):
+        if len(matches) == 1 and os.path.isdir(matches[0]):
             self.allow_appended_space = False
             self.allow_closing_quote = False
 
--- a/cmd2.py
+++ b/cmd2.py
@@ -1599,6 +1599,35 @@ class Cmd(cmd.Cmd):
         :param dir_only: bool - only return directories
         :return: List[str] - a list of possible tab completions
         """
+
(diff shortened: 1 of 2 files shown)


cmd2.py
```diff
@@ -1599,6 +1599,42 @@ def path_complete(self, text, line, begidx, endidx, dir_exe_only=False, dir_only
...
+        def complete_users():
+
+            # We are returning ~user strings that resolve to directories,
+            # so don't append a space or quote in the case of a single result.
+            self.allow_appended_space = False
+            self.allow_closing_quote = False
+
+            users = []
+
+            # Windows lacks the pwd module so we can't get a list of users.
```
tleonhardt: Nice comments throughout

tests/test_completion.py
```diff
@@ -341,25 +342,18 @@ def test_path_completion_doesnt_match_wildcards(cmd2_app, request):
...
-    # Test a missing separator between a ~ and path
-    text = '~Desktop'
-    line = 'shell fake {}'.format(text)
-    endidx = len(line)
-    begidx = endidx - len(text)
-
-    assert cmd2_app.path_complete(text, line, begidx, endidx) == []
+def test_path_completion_expand_user_dir(cmd2_app):
+    # Get the current user
+    user = getpass.getuser()
```
tleonhardt: For some reason this is causing a unit test failure on the AppVeyor Windows CI build/test systems.  All unit tests pass for this branch on my Windows 10 VM for both Python 3.6 and 2.7.  

But for some reason this is hosed on AppVeyor.   As far as I can tell, the problem is with AppVeyor and not with the code.  

If you can't find a good workaround quickly, I recommend just skipping this test on Windows.

kmvanbrunt: For some reason ``getpass.getuser()`` fails when running on AppVeyor. The user must not have any of these environment variables. Unfortunately the fallback imports the ``pwd`` module which doesn't exist on Windows.

```Python
for name in ('LOGNAME', 'USER', 'LNAME', 'USERNAME'):
            user = os.environ.get(name)
            if user:
                return user

# If this fails, the exception will "explain" why
import pwd
return pwd.getpwuid(os.getuid())[0]
```

I am trying a workaround.

tests/test_completion.py
```diff
@@ -341,25 +341,19 @@ def test_path_completion_doesnt_match_wildcards(cmd2_app, request):
...
-    text = '~Desktop'
-    line = 'shell fake {}'.format(text)
-    endidx = len(line)
-    begidx = endidx - len(text)
-
-    assert cmd2_app.path_complete(text, line, begidx, endidx) == []
+def test_path_completion_expand_user_dir(cmd2_app):
+    # Get the current user. We can't use getpass.getuser() since
+    # that doesn't work when running these tests on Windows in AppVeyor.
+    user = os.path.basename(os.path.expanduser('~'))
```
tleonhardt: Assuming the unit test all pass, this looks OK.  They still pass on my system.

PR #353, opened 2018-04-14
Unhackify

kmvanbrunt:
Since we have made our tab completion word breaks consistent with our own shlex/parsing word breaks, we no longer need to call directly into **readline** library to add an opening quote. In all cases now, unquoted text being passed in by **readline** will match how we tokenize the line for tab completion. There were cases before where adding an opening quote required rewriting the whole input line of **readline**. This is no longer the case.

Therefore I removed all the ``ctypes`` functions that were used to add an opening quote and do all of it in Python.


kmvanbrunt:
@tleonhardt 
I added a fix for that extraneous opening quote.

cmd2.py
```diff
@@ -2173,36 +2072,43 @@ def complete(self, text, state):
...
+
+                        # Get the common prefix of all matches. This is the actual tab completion.
+                        common_prefix = os.path.commonprefix(self.completion_matches)
+
+                        # Join all matches into 1 string for ease of searching
+                        all_matches_str = ''.join(self.completion_matches)
+
+                        # If there is a tab completion that will alter the text on the command line and
+                        # any of the matches have a space, then we must add an opening quote to the matches.
+                        if common_prefix != text and ' ' in all_matches_str:
```
tleonhardt: Thanks for this fix.  From some manual testing, this improved the behavior significantly.  The behavior now matches my expectations about how it should work.

cmd2.py
```diff
@@ -1107,8 +1107,11 @@ def __init__(self, completekey='tab', stdin=None, stdout=None, persistent_histor
         # will be added if there is an unmatched opening quote
         self.allow_closing_quote = True
 
-        # If the tab-completion matches should be displayed in a way that is different than the actual match values,
-        # then place those results in this list. path_complete uses this to show only the basename of completions.
+        # Use this list if you are completing strings that contain a common delimiter and you only want to
```
tleonhardt: Ok, so your fix for some previous issues uses ``self.display_matches`` in a few cases where it previously used ``self.completion_matches`` and improves the documentation to explain why.

This seems good to me.

.gitignore
```diff
@@ -1,11 +1,19 @@
+# Python development, test, and build
```
tleonhardt: Mostly I just added some comments to the .gitignore

The substantive change is the the **.vscode** directory containing configuration for Visual Studio Code is now ignored.

PR #354, opened 2018-04-15
Move to Python 3.4+ only

### LINKED ISSUE #260: Discontinuing support for Python 2.7 on April 15, 2018
Supporting multiple old versions of Python is becoming more and more burdensome as good features get added to newer versions of Python which don't get back-ported to older versions.

I don't want to abandon support for versions of Python that users are actively using.  But I also don't want to waste developer time supporting versions nobody cares about.

So whoever reads this, please feel free to add a comment and say what versions of Python you care about ``cmd2`` supporting now and in the near future.

tleonhardt:
``cmd2`` now supports Python 3.4+ only for versions starting with 0.9.0.  This PR starts removing support for Python 2.7 and begins cleaning up things which are no longer needed.

Changes include:
- No longer run Python 2.7 unit tests on either TravisCI or AppVeyor
- On AppVeyor, Windows code coverage analysis is now done for Python 3.6 instead of 2.7
- 3rd party ``mock`` module is now only used for tests on Python 3.5
- Cleaned up most references to Python 2.7 in the docs
- Removed all dependency on the ``six`` module
- Removed conditional code branches which were for Python 2.7
- Started using **super()** to call base class dunder init

This closes #260

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -4,4 +4,4 @@ install:
 build: off
 
 test_script:
-  - python -m tox -e py27-win,py35-win,py36-win
+  - python -m tox -e py35-win,py36-win
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,6 @@ sudo: false  # false enables container-based build for fast boot times on Linux
 matrix:
     include:
         - os: linux
-          python: 2.7
-          env: TOXENV=py27
-        - os: linux
           python: 3.4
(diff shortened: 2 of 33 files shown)


tleonhardt:
I'm going to go ahead and merge this.  We can clean up and/or change in future PRs if need be.  But future work will depend on this, so I'm going to merge it in.

Issue #358, opened 2018-04-17
Batch mode and exit on error with code not 0


albertored:
I need to be able to perform one or more commands directly from CLI call and exit on finish. I've resolved this need overriding the `onecmd_plus_hooks` method of my `Cmd` class to return `True` in some circumstances and adding a flag to the CLI to set the batch mode:

```python
def onecmd_plus_hooks(self, line):
    stop = Cmd.onecmd_plus_hooks(self, line)
   
    if self.mode == self.MODE_BATCH:
        # if there are commands in queue we exec all before stopping
        if self.cmdqueue:
            return stop
        # queue is empty and we are in batch mode, exit
        return True
    else:
        return stop           
```
(This works pretty well so if you think is useful I can try to integrate it in the package)

There is a problem anyway... The expected behaviour of a CLI call in batch mode is that the exit code of the whole call is 0 if the command is successful and different from zero if there are exceptions in command execution or if the user press CTRL+C during the execution.

With the current implementation the command always exits with a 0 code. I'm a bit stuck on how to implement this behaviour. I'm trying modifying a little bit the exceptions handling in `onecmd_plus_hooks`  and `_cmdloop` methods but I don't know if it is the correct way to get a solution. Do you have any comment/suggestion?

If you think is worth I can push a branch with my attempt

albertored:
Thanks @tleonhardt for your reply, nice think that of using `quit` for batch processing!

For  what regards the exit code my use case is calling the cli in some script and being able to catch failures. Using `postloop` seems a lot verbose, the only way I can think of is surrounding all my commands with `try` statements, setting some class attribute in exceptions and use that in `postloop` for exiting with a non-zero code. It also seems not easy to handle exiting with non-zero code on sigint.

I'm pushing a PR with a generic way of solving this, it seems to work but it still under development (some tests are failing). It is not related to the piece of code for handling batch mode that I've shared before. You can take a look and let me know what you think.

Thanks

tleonhardt:
As far as returning an exit code, maybe you just need to change **cmdloop()** so that it saves the return value of **_cmdloop()**.

But I'm curious to see what you come up with.  I'll take a look at your PR.  Thanks.

tleonhardt:
The is some good discussion in PR #359 in relation to this issue.

I think the tentative resolution is that we will leave this issue open and revisit it after Issue #37 is complete and ``cmd2``'s underlying parsing logic has been refactored.

[...]

PR #359, opened 2018-04-17
[Add] exit_with_error_code attribute

albertored:
**UNDER DEPLOYMENT - DO NOT MERGE**

I added an exception containing an exit code, that is used in `_cmdloop` for exiting the CLI with that exit code. The exception can be raised by users in custom commands or automatically for every command. This last feature can be activated by setting the class attribute `exit_with_error_code` to `True`.

I know it is on the wrong branch, if you think it's worth merging I will fix tests, and move to the right place.

See #358

### DIFF
--- a/cmd2.py
+++ b/cmd2.py
@@ -668,6 +668,16 @@ class EmptyStatement(Exception):
     pass
 
 
+class ExitWithError(Exception):
+    """Custom exception class for handling exit code of CLI
+
+    Attributes:
+        code -- the exit code
+    """
+    def __init__(self, code=1):
+        self.code = code
+
+
 # Regular expression to match ANSI escape codes
 ANSI_ESCAPE_RE = re.compile(r'\x1b[^m]*m')
 
@@ -954,10 +964,11 @@ class Cmd(cmd.Cmd):
     terminators = [';']     # make sure your terminators are not in legalChars!
 
     # Attributes which are NOT dynamically settable at runtime
-    allow_cli_args = True       # Should arguments passed on the command-line be processed as commands?
-    allow_redirection = True    # Should output redirection and pipes be allowed
-    default_to_shell = False    # Attempt to run unrecognized commands as shell commands
-    quit_on_sigint = False      # Quit the loop on interrupt instead of just resetting prompt
+    allow_cli_args = True         # Should arguments passed on the command-line be processed as commands?
+    allow_redirection = True      # Should output redirection and pipes be allowed
+    default_to_shell = False      # Attempt to run unrecognized commands as shell commands
+    quit_on_sigint = False        # Quit the loop on interrupt instead of just resetting prompt
+    exit_with_error_code = False  # Exit with a code different from zero on exceptions or sigint
     reserved_words = []
 
     # Attributes which ARE dynamically settable at runtime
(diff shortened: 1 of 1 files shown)


tleonhardt:
@albertored I'm not sure that I see this as been generally applicable to all situations.  When we can, we try to steer clear of design/implementation decisions which would in any way impact overall application architecture because we want developers to choose what makes sense for themselves.

If you want to take the time to clean it up, base it off of the correct branch, and get all of the tests to pass and add an example then I would be happy to take a look at it and maybe we can find a way to incorporate it that would make sense.  But I don't want you to waste your time.  My tentative thought is that perhaps this is left out of ``cmd2`` and up to the end user.

@kmvanbrunt or @kotfu do you have any thoughts on this?

kmvanbrunt:
@albertored 
While this may a good feature for your application, it doesn't seem generic enough to be included in the cmd2 code. I have the same thing in my own cmd2-based applications. Many of the features I implement are nice for a particular program, but don't have the broader appeal needed for addition to cmd2.

albertored:
I completely agree with you when you say that we should include in a library only features that can have a broad appeal for the users. If you think this is not a feature that can have this appeal I respect your choice but personally I disagree.
I disagree on the usefulness of the feature, this particular implementation can be totally ugly or wrong but I think having a way to detect when some command fails (not only seeing an error message) is useful.

I still don't see a way of implementing this in my own cmd2-based application only by overriding cmd2 methods... @kmvanbrunt if I understand correctly you managed to do this, right? Can I ask how?

Thanks for reviewing

kmvanbrunt:
@albertored 
I agree that knowing the return code of a command is useful. However, I'm not ready to accept any solution to this problem until **cmd2** stabilizes and we have time to consider what solution is most universal. Currently **cmd2**'s parsing logic is being completely rewritten, tab completion is being extended with the addition of AutoCompleter, and its being broken up into multiple files.

The need for the feature you suggest does exist. Without it, using ``cmdqueue`` is fairly limited since commands keep running until then end in all cases. Therefore ``load`` can only be used in simple cases. I have typically used the more featured ``pyscript`` command when logic is needed to determine what to do after a command runs. But many times this involves more than a return code since the parsing of a command's output can also be useful in knowing what to do next. I've done this using a ``CmdResult`` class member. This class is defined in cmd2.py and I believe some of the example scripts show its usage. But that approach is probably overkill for a batch mode like you are using.

There are other ways to solve this that need to be examined. Perhaps the solution would include something syntactical like the ``&&`` operator in shells like **bash**. That provides more granularity than the ``exit_with_error_code`` flag which always results in exiting upon command failure. Sometimes a developer cares if a particular command fails and would want to exit, but other times the failure of that same command won't affect the next command in the list and exiting would be the wrong thing to do. Logging is usually enough in those cases.

I believe the best solution will be implemented if we have more time to consider all possibilities.

Thoughts?

tleonhardt:
Ok, how about we leave the Issue related to this open and revisit it after ``cmd2``'s parsing logic has been refactored?

It would be nice to have some standard supported way of returning a non-zero exit code to the shell under certain circumstances.  We just want to be careful about how we implement it.  

One thing to consider is the fact that ``cmd2`` is an extension of ``cmd`` somewhat constrains the kind of architectural changes we can make to support this feature.  Really this is the kind of feature that should be built into ``cmd``, but that is a whole other can of works ;-)

albertored:
It sounds good to me, I'll close this and create a new one when it's time

PR #360, opened 2018-04-18
Convert cmd2 to a multi-file Python package

### LINKED ISSUE #356: Convert to a package directory structure
``cmd2`` has grown to the point where the **cmd2.py** file is over 4000 lines and with new features being added the overall code size is likely to grow substantially.

It has probably reached the point that ``cmd2`` would benefit by becoming a real Python package with multiple source code files.

For this to happen, we will need a **cmd2** directory in which all of the source code lives and there is a dunder init.py file to make it a package.  This article on [Structuring Your Project](http://docs.python-guide.org/en/latest/writing/structure/) has good basic advice on how to approach this.

tleonhardt:
Thie PR converts ``cmd2`` to a multi-file Python package with a ``__init__.py`` and source code located in the **cmd2** directory.

All unit tests pass (at least on my system) and the Sphinx docs appear to build correctly.

Installation via **pip** and releasing to **PyPI** remains to be tested.

This closes #356.


tleonhardt:
History and HistoryItem are mostly intended for internal use only.  Anyone is welcome to use them, but I would view that as uncommon.  I'll go ahead and merge.

Issue #349, opened 2018-04-12
argparse tab autocompleter


anselor:
I'm working on a tab completion implementation that automatically generates completions based on an argparse definition. Creating this issue to enable discussions.

tleonhardt:
f we move to the state where we have a 2nd file as part of the cmd2 module instead of just the single cmd2.py file, then we should change the directory structure to create a src directory and make cmd2 a proper Python package by including a init.py and such.

Hopefully this should be a small change, but it will likely require small tweaks in all of the following areas:

packaging for distribution to PyPI
Sphinx documentation - where it looks for the cmd2 source when using autodoc features
unit testing - where it looks for the cmd2 source and what it does to install before running tests

tleonhardt:
When you create a PR for this, make sure you update **setup.py** to add a dependency on ``colorama`` if you use that module. I'm fine with that - it is a very common library which I always end up using anyways.

For a long time I've been considering adding a dependency on ``colorama`` and retrofitting ``cmd2``'s existing colorizing code to use ``colorama`` so that it will work on Windows as well as the POSIX OSes.

tleonhardt:
From the limited experimentation I have done playing with this feature on the **autocompleter** branch, I think it rocks socks.  It is a very slick feature which makes tab completing everything related to ``argparse`` arguments quite simple.  And the way it provides a hint specifically for the argument you are attempting to complete is very helpful.

tleonhardt:
@anselor 
I love this feature.  It combines so synergistically with the existing ``argparse`` decorators and existing tab-completion functionality that it goes a long way to make it just so damn easy to provide awesome tab-completion that end users will love pretty much "out-of-the-box".  

I recommend that you work in conjunction with @kmvanbrunt  to update the "front-page" code example in **README.md** to make use this feature as well as the argument and/or index based tab completion and make that part of the PR that merges this feature in.  Keep it simple, but make it readily obvious that these features exist.

Basically I want to make these powerful tab-completion capabilities very discoverable for new ``cmd2`` users as I believe they are something which really set ``cmd2`` apart from competing alternatives.

anselor:
@tleonhardt 

Instead of updating the example in README,md, I'm thinking I'll change how cmd2 handles all with_argparse commands such that if a complete_COMMAND function isn't provided cmd2 will automatically create an AutoCompleter. Users can still optionally implement the complete_CCOMMAND function to do whatever they want.

tleonhardt:
[...]

PR #362, opened 2018-04-18
Autocompleter

anselor:
This PR addresses most of the feature described in #349 and adds examples and unit tests exercising the new features.
In a future round of changes, cmd2 will be updated to automatically use AutoCompleter for completion on all argparse-driven commands (unless overridden with a custom completion function)

### DIFF
--- a/cmd2/argparse_completer.py
+++ b/cmd2/argparse_completer.py
@@ -52,6 +52,9 @@ How to supply completion choice lists or functions for sub-commands:
     The subcommand group dictionary maps subcommand names to tuple(arg_choices, subcmd_args_lookup)
 
     For more details of this more complex approach see tab_autocompletion.py in the examples
+
+Copyright 2018 Eric Lin <anselor@gmail.com>
+Released under MIT license, see LICENSE file
 """
 
 import argparse
@@ -261,6 +264,12 @@ class AutoCompleter(object):
         matched_flags = []
         current_is_positional = False
         consumed_arg_values = {}  # dict(arg_name -> [values, ...])
+
+        # the following are nested functions that have full access to all variables in the parent
+        # function including variables declared and updated after this function.  Variable values
+        # are current at the point the nested functions are invoked (as in, they do not receive a
+        # snapshot of these values, they directly access the current state of variables in the
+        # parent function)
 
         def consume_flag_argument() -> None:
             """Consuming token as a flag argument"""
--- a/cmd2/rl_utils.py
+++ b/cmd2/rl_utils.py
@@ -18,7 +18,7 @@ except ImportError:
     try:
         # noinspection PyUnresolvedReferences
         import readline
-    except ImportError:
+    except ImportError:  # pragma: no cover
         pass
 
 
@@ -51,7 +51,8 @@ def rl_force_redisplay() -> None:
     """
     if not sys.stdout.isatty():
         return
-    if rl_type == RlType.GNU:
+
+    if rl_type == RlType.GNU:  # pragma: no cover
         # rl_forced_update_display() is the proper way to redraw the prompt and line, but we
         # have to use ctypes to do it since Python's readline API does not wrap the function
         readline_lib.rl_forced_update_display()
@@ -60,6 +61,6 @@ def rl_force_redisplay() -> None:
         display_fixed = ctypes.c_int.in_dll(readline_lib, "rl_display_fixed")
         display_fixed.value = 1
 
-    elif rl_type == RlType.PYREADLINE:
+    elif rl_type == RlType.PYREADLINE:  # pragma: no cover
         # noinspection PyProtectedMember
         readline.rl.mode._print_prompt()
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -1591,7 +1591,7 @@ class Cmd(cmd.Cmd):
         return compfunc(text, line, begidx, endidx)
 
     @staticmethod
-    def _pad_matches_to_display(matches_to_display):
+    def _pad_matches_to_display(matches_to_display):  # pragma: no cover
         """
         Adds padding to the matches being displayed as tab completion suggestions.
         The default padding of readline/pyreadine is small and not visually appealing
@@ -1613,7 +1613,7 @@ class Cmd(cmd.Cmd):
 
         return [cur_match + padding for cur_match in matches_to_display], len(padding)
 
-    def _display_matches_gnu_readline(self, substitution, matches, longest_match_length):
+    def _display_matches_gnu_readline(self, substitution, matches, longest_match_length):  # pragma: no cover
         """
         Prints a match list using GNU readline's rl_display_match_list()
         This exists to print self.display_matches if it has data. Otherwise matches prints.
@@ -1664,7 +1664,7 @@ class Cmd(cmd.Cmd):
             # Redraw prompt and input line
             rl_force_redisplay()
 
-    def _display_matches_pyreadline(self, matches):
+    def _display_matches_pyreadline(self, matches):  # pragma: no cover
         """
         Prints a match list using pyreadline's _display_completions()
         This exists to print self.display_matches if it has data. Otherwise matches prints.
@@ -3344,7 +3344,7 @@ Script should contain one command per line, just like command would be typed in 
             # self._script_dir list when done.
             with open(expanded_path, encoding='utf-8') as target:
                 self.cmdqueue = target.read().splitlines() + ['eos'] + self.cmdqueue
-        except IOError as e:
+        except IOError as e:  # pragma: no cover
             self.perror('Problem accessing script from {}:\n{}'.format(expanded_path, e))
             return
 
@@ -3371,7 +3371,7 @@ Script should contain one command per line, just like command would be typed in 
                 # noinspection PyUnusedLocal
                 if sum(1 for line in f) > 0:
                     valid_text_file = True
-        except IOError:
+        except IOError:  # pragma: no cover
             pass
         except UnicodeDecodeError:
             # The file is not ASCII. Check if it is UTF-8.
@@ -3381,7 +3381,7 @@ Script should contain one command per line, just like command would be typed in 
                     # noinspection PyUnusedLocal
                     if sum(1 for line in f) > 0:
                         valid_text_file = True
-            except IOError:
+            except IOError:  # pragma: no cover
                 pass
             except UnicodeDecodeError:
                 # Not UTF-8
--- a/examples/tab_autocompletion.py
+++ b/examples/tab_autocompletion.py
@@ -1,6 +1,10 @@
 #!/usr/bin/env python
 # coding=utf-8
-"""A simple example demonstrating how to use flag and index based tab-completion functions
+"""
+A example usage of the AutoCompleter
+
+Copyright 2018 Eric Lin <anselor@gmail.com>
+Released under MIT license, see LICENSE file
 """
 import argparse
 import itertools


cmd2/AutoCompleter.py
```diff
@@ -0,0 +1,796 @@
...
+
+        flag_arg = AutoCompleter._ArgumentState()
+        flag_action = None
+
+        matched_flags = []
+        current_is_positional = False
+        consumed_arg_values = {}  # dict(arg_name -> [values, ...])
+
+        def consume_flag_argument() -> None:
+            """Consuming token as a flag argument"""
```
tleonhardt: While I believe that this function and the next one do work properly within the context of your code, it feels slightly awkward that they are using the ``token`` local variable which isn't passed in to them, but exists within the calling function's scope.

Is there a good reason for doing it this way?

I guess it is OK.  But consider possibly refactoring.  It just took me a little longer to read the code and figure out what is going on.  Initially I was like "WTF ``token`` is an undefined variable".

cmd2/AutoCompleter.py
```diff
@@ -0,0 +1,796 @@
...
+                        # reset some tracking values
+                        flag_arg.reset()
+                        # don't reset positional tracking because flags can be interspersed anywhere between positionals
+                        flag_action = None
+
+                        # does the token fully match a known flag?
+                        if token in self._flag_to_action.keys():
+                            flag_action = self._flag_to_action[token]
+                        elif hasattr(self._parser, 'allow_abbrev') and self._parser.allow_abbrev:
+                            candidates_flags = [flag for flag in self._flag_to_action.keys() if flag.startswith(token)]
```
tleonhardt: In Python, iterating through a dictionary just iterates through the keys, so no need to call the **.keys()** method, can just iterate through the dictionary directly.

cmd2/AutoCompleter.py
```diff
@@ -0,0 +1,796 @@
...
+                            pos_action = None
+
+                            # are we at a sub-command? If so, forward to the matching completer
+                            if pos_index < len(self._positional_actions):
+                                action = self._positional_actions[pos_index]
+                                pos_name = action.dest
+                                if pos_name in self._positional_completers.keys():
+                                    sub_completers = self._positional_completers[pos_name]
+                                    if token in sub_completers.keys():
+                                        return sub_completers[token].complete_command(tokens, text, line,
```
tleonhardt: This code near here-ish has a whole lot of nested conditionals.  At his point I think we are 7 or 8 levels deep.  I'm not entirely sure how to make it less nested.  Just something to think about.  It just makes it somewhat difficult to follow the logic.

cmd2/AutoCompleter.py
```diff
@@ -0,0 +1,796 @@
...
+            else:
+                arg_state.min = action.nargs
+                arg_state.max = action.nargs
+
+    def _complete_for_arg(self, action: argparse.Action,
+                          text: str,
+                          line: str,
+                          begidx: int,
+                          endidx: int,
+                          used_values=list()) -> List[str]:
```
tleonhardt: **WARNING: Mutable default arguments in Python behave like static local variables in C** - they are forever persistent and if you change the value within the function, you have permanently altered the default for the next time you call the function.

If this is how you intend this function to work, then please document to make it clear.

If this is NOT how you intend this function to work, then use a non-mutable default argument.  If a tuple can work for you then that is fine.  If you really need that argument to be a list then you can default it to ``None`` and then immediately first thing within the function check if it is None and if so assign an empty list to it.

cmd2/AutoCompleter.py
```diff
@@ -0,0 +1,796 @@
...
+        # prefix with 'usage:'
+        return '%s%s\n\n' % (prefix, usage)
+
+    def _format_action_invocation(self, action):
+        if not action.option_strings:
+            default = self._get_default_metavar_for_positional(action)
+            metavar, = self._metavar_formatter(action, default)(1)
+            return metavar
+
+        else:
```
tleonhardt: This **else** is unnecessary since within the **if** above, you have a **return**.

cmd2/AutoCompleter.py
```diff
@@ -0,0 +1,796 @@
...
+
+            # if the Optional doesn't take a value, format is:
+            #    -s, --long
+            if action.nargs == 0:
+                parts.extend(action.option_strings)
+                return ', '.join(parts)
+
+            # if the Optional takes a value, format is:
+            #    -s ARGS, --long ARGS
+            else:
```
tleonhardt: This **else** is unnecessary since within the **if** above, you have a **return**.

cmd2/AutoCompleter.py
```diff
@@ -0,0 +1,796 @@
+# coding=utf-8
+import argparse
+import re as _re
+import sys
+from argparse import OPTIONAL, ZERO_OR_MORE, ONE_OR_MORE, REMAINDER, PARSER, ArgumentError, _
+from .rl_utils import rl_force_redisplay
```
tleonhardt: In general I try to order imports like so:
1st:  built-in Python modules
2nd: 3rd-party modules
3rd: Our code
(and each section is alphabetical within that).

So recommend moving import from .rl_utils to end.

cmd2/rl_utils.py
```diff
@@ -0,0 +1,63 @@
...
+    # Try to import readline, but allow failure for convenience in Windows unit testing
+    # Note: If this actually fails, you should install readline on Linux or Mac or pyreadline on Windows
+    try:
+        # noinspection PyUnresolvedReferences
+        import readline
+    except ImportError:
+        pass
+
+
+# Check what implementation of readline we are using
```
tleonhardt: Convert comment before class to docstring after

docs/requirements.txt
```diff
@@ -1,3 +1,4 @@
 pyparsing
 pyperclip
 wcwidth
+colorama
```
tleonhardt: Thanks for adding this.  We need to keep this up-to-date so that the documentation builds correctly on ReadTheDocs

tests/test_autocompletion.py
```diff
@@ -0,0 +1,265 @@
...
+def test_autcomp_custom_func_list_and_dict_arg(cmd2_app):
+    text = ''
+    line = 'library show add SW_REB {}'.format(text)
+    endidx = len(line)
+    begidx = endidx - len(text)
+
+    first_match = complete_tester(text, line, begidx, endidx, cmd2_app)
+    assert first_match is not None and \
+           cmd2_app.completion_matches == ['S01E02', 'S01E03', 'S02E01', 'S02E03']
+
```
tleonhardt: Probably don't need quite this many trailing newlines at end of file

tox.ini
```diff
@@ -20,7 +20,7 @@ deps =
   pytest-xdist
   wcwidth
 commands =
-  py.test {posargs: -n 2} --cov=cmd2 --cov-report=term-missing --forked
+  py.test {posargs: -n 2} --cov --cov-report=term-missing --forked
```
tleonhardt: Thanks for fixing these

cmd2/AutoCompleter.py
```diff
@@ -0,0 +1,796 @@
+# coding=utf-8
+import argparse
+import re as _re
+import sys
+from argparse import OPTIONAL, ZERO_OR_MORE, ONE_OR_MORE, REMAINDER, PARSER, ArgumentError, _
```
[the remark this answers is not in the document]
anselor: Nope, they have some sort of function called '_' that they use. I had to import it to override their behavior. Bizarre.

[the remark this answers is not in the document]
tleonhardt: Huh ... Well ain't that special ;-)  They win the prize for the single worst function name I have ever seen!!!

cmd2/AutoCompleter.py
```diff
@@ -0,0 +1,796 @@
...
+        elif action.nargs == REMAINDER:
+            result = '...'
+        elif action.nargs == PARSER:
+            result = '%s ...' % get_metavar(1)
+        else:
+            formats = ['%s' for _ in range(action.nargs)]
+            result = ' '.join(formats) % get_metavar(action.nargs)
+        return result
+
+    def _metavar_formatter(self, action, default_metavar):
```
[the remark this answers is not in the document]
anselor: All off these are overriding parent methods. I'm adding a big comment block clearly that divides 'our' code from 'their' code. Everything below that point is an override that copies and tweaks the official behavior from Python.

cmd2/AutoCompleter.py
```diff
@@ -0,0 +1,796 @@
...
+                pos_usage = format(positionals, groups)
+                req_usage = format(required_options, groups)
+                opt_parts = _re.findall(part_regexp, opt_usage)
+                pos_parts = _re.findall(part_regexp, pos_usage)
+                req_parts = _re.findall(part_regexp, req_usage)
+                assert ' '.join(opt_parts) == opt_usage
+                assert ' '.join(pos_parts) == pos_usage
+                assert ' '.join(req_parts) == req_usage
+
+                # helper for wrapping lines
```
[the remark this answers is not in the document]
anselor: That comment is a copy from the parent. Where possible I'm preserving it exactly as it was in argparse for easier comparison when they make changes

cmd2/AutoCompleter.py
```diff
@@ -0,0 +1,796 @@
+# coding=utf-8
+import argparse
+import re as _re
+import sys
+from argparse import OPTIONAL, ZERO_OR_MORE, ONE_OR_MORE, REMAINDER, PARSER, ArgumentError, _
+from .rl_utils import rl_force_redisplay
+try:
+    from typing import List, Dict, Tuple, Callable, Union
+except:
```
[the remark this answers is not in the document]
anselor: I'm going to just take this out since it didn't actually fix the problem. The addition of a dependency on the typing backport to Python 3.4 fixes this.

cmd2/AutoCompleter.py
anselor: These functions need a lot of the variables from the parent function. I would need to pass a lot of state arguments to a nested function that's clearly isolated within the scope of this function. It is a bit weird that it lets you access variables declared after this point. Python scoping is a bit weird at times.

cmd2/AutoCompleter.py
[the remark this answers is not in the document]
tleonhardt: I like that idea.

cmd2/AutoCompleter.py
anselor: Yeah, I think it would actually be harder to follow if I break this up differently. I tried to comment clearly what I'm trying to do here. Hopefully that helps.

cmd2/AutoCompleter.py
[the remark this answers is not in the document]
tleonhardt: Oh got it.  Cool.

cmd2/AutoCompleter.py
tleonhardt: Maybe just put a brief comment?

cmd2/AutoCompleter.py
anselor: This is more code that's copied from argparse. I'm preserving their code as much as I can.

cmd2/argparse_completer.py
```diff
@@ -0,0 +1,821 @@
+# coding=utf-8
+"""
+AutoCompleter interprets the argparse.ArgumentParser internals to automatically
```
tleonhardt: Changes all look good to me.

anselor:
No idea if anyone else plans on reviewing this.

cmd2/rl_utils.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def rl_force_redisplay() -> None:
+    """
+    Causes readline to redraw prompt and input line
+    """
+    if not sys.stdout.isatty():
+        return
+
+    if rl_type == RlType.GNU:  # pragma: no cover
```
tleonhardt: This is covered on macOS when the ``gnureadline`` module is installed.

cmd2/rl_utils.py
```diff
@@ -0,0 +1,66 @@
...
+        # rl_forced_update_display() is the proper way to redraw the prompt and line, but we
+        # have to use ctypes to do it since Python's readline API does not wrap the function
+        readline_lib.rl_forced_update_display()
+
+        # After manually updating the display, readline asks that rl_display_fixed be set to 1 for efficiency
+        display_fixed = ctypes.c_int.in_dll(readline_lib, "rl_display_fixed")
+        display_fixed.value = 1
+
+    elif rl_type == RlType.PYREADLINE:  # pragma: no cover
```
tleonhardt: This is covered on Windows when the ``pyreadline`` module is installed, which is a conditional dependency on Windows so it should be.  So this should be covered by our AppVeyor CI builds.

cmd2/rl_utils.py
anselor: The reason I blocked this out is because it can't be covered  by the automated tests because there's isn't a terminal.

anselor: I guess it depends on where we care about seeing the coverage.

tleonhardt: Oh good point.  I think it is reasonable to block it out if we know it won't be hit by any of our CI systems.

tleonhardt:
@kmvanbrunt I'm good to merge this in whenever.  Let me know if you would like to review it or if we should just merge it in.

Issue #287, opened 2018-03-01
Exponential growth in time required to run large numbers of tests


kotfu:
I'm still investigating the root cause of this issue, but it is reproducible.

I discovered this problem while working on https://github.com/tomcatmanager/tomcatmanager, a project which utilizes `cmd2`. A complete run of the test suite for that project makes several hundred calls to `cmd2.Cmd.onecmd_plus_hooks()`.  As the test suite has grown, I have noticed the performance of the test suite degrading in a non-linear way. I finally decided to try and track down why.

I discovered that as the number of calls to `cmd2.Cmd.onecmd_plus_hooks()` increases in a test run, the time required to execute each call grows very quickly. It's easy to duplicate using the test suite in `cmd2`. First, install the `pytest-repeat` module, which allows you to repeat each test in a suite any number of times.

Then run:
```
$ pytest tests/test_argparse.py -v --duration=0
```

On my computer those 18 tests run in less than 1 second. However, in the duration report at the end, which shows the clock time required for each test to run, you can see that some of the tests completed in 0.01 seconds, and others took 0.09 seconds. None of the tests in that file are an order of magnitude more expensive to run than the others, but you could maybe chalk it up to variation in what else is running on your computer.

Now try:
```
$ pytest tests/test_argparse.py -v --duration=0 --count=10
```
This will run each test 10 times. I expect that the time to run the tests would grow linearly with the number of iterations, so it should take perhaps 10 seconds to complete the run. Go ahead and go make some tea, because on my computer it took 464 seconds (almost 8 minutes!) to complete.

Examine the duration report at the end of the test run. Here's what I saw as the 10 slowest tests:
```
============================ slowest test durations ============================
9.28s call     tests/test_argparse.py::test_subcommand_invalid_help[8/10]
9.26s call     tests/test_argparse.py::test_subcommand_invalid_help[4/10]
9.20s call     tests/test_argparse.py::test_subcommand_invalid_help[10/10]
9.19s call     tests/test_argparse.py::test_subcommand_invalid_help[7/10]
9.17s call     tests/test_argparse.py::test_subcommand_invalid_help[9/10]
9.02s call     tests/test_argparse.py::test_subcommand_invalid_help[2/10]
8.78s call     tests/test_argparse.py::test_subcommand_invalid_help[1/10]
8.78s call     tests/test_argparse.py::test_subcommand_invalid_help[5/10]
8.76s call     tests/test_argparse.py::test_subcommand_invalid_help[3/10]
[...]

Issue #37, opened 2017-02-04
Replacing pyparsing with shlex


tleonhardt:
Python's built-in "shlex" module for simple lexical analysis is probably good enough to handle our command-line parsing in cmd2 and it is implemented in C, so its performance would probably be a lot better than pyparsing, which is implemented in pure Python.

ANTLR and PLY also look like potentially superior alternatives to pyparsing, which is slow, dated, and at least in our case requires an overly-complicated grammar.

But shlex should be preferred since it is built-in to Python and not 3rd-party.

shlex:
https://docs.python.org/3.6/library/shlex.html

pyparsing:
http://pyparsing.wikispaces.com

ANTLR:
https://github.com/antlr/antlr4

PLY:
https://github.com/dabeaz/ply

This is a long-term "down the road" idea.

tleonhardt:
@FedericoCeratto 
I've researched shlex a little bit and it looks like it could serve as a a viable replacement for pyparsing.  

The advantages of using shlex over pyparsing would be:

- Reduce dependency on 3rd-party modules which aren't built into the Python standard library
- Likely a significant performance boost in terms of CPU time to parse

But moving to shlex would necessitate a non-backwards compatible feature change in how we deal with comments.  Currently both Python-style ("#...") and C-style ("/* ... */") comments are supported.  If we switch from pyparsing to shlex, then we would only be able to support Python-style comments.  The usage of comments really only makes sense in scripts.

In your opinion, would this minor backwards incompatibility be acceptable?

tleonhardt:
shlex is just a lexer, not a full parser.  So shlex is capable of doing a fine job of tokenizing the input.  But we would still need to write custom logic to properly parse the tokens.

kotfu:
I have no experience with any of the parsing tools linked above, but in 30 minutes of trying to learn a bit about each one, I think PLY seems like the best fit for cmd2. ANTLR doesn't feel very python-ish, given that you use a java program to define the grammar and then have it auto-generate python code for you. shlex only tokenizes and would require us to write a custom parser. pyparsing is slow (for us, and maybe for others too). ply has a pythonesque interface, is well maintained, and doesn't require any C extensions.

tleonhardt:
I think either ``PLY`` or ``shlex`` would be reasonable options, with ``PLY`` probably being the path of least resistance.

Regardless this issue will be somewhat of a pain to work on because it will involve two learning curves:
[...]

PR #370, opened 2018-04-26
Switch parsing logic from pyparsing to shlex

### LINKED ISSUE #327: Explore eliminating POSIX_SHLEX  and STRIP_QUOTES_FOR_NON_POSIX
cmd2 currently defaults to non-POSIX parsing with shlex and strips quotes before passing args to functions.

The basic difference between POSIX and non-POSIX parsing rules in shlex deals with how quotes are processed. See [https://docs.python.org/3/library/shlex.html#parsing-rules for more details.](url)

Previously it was decided to default to non-POSIX since the quoting rules are more intuitive.
From a maintenance perspective, it would be nice to have 1 parsing behavior for our shell and not try to emulate multiple shells. This is especially true when trying to parse tokens for tab-completion.

Are you OK if I remove these flags and preserve the default behavior?

Thoughts? @tleonhardt @kotfu

kotfu:
When I created this branch, I thought we should use `ply` to parse user input. After struggling to make it happen, I switched to `shlex`, and it worked out much better. It's a pain to rename branches, so I didn't.

This PR switches the parsing engine from `pyparsing` to `shlex` as described in #37, which reduced the time required to run the test suite by more than 3 orders of magnitude, resolving issue #287. I also removed POSIX_SHLEX and STRIP_QUOTES_FOR_NON_POSIX to close #327.

There are a few items remaining before this PR should be merged:

- [ ] Consider refactoring `complete()` to use `StatementParser.parse()` instead of `parseline()`. We may need to create a new method in `StatementParser` to get the desired behavior, because tab completion needs to work even if there are current syntax errors in the line of user input.
- [ ] Refactor `parseline()` to use `StatementParser.parse()`
- [x] Remove all use of ``pytest-xdist`` and ``pytest-forked`` from **tox.ini** for running the unit tests
- [x] Update CONTRIBUTING guide to remove all reference to ``pytest-xdist`` and ``pytest-forked``
- [x] Consider doing a full code coverage analysis for all test runs since tests are now much faster
- [ ] Consider adding more versions of Python for Windows tests on AppVeyor
- [x] Update **CHANGELOG.md** with changes accumulated in **SHLEX_TODO.txt**
- [x] delete **SHLEX_TODO.txt**

Due to the extensive nature of these changes, I welcome any additional manual testing that anyone is able to do.


tleonhardt:
Huh I didn't know we can put checkboxes in a PR description.  Cool!

tleonhardt:
Damn @kotfu you sped things up:
screen shot 2018-04-26 at 7 41 03 pm

kotfu:
I did the first part of refactoring the `cmd2.complete()` method to use the new `StatementParser`, including a bespoke method to return just what the completion code needs. However, when I try and graft that new code into `complete()` and replace the call to `parseline()` it breaks most of the submenu completion unit tests: they all complain about not having access to `self.command_parser`. I had a look at the submenu code, and it's pretty complex. Looks like it copies attribute values from the "main" cmd2 class into the submenu class. Before I burn a bunch of time trying to figure it out, I thought I'd reach out for help. 

@kmvanbrunt, any suggestions on how to move forward here?

kmvanbrunt:
@kotfu @tleonhardt 
The deeper integration with readline that I recently added already broke submenu tab completion, even though the unit tests still worked. I just didn't discover it until I had already committed. Unfortunately, I don't have a good way to fix that bug either. Having **cmd2** instances running within **cmd2** instances is pretty problematic.

tleonhardt:
@kmvanbrunt @kotfu 
For whatever it is worth, I'm find with removing the Submenu feature if neither of you need it and it is judged to be too difficult to maintain.  It is pretty complex and at least mildly in conflict with various assumptions made by cmd/cmd2.

kotfu:
OK, my brain is bent. Working on cleaning up **tox.ini**. The following commands pass all unit tests:
```
$ pytest -n2 --cov
$ pytest
```

While this command has 2 failures:
```
$ pytest --cov
```

🔥 😠 ☠️ 

Too sleepy to keep working on it tonight...

tleonhardt:
There is something fundamentally screwy with those two unit tests and pytest.

The following ways of running pytest all work fine:
```Bash
py.test
py.test -n2
py.test -n2 --forked
py.test -n2 --cov
```

However, with the following ways of running py.test, those two tests fail:
```Bash
py.test -k interrupt
py.test --forked
py.test --cov
```

Fixing whatever is wrong with those tests would be ideal, but I don't see anything wrong.  It may be that you just can't easily/reliably unit test signal handlers with pytest.  I suspect there is some way we could do it, but it might be difficult and time consuming to do it right.

I can see two easy solutions to move forward:
* Use ``pytest-xdist`` and **-n2** to run the tests when using --cov
* Remove those two unit tests

I'd be happy with either solution.

cmd2/parsing.py
```diff
@@ -0,0 +1,363 @@
...
+        return '{} {}'.format('' if self.command is None else self.command, self.args).strip()
+
+class StatementParser():
+    """Parse raw text into command components.
+    
+    Shortcuts is a list of tuples with each tuple containing the shortcut and the expansion.
+    """
+    def __init__(
+            self,
+            quotes=['"', "'"],
```
anselor: Some of these common definitions have been duplicated in a few files now. I'm not sure what the most pythonic way to do this is but it seems like having it duplicated is less than ideal. In my bash completion branch I moved some of these common definitions into __init__. That may not be the best place for it - maybe we need a separate file of common defines?

cmd2/parsing.py
```diff
@@ -0,0 +1,363 @@
...
+        self.quotes = quotes
+        self.allow_redirection = allow_redirection
+        self.redirection_chars = redirection_chars
+        self.terminators = terminators
+        self.multilineCommands = multilineCommands
+        self.aliases = aliases
+        self.shortcuts = shortcuts
+
+        self.comment_pattern = re.compile(
+            r'/\*.*?(\*/|$)|\'(?:\\.|[^\\\'])*\'|"(?:\\.|[^\\"])*"',
```
anselor: I see the first pattern searches for `/*` `*/` comment blocks. I'm not following what the remaining patterns are searching for. Perhaps an explanation would be helpful?

anselor: Also, did we support c-style comments before?

kotfu: Just pushed a commit with a robust explanation of the regex:

```
        # this regular expression matches C-style comments and quoted
        # strings, i.e. stuff between single or double quote marks
        # it's used with _comment_replacer() to strip out the C-style
        # comments, while leaving C-style comments that are inside either
        # double or single quotes.
        #
        # this big regular expression can be broken down into 3 regular
        # expressions that are OR'ed together.
        #
        # /\*.*?(\*/|$)          matches C-style comments, with an optional
        #                        closing '*/'. The optional closing '*/' is
        #                        there to retain backward compatibility with
        #                        the pyparsing implementation of cmd2 < 0.9.0
        # \'(?:\\.|[^\\\'])*\'   matches a single quoted string, allowing
        #                        for embedded backslash escaped single quote
        #                        marks
        # "(?:\\.|[^\\"])*"      matches a double quoted string, allowing
        #                        for embedded backslash escaped double quote
        #                        marks
        #
        # by way of reminder the (?:...) regular expression syntax is just
        # a non-capturing version of regular parenthesis. We need the non-
        # capturing syntax because _comment_replacer() looks at match
        # groups
```

And yes, we did support C-style comments before. Since it was relatively easy to continue
to support them, I did so for maximum backwards compatibility.

cmd2/constants.py
```diff
@@ -6,7 +6,7 @@
 
 # Used for command parsing, tab completion and word breaks. Do not change.
 QUOTES = ['"', "'"]
```
tleonhardt: My gut feeling is that this file will grow over time.  There are probably more constants that can be moved here.  But this is a great start.

cmd2/cmd2.py
```diff
@@ -1942,13 +1873,13 @@ def preparse(self, raw):
         return raw
 
     # noinspection PyMethodMayBeStatic
-    def postparse(self, parse_result):
-        """Hook that runs immediately after parsing the command-line but before ``parsed()`` returns a ParsedString.
+    def postparse(self, statement):
```
tleonhardt: Recommend adding type hinting for new/modified code.

cmd2/parsing.py
```diff
@@ -0,0 +1,387 @@
...
+    s = match.group(0)
+    if s.startswith('/'):
+        # treat the removed comment as an empty string
+        return ''
+    else:
+        return s
+
+class Statement(str):
+    """String subclass with additional attributes to store the results of parsing.
+
```
tleonhardt: Recommend configuring IDE or text editor to automatically strip trailing whitespace on save.

cmd2/parsing.py
```diff
@@ -0,0 +1,387 @@
...
+        return '{} {}'.format('' if self.command is None else self.command, self.args).strip()
+
+class StatementParser():
+    """Parse raw text into command components.
+    
+    Shortcuts is a list of tuples with each tuple containing the shortcut and the expansion.
+    """
+    def __init__(
+            self,
+            quotes=['"', "'"],
```
tleonhardt: **WARNING: Strongly recommend against using mutable default arguments because they behave in an extremely non-intuitive manner nearly identical to static local variables in C/C++.**

See this reference for more info:
http://docs.python-guide.org/en/latest/writing/gotchas/

Using a tuple as the default argument value is acceptable.  If it really needs to be a mutable type like a list or dictionary, then set it to None by default and the first thing in the function check if it is None and if so set it to an empty list/dict.

This comment applies to other default argument values as well which are also mutable.

In the unlikely event that you actually want the behavior provided by mutable default arguments, then please comment to explain this.

cmd2/parsing.py
```diff
@@ -0,0 +1,387 @@
...
+            allow_redirection=True,
+            redirection_chars=['|', '<', '>'],
+            terminators=[';'],
+            multilineCommands = [],
```
tleonhardt: Recommend not including spaces for default argument values as this is non-idiomatic.

Also:  Don't use mutable default values for this and the next two arguments.

cmd2/parsing.py
```diff
@@ -0,0 +1,387 @@
...
+import re
+import shlex
+from typing import List, Tuple
+
+import cmd2
+
+BLANK_LINE = '\n'
+
+def _comment_replacer(match):
+    s = match.group(0)
```
tleonhardt: Ideally use a more descriptive variable name than "s"

cmd2/parsing.py
```diff
@@ -0,0 +1,387 @@
...
+    if s.startswith('/'):
+        # treat the removed comment as an empty string
+        return ''
+    else:
```
tleonhardt: ``else`` is unecessary since the ``if`` has a return in it

cmd2/parsing.py
```diff
@@ -0,0 +1,387 @@
...
+    need a place to capture the additional output of the command parsing, so we add
+    our own attributes to this subclass.
+
+    The string portion of the class contains the arguments, but not the command, nor
+    the output redirection clauses.
+    """
+    def __init__(self, object):
+        self.raw = str(object)
+        self.command = None
+        self.multilineCommand = None
```
tleonhardt: Recommend using PEP8-compliant "snake_case" for member variable names for new and/or significantly modified code.  ``cmd2`` didn't originally comply with this.  However, moving in this direction makes us play more nicely with tools like pylint.

This comment applies to a number of member variable names.

cmd2/parsing.py
```diff
@@ -0,0 +1,387 @@
...
+    
+    The cmd module in the standard library passes commands around as a
+    string. To retain backwards compatibility, cmd2 does the same. However, we
+    need a place to capture the additional output of the command parsing, so we add
+    our own attributes to this subclass.
+
+    The string portion of the class contains the arguments, but not the command, nor
+    the output redirection clauses.
+    """
+    def __init__(self, object):
```
tleonhardt: Strongly recommend changing name of argument to something other than **object** since **object** is a built-in type.

cmd2/parsing.py
```diff
@@ -0,0 +1,387 @@
...
+        self.raw = str(object)
```
tleonhardt: Statement inherits from ``str`` base class, but doesn't call the super class ``__init__()``.  Typically it is a good idea to call super class ``__init__()`` unless there is a good reason not to do so.

cmd2/parsing.py
```diff
@@ -0,0 +1,387 @@
...
+        else:
+            multilineCommand = None
+
+        # build Statement object
+        result = Statement(args)
+        result.raw = rawinput
+        result.command = command
+        result.args = args
+        result.terminator = terminator
+        result.inputFrom = inputFrom
```
tleonhardt: **inputFrom** member variable is not defined in the ``Statement`` class ``__init__()``

cmd2/parsing.py
```diff
@@ -0,0 +1,387 @@
...
+            for cur_alias in tmp_aliases:
+                keep_expanding = False
+                if command == cur_alias:
+                    command = self.aliases[cur_alias]
+                    tmp_aliases.remove(cur_alias)
+                    keep_expanding = len(tmp_aliases) > 0
+                    break
+        return command
+
+    def _command_and_args(self, tokens: List[str]) -> Tuple[str, str]:
```
tleonhardt: Recommend making this a static method since ``self`` isn't used within

cmd2/parsing.py
```diff
@@ -0,0 +1,387 @@
+#
+# -*- coding: utf-8 -*-
+"""Statement parsing classes for cmd2"""
+
+import re
+import shlex
+from typing import List, Tuple
+
+import cmd2
```
tleonhardt: Strongly recommend removing import of ``cmd2``.  First off, it isn't used in this file.  Secondly, it results effectively in a circular import loop.

cmd2/cmd2.py
```diff
@@ -1648,6 +1574,11 @@ def complete(self, text, state):
 
                 # Parse the command line
                 command, args, expanded_line = self.parseline(line)
+
```
tleonhardt: Recommending setting IDE/editor to automatically strip trailing whitespace on save

tleonhardt:
@kotfu 
My understanding is that this PR intentionally removes support for input redirection to ease maintenance and because to the best of our knowledge nobody ever used it.

If that is accurate then we should update the feature list in  **README.md** and **setup.py** accordingly.

If that is inaccurate, then input redirection is broken and should be fixed.

tleonhardt:
@kotfu @kmvanbrunt 
The **alias** feature is completely broken on this PR branch.  The fact that our unit tests didn't catch that tells me that we really need to improve the unit tests associated with this feature.

cmd2/cmd2.py
kotfu: Fixed and pushed to `ply` branch.

cmd2/parsing.py
kotfu: Fixed and pushed to `ply` branch.

cmd2/parsing.py
kotfu: Yup. Fixed and pushed to `ply` branch.

cmd2/parsing.py
kotfu: All args fixed and pushed to `ply` branch.

cmd2/parsing.py
kotfu: Fixed and pushed to `ply` branch.

cmd2/parsing.py
kotfu: Fixed and pushed to `ply` branch.

cmd2/parsing.py
kotfu: I had preserved the camel case in an effort to improve backward compatibility.  However, since anyone who is using the pyparsing objects in their own code will have to fix it anyway, this is the best (and maybe only) time to move it to snake case. Fixed and pushed to `ply` branch.

cmd2/parsing.py
kotfu: Fixed and pushed to `ply` branch.

cmd2/parsing.py
kotfu: Fixed and pushed to `ply` branch.

cmd2/parsing.py
kotfu: Fixed and pushed to `ply` branch.

cmd2/parsing.py
kotfu: Import removed and pushed to `ply` branch.

cmd2/cmd2.py
kotfu: Fixed and pushed to `ply` branch.

kotfu:
@tleonhardt Can you give some more information about the broken alias feature in this PR? I have done some manual testing and it always works for me.

tleonhardt:
@kotfu I'm using Python 3.6.5 on macOS as installed from the Anaconda distribution and the conda package manager in a virtual environment.  

On the ever so increasingly misnamed **ply** branch associated with this shlex PR I do the following:
```Bash
(Cmd) alias ls !ls -hal
Alias 'ls' created
(Cmd) ls
*** Unknown syntax: ls
```

However, on the **master** branch, it works:
```Bash
(Cmd) alias ls !ls -hal
Alias 'ls' created
(Cmd) ls
total 840
drwxr-xr-x  33 toddleonhardt  staff   1.0K Apr 29 16:33 .
drwxr-xr-x  36 toddleonhardt  staff   1.1K Feb 20 18:19 ..
-rw-r--r--@  1 toddleonhardt  staff   6.0K Mar 21 19:55 .DS_Store
-rw-r--r--   1 toddleonhardt  staff   104B Apr 28 08:18 .appveyor.yml
...
```

tleonhardt:
I verified that the same problem with the **alias** command is occurring on other versions of Python on macOS and also on Linux.  I haven't tested on Windows yet.

kotfu:
Aliases worked fine, but shortcuts nested in aliases didn't get expanded properly. Fixed, with some new unit tests to ensure it works in the future.

cmd2/parsing.py
```diff
@@ -129,10 +129,35 @@ def __init__(
             re.DOTALL | re.MULTILINE
         )
 
+        # aliases have to be a word, so make a regular expression
```
tleonhardt: aliases now work great

setup.py
```diff
@@ -72,18 +69,7 @@
     ":python_version<'3.5'": ['contextlib2', 'typing'],
 }
 
-if int(setuptools.__version__.split('.')[0]) < 18:
```
tleonhardt: @kotfu 
setuptools 18 dates back to early/mid 2015, so I don't think we really need to worry about people using a version prior to this.  If you think I'm off-base, please feel free to revert this.

I probably should have put this on master, but you made some changes to it as well so I didn't want to create any conflicts.

setup.py
```diff
@@ -98,6 +84,7 @@
     platforms=['any'],
     packages=['cmd2'],
     keywords='command prompt console cmd',
+    python_requires='>=3.4',
```
tleonhardt: I figure we need some way of ensuring nobody tries to install the new version of ``cmd2`` on older un-supported versions of Python.  If we can make the assumption that people will have a pretty new version of setuptools >= 34.4 then this is a really easy way.  If we think it isn't wise to make that assumption, then we could remove this and do something more like:
```Python
import sys
if sys.version_info < (3,4):
    sys.exit('Sorry, Python < 3.4 is not supported by this version of cmd2.  Install cmd2 0.8.x instead')
```

anselor:
I tested merges between bash_completion and pyscript as well as between ply and pyscript and both resulted in very minor merge conflicts that were easy to resolve.

The main conflict is between the bash_completion and ply branches because I got rid of imports in `__init__.py` to improve response time in bash completion. Removing those imports caused changes in imports in all of the tests and examples that are also changed in the ply branch.

The bulk of the changes in both the bash_completion and pyscript branches are in new files. We can probably opt to favor the changes in the ply branch for most of the conflicting changes in the examples and tests. 

@kotfu 
@tleonhardt

tleonhardt:
@anselor 
Since all three PRs are essentially ready to merge (with very minor changes to this one to remove a TODO file), do you have any recommendations regarding merge order of the three that might result in the least pain?

anselor:
Ok, both bash_completion and pyscript are merged into master. I've also merged the result of that into ply and got the unit tests passing again. 

@kotfu please take a look and make sure I didn't jack up your changes.

kotfu:
@anselor Thanks for doing all the hard work to get these three branches to merge. My review and testing of the merged code looks good. I'm going to merge ply into master.

PR #376, opened 2018-04-28
Fixed issue where completion display function was overwritten when a …

kmvanbrunt:
Fixed issue where completion display function was overwritten when a submenu quits.
Fixed issue where submenus did not pass up all completion parameters to complete().

@kotfu 
The second issue fixed might apply to the problem you are having. Look at the code I added in ``complete_submenu()``

@tleonhardt 
Can you check the "Unbound local variable" warnings and see if you have a good way of getting rid of them? I didn't want to add four **# noinspection PyUnboundLocalVariable** statements.

If this fix is OK, then I can add it to the Python 2 branch.

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -71,10 +71,6 @@ elif rl_type == RlType.GNU:
     import ctypes
     from .rl_utils import readline_lib
 
-    # Save address that rl_basic_quote_characters is pointing to since we need to override and restore it
-    rl_basic_quote_characters = ctypes.c_char_p.in_dll(readline_lib, "rl_basic_quote_characters")
-    orig_rl_basic_quote_characters_addr = ctypes.cast(rl_basic_quote_characters, ctypes.c_void_p).value
-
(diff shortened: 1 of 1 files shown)


tleonhardt:
@kmvanbrunt I verified that the "Unbound local variable" warnings are all innocuous false positives.  They occur because PyCharm isn't smart enough to realize that they exist within identical conditionals in different parts of the code.

cmd2/cmd2.py
```diff
@@ -595,10 +591,17 @@ def complete_submenu(_self, text, line, begidx, endidx):
             try:
                 # copy over any shared attributes
                 self._copy_in_shared_attrs(_self)
+                submenu.allow_appended_space = _self.allow_appended_space
+                submenu.allow_closing_quote = _self.allow_closing_quote
+                submenu.display_matches = _self.display_matches
+
```
anselor: Instead of copying these attributes back and forth, why not put these attributes in its own object and pass a reference down the submenus? If submenus modify these values it will be on the same instance the outer layer is using.

Issue #369, opened 2018-04-23
The default imports in __init__ causes significant lag


anselor:
I have a module where I hijack argcomplete's bash completion and tie it into the completion function provided by `AutoCompleter`. Since `AutoCompleter` is a part of `Cmd2` and `__init__` imports `Cmd` and a bunch of other things, it causes about half a second lag on my machine to perform tab completion in bash. If I comment out all of the imports in `__init__` there is no noticeable lag.

Since the next release is going to have a number of big changes anyway, I'm proposing we take out the default imports and force everyone to update their code to import from `cmd2.cmd2`. What do you guys think?

@tleonhardt @kmvanbrunt @kotfu

anselor:
Unfortunately it's the import of `cmd2.Cmd` that causes the lag. If we remove `cmd2.Cmd` from the default imports we may as well remove all of them.

kotfu:
Still formulating an opinion here, but the good news is that in the ply branch, where we no longer import pyparsing, importing `cmd2` takes about 60% as long as it does on the master branch.

tleonhardt:
I'd say go ahead and put in your PR where you make this change and include a demo example showing the use of the ``argparse`` auto-completion at the bash command line via the aid of ``cmd2``.  This does sound like a really nice capability.  I can just be a curmudgeon sometimes when it comes to backward incompatibility ...

I've loved everything else you have done, so I believe there is a high probability that I will love this too.

There will probably be numerous small changes required within the unit test and example code to deal with the default imports no longer being present.

kmvanbrunt:
I've seen the argcomplete/AutoCompleter module and it is very cool, so naturally I support removing the imports. Basically it gives you AutoCompleter in bash when typing arguments for a cmd2-based app.

I also understand how annoying it is for developers to have compatibility issues each time a new version of software comes out. Perhaps we should make 0.9.0 a release candidate version since it will incorporate so many new features. That will give developers time to update their code and us time to debug the new stuff. After we know its stable, we can release the final version as 1.0.0.

Thoughts?

tleonhardt:
I think it will be sufficient to just put big bold warnings in the release notes for 0.9.0 which mention the ways in which backwards compatibility has been broken - no need to call it a "release candidate" or anything like that.  

[...]

PR #377, opened 2018-04-28
Bash completion and __init__.py lag

anselor:
For #369 
I've removed all of the default imports in `__init__.py` and updated all of the examples and unit tests to import directly from the new locations.
This is the first iteration of the bash completion capability. I don't know how to unit test the bash completion code - I'm open to any ideas there. When I have a bit more time I intent to detach this from argcomplete and implement our own bash bridge script that may be a better fit for what our implementation does.


tleonhardt:
The only way I know how to unit test things like Bash tab-completion are using ``pexpect``.  I've used this successfully on my own machine and VMs, however I've never gotten it to work correctly on either of our CI systems (TravisCI or AppVeyor).

cmd2/__init__.py
```diff
@@ -1,5 +1,4 @@
 #
 # -*- coding: utf-8 -*-
-#
-from .cmd2 import __version__, Cmd, set_posix_shlex, set_strip_quotes, AddSubmenu, CmdResult, categorize
-from .cmd2 import with_argument_list, with_argparser, with_argparser_and_unknown_args, with_category
+
+
```
tleonhardt: Recommend configuring IDE/editor to strip trailing newlines after the last one

cmd2/argcomplete_bridge.py
```diff
@@ -0,0 +1,248 @@
...
+            """
+            self.__init__(argument_parser, always_complete_options=always_complete_options, exclude=exclude,
+                          validator=validator, print_suppressed=print_suppressed, append_space=append_space,
+                          default_completer=default_completer)
+
+            if "_ARGCOMPLETE" not in os.environ:
+                # not an argument completion invocation
+                return
+
+            global debug_stream
```
tleonhardt: Global variable **debug_stream** isn't defined at the module level.  Should it be?

Also, is this really a constant?  If so maybe use all uppercase and put it at the module level?

cmd2/argcomplete_bridge.py
```diff
@@ -0,0 +1,248 @@
...
+            try:
+                debug_stream = os.fdopen(9, "w")
+            except:
```
tleonhardt: Never use a bare **except** unless you are 100% sure that is what you really want to do (and it probably isn't).  Ideally, catch a specific tuple of exceptions.  But worst case, catch **Exception**:
```Python
except Exception:
    blah
```

This same comment applies a few lines below.

examples/alias_startup.py
```diff
@@ -6,10 +6,10 @@
 """
 import argparse
```
tleonhardt: I'm pretty sure that both ``argparse`` and ``pyparsing`` are unused imports within this example and should be removed

examples/alias_startup.py
```diff
@@ -6,10 +6,10 @@
...
 
-import cmd2
+from cmd2 import cmd2
 import pyparsing
 
-from cmd2 import with_argument_list, with_argparser, with_argparser_and_unknown_args
+from cmd2.cmd2 import with_argument_list, with_argparser, with_argparser_and_unknown_args
```
tleonhardt: Actually I think all of the imports in this file are unused either than the **from cmd2 import cmd2**

examples/help_categories.py
```diff
@@ -4,7 +4,7 @@
 A sample application for tagging categories on commands.
 """
 
-from cmd2 import Cmd, categorize, __version__, with_argparser, with_category
+from cmd2.cmd2 import Cmd, categorize, __version__, with_argparser, with_category
 import argparse
```
tleonhardt: ``argparse`` is built-in to Python, so should generally be imported at the top before 3rd-party or custom imports.

cmd2/argcomplete_bridge.py
anselor: This is all code inherited from argcomplete. There are just a few places that I've changed to hijack their behavior.

cmd2/argcomplete_bridge.py
anselor: I probably need to import that from argcomplete.

cmd2/argcomplete_bridge.py
tleonhardt: Just because ``argcomplete`` did it in a crappy way doesn't mean you should continue to do so as well ;-)

anselor: I haven't figured out how I want to do it yet, but I intend to borrow the bash completion script, install script, and other supporting scripts from argcomplete and completely drop any dependency we have on them. We're currently only using them to get access to their bash scripts.

tleonhardt: I like the idea of not having even an optional dependency on ``argcomplete`` in the long run ...

cmd2/argcomplete_bridge.py
```diff
@@ -0,0 +1,246 @@
...
+
+            try:
+                argcomplete.debug_stream = os.fdopen(9, "w")
+            except IOError:
+                argcomplete.debug_stream = sys.stderr
+
+            if output_stream is None:
+                try:
+                    output_stream = os.fdopen(8, "wb")
+                except IOError:
```
tleonhardt: Thank you for fixing the bare excepts!

PR #386, opened 2018-05-04
Bash completion

anselor:
Figured out how to detect which time the user pressed tab. Changed the argument hinting to print to stderr after the first tab.

### DIFF
--- a/tests/test_bashcompletion.py
+++ b/tests/test_bashcompletion.py
@@ -32,6 +32,8 @@ if skip_reason3:
 
 skip = skip_reason1 or skip_reason2 or skip_reason3
 
+skip_mac = sys.platform.startswith('dar')
+
 
 actors = ['Mark Hamill', 'Harrison Ford', 'Carrie Fisher', 'Alec Guinness', 'Peter Mayhew',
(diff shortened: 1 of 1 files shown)


cmd2/argcomplete_bridge.py
```diff
@@ -228,15 +228,14 @@ def __call__(self, argument_parser, completer=None, always_complete_options=True
...
-
                 # trick the bash completion into thinking there are 2 completions that are unlikely
                 # to ever match.
-                outstr = outstr.replace('\n', ' ').replace('\t', ' ').replace('    ', ' ').strip()
-                # generate a filler entry that should always sort first
-                filler = ' {0:><{width}}'.format('', width=len(outstr)/2)
-                outstr = ifs.join([filler, outstr])
 
-                output_stream.write(outstr.encode(argcomplete.sys_encoding))
+                comp_type = int(os.environ["COMP_TYPE"])
```
tleonhardt: Strange.  But if it works, cool.

anselor: Yeah. The first tab press comes in as 9. Makes total sense.

PR #391, opened 2018-05-06
Remove check on self.identchars in do_alias()

kotfu:
self.identchars is no longer used by cmd2.

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -2434,13 +2434,6 @@ Usage:  Usage: alias [name] | [<name> <value>]
             name = arglist[0]
             value = ' '.join(arglist[1:])
 
-            # Check for a valid name
-            for cur_char in name:
-                if cur_char not in self.identchars:
-                    self.perror("Alias names can only contain the following characters: {}".format(self.identchars),
-                                traceback_war=False)
-                    return
-
             # Set the alias
             self.aliases[name] = value
             self.poutput("Alias {!r} created".format(name))


kmvanbrunt:
This change allows the following to work.

> (Cmd) alias ">" help
> Alias '>' created

Can you add a restriction that prevents creating aliases containing redirection characters?

Also, we need to fix alias expansion to allow a redirection character without a space.

> (Cmd) alias h help
> Alias 'h' created

This works: `help>out.txt`
This does not work: `h>out.txt`

kotfu:
Given `alias aliasname expansion` as user input:

We shouldn't allow `aliasname` to include redirection characters (i.e. '>' and '|')

Should we allow `aliasname` to include whitespace? I think no.

Should we allow `aliasname` to include terminator characters? I think no.

tleonhardt:
My thoughts on what we should allow in alias names are as follows:
- Redirection chars like > and |:  **no**
- Whitespace:  **no**
- Terminator characters:  **no**

So basically, I'd say don't allow any of that in alias names to keep things as simple as possible.

cmd2/cmd2.py
```diff
@@ -2380,12 +2396,11 @@ def do_alias(self, arglist):
             name = arglist[0]
             value = ' '.join(arglist[1:])
 
-            # Check for a valid name
-            for cur_char in name:
-                if cur_char not in self.identchars:
-                    self.perror("Alias names can only contain the following characters: {}".format(self.identchars),
-                                traceback_war=False)
-                    return
+            # Validate the alias to ensure it doesn't include wierd characters
```
tleonhardt: Weird not wierd

cmd2/parsing.py
```diff
@@ -142,14 +142,34 @@ def __init__(
...
+        #       the string (\Z matches the end of the string even if it
+        #       contains multiple lines)
+        second_group_items = []
+        second_group_items.extend(constants.REDIRECTION_CHARS)
+        second_group_items.extend(terminators)
+        # escape each item so it will for sure get treated as a literal
+        second_group_items = [re.escape(x) for x in second_group_items]
+        # add the whitespace and end of string, not escaped because they
+        # are not literals
+        second_group_items.extend([r'\s', r'\Z'])
```
tleonhardt: Is this the same as invalid_items just with one more item appended?  If so recommend making a copy of the former to minimize duplicated code

kotfu: Good idea. As I think about this more, I think it would be wise to put both of those regexes into the StatementParser class, and to switch the `invalid_alias_pattern` to be `valid_alias_pattern`. That way the logic in using both regexes is the same, i.e. if match: then it's valid.

PR #397, opened 2018-05-10
Extract submenu code to a new project

kotfu:
The submenu code has been extracted and added to a new project at https://github.com/python-cmd2/cmd2-submenu. The new project still needs a little work: better README, deploy to pypi, etc. However, it's far enough along that I think we should remove the code from the main cmd2 project.


cmd2/cmd2.py
```diff
@@ -327,270 +327,6 @@ class EmptyStatement(Exception):
     pass
 
 
-def _pop_readline_history(clear_history: bool=True) -> List[str]:
```
tleonhardt: Looks ok to me

PR #398, opened 2018-05-10
Remove cmd2.Cmd.redirector for #381

### LINKED ISSUE #381: Cmd2.redirector isn't honored any more, either make it work or deprecate it
With the merge of PR #370, `Cmd2.redirector` is no longer honored: the redirector symbol is currently hard-coded as '>'. The documentation still states that you can set `Cmd2.redirector` to something like '->' and have the parsing logic honor that change. We need to either fix the code to match the documentation, or deprecate the attribute and fix the documentation to say that you can't change the redirector.

Related to #392.

kotfu:
This closes #381


cmd2/constants.py
```diff
@@ -4,9 +4,13 @@
...
 
-# Used for command parsing, tab completion and word breaks. Do not change.
+# Used for command parsing, output redirection, tab completion and word
+# breaks. Do not change.
 QUOTES = ['"', "'"]
-REDIRECTION_CHARS = ['|', '>']
+REDIRECTION_PIPE = '|'
+REDIRECTION_OUTPUT = '>'
+REDIRECTION_APPEND = '>>'
+REDIRECTION_CHARS = [REDIRECTION_PIPE, REDIRECTION_OUTPUT]
```
kmvanbrunt: Good use of constants here

cmd2/cmd2.py
```diff
@@ -1820,7 +1816,7 @@ def _redirect_output(self, statement):
 
             # We want Popen to raise an exception if it fails to open the process.  Thus we don't set shell to True.
             try:
-                self.pipe_proc = subprocess.Popen(shlex.split(statement.pipe_to), stdin=subproc_stdin)
+                self.pipe_proc = subprocess.Popen(statement.pipe_to, stdin=subproc_stdin)
```
tleonhardt: I don’t think you want to drop the shlex.split call within the first argument of the call to subprocess.Popen.

kotfu: `statement.pipe_to` gets used in exactly one place, and it's right here. in `parsing.py` we took a list of shlexed tokens and joined them with a space to generate `statement.pipe_to`, then we split them apart to pass it to the subprocess command. I just changed `statement.pipe_to` to have the list of tokens, instead of a string.

Is your concern passing a string to subprocess.Popen()? I agree that would be a concern, but we are passing a list of arguments, just like we were before.

Or is there another concern?

kotfu: We might want to add `shell=True` here anyway and switch to a string so that we can support a command like:

```
(Cmd) help | wc > "count it.txt"
```

tleonhardt: Oh, if ``statement.pipe_to`` is already a list then we are good.  And if you think there are advantages in using ``shell=True`` we can try it.  For some things it can be helpful, but other times it has messed us up.

kotfu: I think for now I'll just leave it as it is with the list. That will be no change in current pipe functionality. If we want to revisit the `shell=True` option later, we can.

PR #400, opened 2018-05-16
Completing alias names in value field of alias command since aliases …

kmvanbrunt:
…can be nested


cmd2/cmd2.py
```diff
@@ -2142,13 +2142,15 @@ def do_alias(self, arglist):
                 errmsg = "Aliases can not contain: {}".format(invalidchars)
                 self.perror(errmsg, traceback_war=False)
 
-
     def complete_alias(self, text, line, begidx, endidx):
         """ Tab completion for alias """
+        alias_names = set(self.aliases.keys())
```
tleonhardt: Looks good

PR #405, opened 2018-05-17
Pyscript updates

anselor:
Applied the data isolation added to the `py` command to the `ipy` command.
Added stdout/stderr suppression by default when executing application commands from a python script. Setting app.cmd_echo = True enables output.


cmd2/pyscript_bridge.py
```diff
@@ -41,54 +40,78 @@ def __bool__(self):
...
+            return self.__dict__[item]
+        else:
+            return getattr(self.inner_stream, item)
+
 
-def _exec_cmd(cmd2_app, func):
+def _exec_cmd(cmd2_app, func: Callable, echo: bool):
     """Helper to encapsulate executing a command and capturing the results"""
-    copy_stdout = CopyStream(sys.stdout)
-    copy_stderr = CopyStream(sys.stderr)
```
tleonhardt: Should echo default to here as well?

cmd2/pyscript_bridge.py
```diff
@@ -41,54 +40,78 @@ def __bool__(self):
...
 
+    def __getattr__(self, item: str):
+        if item in self.__dict__:
+            return self.__dict__[item]
+        else:
+            return getattr(self.inner_stream, item)
+
 
-def _exec_cmd(cmd2_app, func):
+def _exec_cmd(cmd2_app, func: Callable, echo: bool):
```
tleonhardt: Should this ``echo`` also default to **False**?

anselor: Since this is intended to be an internal function I think I'd prefer the caller to be explicit about it.

PR #409, opened 2018-05-19
Autocompleter

### LINKED ISSUE #407: Autocompleter - type invalid text for a subcommand and press tab and ==SUPPRESS== is displayed.
Type invalid text for a subcommand and press tab and ==SUPPRESS== is displayed.
If the text typed doesn't match anything in the subcommand list for a command, autocompleter attempts to print the Hint for the parameter. For a subcommand that comes out as == SUPPRESS ==.

anselor:
Resolves #407

### DIFF
--- a/cmd2/argcomplete_bridge.py
+++ b/cmd2/argcomplete_bridge.py
@@ -6,10 +6,15 @@ try:
     import argcomplete
 except ImportError:  # pragma: no cover
     # not installed, skip the rest of the file
-    pass
-
+    DEFAULT_COMPLETER = None
 else:
     # argcomplete is installed
+
+    # Newer versions of argcomplete have FilesCompleter at top level, older versions only have it under completers
+    try:
+        DEFAULT_COMPLETER = argcomplete.FilesCompleter()
+    except AttributeError:
+        DEFAULT_COMPLETER = argcomplete.completers.FilesCompleter()
 
     from contextlib import redirect_stdout
     import copy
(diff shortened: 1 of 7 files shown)


cmd2/argcomplete_bridge.py
```diff
@@ -6,11 +6,16 @@
     import argcomplete
 except ImportError:  # pragma: no cover
     # not installed, skip the rest of the file
-    pass
-
+    DEFAULT_COMPLETER = None
```
tleonhardt: @anselor Would you please review the changes I made to your PR.  These changes were to support older versions of ``argcomplete`` which have a few API differences including:
- ``FilesCompleter`` is not accessible at the top level
- Fewer keyword arguments to an ``__init__`` method

These older versions of ``argcomplete`` would commonly be found on older versions, in particular Python 3.4.

cmd2/argparse_completer.py
```diff
@@ -472,23 +472,43 @@ def _complete_for_arg(self, action: argparse.Action,
         if action.dest in self._arg_choices:
             arg_choices = self._arg_choices[action.dest]
 
-            if isinstance(arg_choices, tuple) and len(arg_choices) > 0 and callable(arg_choices[0]):
-                completer = arg_choices[0]
+            # if arg_choices is a tuple
```
tleonhardt: Everything looks OK to me

cmd2/argparse_completer.py
```diff
@@ -535,7 +565,10 @@ def _print_action_help(self, action: argparse.Action) -> None:
 
             prefix = '{}{}'.format(flags, param)
         else:
-            prefix = '{}'.format(str(action.dest).upper())
+            if action.dest != SUPPRESS:
```
tleonhardt: Does this PR close Issue #407?  If so, you should mark it as doing so in the description so that it auto-closes on merge.

cmd2/argcomplete_bridge.py
anselor: Looks good to me.

PR #411, opened 2018-05-22
Fixes problem with not passing the parameter hint suppression down to…

anselor:
… sub-commands

Added hint suppression on a per-parameter basis
Added helper function to force an parameter to fall back to bash completion instead of using Cmd2 completion.
  - Hinting is still enabled by default but can be suppressed in the helper function.

### DIFF
--- a/cmd2/argcomplete_bridge.py
+++ b/cmd2/argcomplete_bridge.py
@@ -16,6 +16,7 @@ else:
     except AttributeError:
         DEFAULT_COMPLETER = argcomplete.completers.FilesCompleter()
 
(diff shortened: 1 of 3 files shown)


cmd2/argcomplete_bridge.py
```diff
@@ -16,6 +16,7 @@
     except AttributeError:
         DEFAULT_COMPLETER = argcomplete.completers.FilesCompleter()
 
+    from cmd2.argparse_completer import ACTION_ARG_CHOICES, ACTION_SUPPRESS_HINT
```
tleonhardt: Are you happy with **argcomplete_bridge.py** depending on **argparse_completer.py**?  If so then everything looks fine to me.

PR #413, opened 2018-05-24
Defer certain imports

### LINKED ISSUE #369: The default imports in __init__ causes significant lag

kotfu:
In an attempt to resolve #369, this PR:

- puts a robust set of default imports in `cmd2/__init__.py` so those modules and functions are available when you **import cmd2**. In between version 0.8.5 and where **master** is now, we removed those imports to speed up the wall time for **import cmd2**, which also broke backwards compatibility
- defers many imports of infrequently used libraries until we actually need them. This reduces the number of modules that are imported when you do **import cmd2**.
- reduces by 10% the lines of code in `cmd2.py` so there is less code to parse every time you **import cmd2**
- adds a note to `CHANGELOG.md` about the still unresolved #369.

#369 is still not resolved, and we continue to work that issue, but will go ahead with 0.9.0 without resolving.


cmd2/cmd2.py
```diff
@@ -146,16 +134,6 @@ def categorize(func: Union[Callable, Iterable], category: str) -> None:
     else:
         setattr(func, HELP_CATEGORY, category)
 
-
-def _which(editor: str) -> Optional[str]:
```
tleonhardt: Thanks for moving this to **utils.py**, that is a more appropriate location for it

cmd2/cmd2.py
```diff
@@ -185,6 +163,8 @@ def with_argument_list(func: Callable) -> Callable:
     method. Default passes a string of whatever the user typed.
     With this decorator, the decorated method will receive a list
     of arguments parsed from user input using shlex.split()."""
+    import functools
```
tleonhardt: I don't know enough about how Python importing works ...

If we call this function 1000 times, it just gets imported once right?  i.e. there isn't a performance penalty here is there?

Also, if we then call the function below and hit the same import on line 184 below after calling this function, it doesn't import it again there right?

cmd2/cmd2.py
```diff
@@ -22,30 +22,18 @@
 
 Git repository on GitHub at https://github.com/python-cmd2/cmd2
 """
+# many imports are lazy-loaded when they are needed
```
tleonhardt: Maybe briefly comment on **why** this is being done?

cmd2/cmd2.py
```diff
@@ -22,30 +22,18 @@
...
 import os
 import platform
 import re
 import shlex
-import signal
-import subprocess
 import sys
-import tempfile
-import traceback
 from typing import Callable, List, Optional, Union, Tuple
```
tleonhardt: I think ``Optional`` is now an unused import

cmd2/cmd2.py
```diff
@@ -3185,232 +3125,7 @@ def restore(self):
                 setattr(self.obj, attrib, getattr(self, attrib))
 
 
-class OutputTrap(object):
```
tleonhardt: Thanks for moving all of the transcript testing stuff to its own file.  **cmd2.py** is now feeling a little more focused

cmd2/transcript.py
```diff
@@ -0,0 +1,213 @@
+#
```
tleonhardt: Maybe add a brief comment regarding the overall purpose of the code in this file?

examples/alias_startup.py
```diff
@@ -5,8 +5,7 @@
     2) How to load an initialization script at startup
 """
 
-from cmd2 import cmd2
-
+import cmd2
```
tleonhardt: This general approach of importing looks nice and clean to me

examples/arg_print.py
```diff
@@ -27,11 +26,14 @@ def __init__(self):
         # NOTE: It is critical that the super class __init__ method be called AFTER updating certain parameters which
         # are not settable at runtime.  This includes the shortcuts, multiline_commands, etc.
 
-    def do_aprint(self, arg):
+    def do_aprint(self, statement):
         """Print the argument string this basic command is called with."""
-        self.poutput('aprint was called with argument: {!r}'.format(arg))
+        self.poutput('aprint was called with argument: {!r}'.format(statement))
```
tleonhardt: Nice little improvement

speedup_import.md
```diff
@@ -0,0 +1,99 @@
+# Speedup Import
```
tleonhardt: In the long run I don't think we want this file at the top level.  But it should be fine for now (until we are happy we have the issue resolved).

cmd2/utils.py
```diff
@@ -55,3 +58,89 @@ def namedtuple_with_defaults(typename, field_names, default_values=()):
...
+    else:
+        try:
+            return typ(new)
+        except (ValueError, TypeError):
+            pass
+    print("Problem setting parameter (now %s) to %s; incorrect type?" % (current, new))
+    return current
+
+def which(editor: str) -> Optional[str]:
+    import subprocess
```
tleonhardt: Maybe add a docstring comment?  I know the original probably didn't have one, but now that it is separated from its immediate use it might be helpful.

cmd2/cmd2.py
kotfu: Correct. https://docs.python.org/3/reference/import.html#the-module-cache describes how the import machinery first looks to see if the module is already imported, if so, it just uses the module already in the cache. Think of "import functools" as "import functools if it hasn't already been imported"

cmd2/cmd2.py
kotfu: Yup.

PR #419, opened 2018-05-26
Removed unused function

kmvanbrunt:
The **PyscriptBridge** functionality makes the `onecmd_plus_hooks()` function in `do_py()` unnecessary.

@anselor @tleonhardt 
Someone needs to update [conditional.py](https://github.com/python-cmd2/cmd2/blob/master/examples/scripts/conditional.py) to use **PyscriptBridge** since the script is currently broken.

Please include that fix as part of this PR.

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -2569,15 +2569,6 @@ Usage:  Usage: unalias [-a] name [name ...]
                 except IOError as e:
                     self.perror(e)
 
-            # noinspection PyUnusedLocal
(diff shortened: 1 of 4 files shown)


cmd2/cmd2.py
```diff
@@ -2569,15 +2569,6 @@ def run(filename):
                 except IOError as e:
                     self.perror(e)
 
-            # noinspection PyUnusedLocal
-            def onecmd_plus_hooks(cmd_plus_args):
```
tleonhardt: Thanks for removing this unused function

Issue #410, opened 2018-05-22
Improve hooks to better support plugins


kotfu:
There is a nice system of hooks in place today, which allows subclasses of `cmd2.Cmd()` to inject code at various places in the command processing cycle. See `hooks.rst` in the documentation for full details. With the recent introduction of the idea of plugins, there is a possibility of multiple plugins now wanting to inject code into the same hook.

You can do this today by carefully specifying the order of your mixin classes in your subclassing statement, and ensuring you have the proper `super()` calls in your methods. However, this method is not super intuitive for those who aren't up to speed on the `super()` chain with multiple inheritance.

I propose we implement a more robust system where a plugin can register functions to be called at various points in the command processing cycle. `cmd2.Cmd()` would keep track of all the registered functions for each "hook", and then call them in sequence when the "hook" was activated.

This approach also could make it easier to define command processing plugins. All of the command processing hooks would call functions with a similar parameter and return signature, which would mean you could define a plugin function, and insert it into whichever hook you wanted. Moving your function to a different hook would not require you to modify the function, just the hook to which it was registered.

I think we can maintain backwards compatibility with the current method of calling hooks by fixing `onecmd_plus_hooks()` to call the registered functions for each hook, and then call the hook method (as it currently does). This approach would allow subclasses of `cmd2.Cmd()` to override `postparsing_precmd()`, while still using plugins which would also like to perform work in that same hook.

anselor:
I like this idea.

tleonhardt:
I'm not sure I fully grok how this would be implemented, but I certainly like the sound of it.

kotfu:
New branch **plugin_functions** for this work. Progress coming along nicely. Suggest fixing #420 in this branch too, since they are related.

tleonhardt:
@kotfu I pushed a minor fix for versions of Python prior to 3.6 to your *plugin_functions* branch.

Apparently, prior to Python 3.6, you can't have a trailing comma after the last argument in a method declaration.

PR #420, opened 2018-05-27
Fixed bug where preparse() wasn't called before parsing

### LINKED ISSUE #417: preparse() and postparse() never get called
`preparse()` and `postparse()` methods exist in `cmd2.Cmd` and the docstrings say they get called before and after the user input is parsed. Same methods are referenced in `hooks.rst`. However, these two methods are never called the command processing loop in `onecmd_plus_hooks()`.

We need to either call these methods, or remove them from the code and the documentation.

If we decide to call these methods, we will have to make a decision about when the right time to call `preparse()`, because on a multiline command, the line can be parsed multiple times.

`postparse()` would be called at the same place as the existing `postparsing_precmd()` hook.

tleonhardt:
Deleted ``postparse()`` since it was redundant with ``postparsing_precmd()``
    - Decided to keep ``postparsing_precmd()`` because of its more self-documenting name
- Removed reference to ``postparse()`` in the docs

This partially addresses #417

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@
         * Replaced by default AutoCompleter implementation for all commands using argparse
     * Deleted support for old method of calling application commands with ``cmd()`` and ``self``
     * ``cmd2.redirector`` is no longer supported. Output redirection can only be done with '>' or '>>'
+    * Deleted ``postparse()`` hook since it was redundant with ``postparsing_precmd``
 * Python 2 no longer supported
     * ``cmd2`` now supports Python 3.4+
 * Known Issues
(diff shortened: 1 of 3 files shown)


tleonhardt:
As long as everyone is happy with how I handled this issue, then this is the last thing I wanted to get in before we do a 0.9.0 release.

I think we are ready for an 0.8.6 release as-is from the **python2** branch.

tleonhardt:
@kotfu I think I might revert the ``preparse`` behavior in this PR, but leave the deletion of ``postparse`` if that seems reasonable to you.  And then you can fix ``preparse`` as part of the changes you are working on for #410.

tleonhardt:
@kotfu Are you happy with the ``postparse()`` deletion present in this PR for now?

kotfu:
Yup.

PR #433, opened 2018-06-07
Autocompleter

anselor:
Small adjustments to how completion results are processed to allow optional display of descriptions with completion results in a tabular format.

### DIFF
--- a/cmd2/argparse_completer.py
+++ b/cmd2/argparse_completer.py
@@ -59,6 +59,7 @@ Released under MIT license, see LICENSE file
 
 import argparse
 from colorama import Fore
+import os
 import sys
 from typing import List, Dict, Tuple, Callable, Union
 
@@ -76,6 +77,15 @@ from .rl_utils import rl_force_redisplay
 # define the completion choices for the argument. You may provide a Collection or a Function.
 ACTION_ARG_CHOICES = 'arg_choices'
 ACTION_SUPPRESS_HINT = 'suppress_hint'
+
+
+class CompletionItem(str):
+    def __new__(cls, o, desc='', *args, **kwargs):
+        return str.__new__(cls, o, *args, **kwargs)
+
+    # noinspection PyMissingConstructor,PyUnusedLocal
+    def __init__(self, o, desc='', *args, **kwargs):
+        self.description = desc
 
 
 class _RangeAction(object):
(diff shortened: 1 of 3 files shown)


cmd2/argparse_completer.py
```diff
@@ -422,9 +434,35 @@ def consume_positional_argument() -> None:
                 completion_results = self._complete_for_arg(pos_action, text, line, begidx, endidx, consumed)
                 if not completion_results:
                     self._print_action_help(pos_action)
+                elif len(completion_results) > 1:
+                    completion_results = self._format_completions(pos_action, completion_results)
 
         return completion_results
 
+    def _format_completions(self, action, completions: List[Union[str, CompletionItem]]):
```
tleonhardt: Recommend adding a unit test or two for the new stuff added to argparse_completer.py

examples/tab_autocompletion.py
```diff
@@ -109,6 +109,18 @@ def instance_query_actors(self) -> List[str]:
         """Simulating a function that queries and returns a completion values"""
         return actors
 
+    def instance_query_movie_ids(self) -> List[str]:
```
tleonhardt: It would be better if sub-commands like "video movies add" and "video movies delete" actually mutated the internal list of movies (as viewable by "video movies list").

As is, the example does a great job of demonstrating how tab completion and autocompletion works.  But the sub-commands don't behave in the way I would expect.

At the very least, put a big honking warning at the top of the file that lets users know the subcommands don't actually do what their name implies and exist solely to test tab (auto)completion.  But it would be even better if they actually did what their name suggests.

examples/tab_autocompletion.py
```diff
@@ -109,6 +109,18 @@ def instance_query_actors(self) -> List[str]:
...
+        """Demonstrates showing tabular hinting of tab completion information"""
+        completions_with_desc = []
+
+        for movie_id, movie_entry in self.MOVIE_DATABASE.items():
```
tleonhardt: This is a somewhat complicated feature, but a really nice one.  Please consider updating the Sphinx documentation to refer to it.

You continue to raise the bar of awesome semi-automatic tab-completion. 

You should consider making some sort of public tutorial of how to use all of the wonderful new features.  The examples are useful, but something that provides a little more hand-holding to introduce users to the new features would be amazing.  I'm not sure how discoverable some of the new features are - i.e. they are awesome and powerful, but how do we make sure new users know they exist and give them a basic clue about how to use them?

PR #434, opened 2018-06-07
Fix unexpected redirection behavior

### LINKED ISSUE #431: Unexpected redirection behavior
`help > name with space` creates a file called **name with space**
`help > "name with space"` creates a file called **"name with space"** (quotes included)

I would expect the first command to either output to a file called **name** or fail for invalid syntax.
I would expect the second command to output to a file called **name with space** (unquoted)

tleonhardt:
New behavior:
- help > name with space
    - redirects to a file called "name" (without the quotes)
- help > "name with space"
    - redirects to a file called "name with space" (without the quotes)

TODO:
- [x] Fix Windows unit tests

This closes #431

### DIFF
--- a/cmd2/parsing.py
+++ b/cmd2/parsing.py
@@ -2,6 +2,7 @@
 # -*- coding: utf-8 -*-
 """Statement parsing classes for cmd2"""
 
+import os
 import re
 import shlex
 from typing import List, Tuple
@@ -38,7 +39,7 @@ class Statement(str):
                             from the elements of the list, and aliases and shortcuts
                             are expanded
     :type argv:             list
-    :var terminator:        the charater which terminated the multiline command, if
+    :var terminator:        the character which terminated the multiline command, if
                             there was one
     :type terminator:       str or None
     :var suffix:            characters appearing after the terminator but before output
@@ -49,7 +50,7 @@ class Statement(str):
     :type pipe_to:          list
     :var output:            if output was redirected, the redirection token, i.e. '>>'
     :type output:           str or None
-    :var output_to:         if output was redirected, the destination, usually a filename
+    :var output_to:         if output was redirected, the destination file
     :type output_to:        str or None
 
     """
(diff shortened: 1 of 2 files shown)


tleonhardt:
@kmvanbrunt @anselor @kotfu If any of you have any insight into why these changes make a few of the Windows unit tests associated with output redirection fail, then please enlighten me.  I don't see what is going on (chances are I am just over-tired).

cmd2/parsing.py
```diff
@@ -309,7 +315,12 @@ def parse(self, rawinput: str) -> Statement:
         try:
             output_pos = tokens.index(constants.REDIRECTION_OUTPUT)
             output = constants.REDIRECTION_OUTPUT
-            output_to = ' '.join(tokens[output_pos+1:])
+
+            # Check if we are redirecting to a file
```
tleonhardt: I like the approach of doing the extra string manipulation within the parsing logic which you have taken here.

cmd2/parsing.py
```diff
@@ -297,6 +298,11 @@ def parse(self, rawinput: str) -> Statement:
             pipe_pos = tokens.index(constants.REDIRECTION_PIPE)
             # save everything after the first pipe as tokens
             pipe_to = tokens[pipe_pos+1:]
+
+            for pos, cur_token in enumerate(pipe_to):
+                unquoted_token = utils.strip_quotes(cur_token)
+                pipe_to[pos] = os.path.expanduser(unquoted_token)
```
tleonhardt: It seems like we do need to do something vaguely like this for pipes as well.  But since we are piping to a **process** and not **redirecting** to a file, I don't think this is exactly what we should be doing.

By doing this I think we might be eliminating the ability to pipe to an executable which isn't backed by a file on the user's PATH.  For example, can we pipe to commands built into the shell (Bash, cmd.exe, etc)?

Built-in Bash commands are listed here:
http://manpages.ubuntu.com/manpages/xenial/man7/bash-builtins.7.html

PR #441, opened 2018-06-12
Added pager and pager_chop attributes to Cmd class

### LINKED ISSUE #393: Modify ppaged() to look for PAGER environment variable
On *nix-ish systems, a user can set the environment variable **PAGER** to be their preferred output paginator. We should modify `ppaged()` to use the contents of that environment variable before defaulting back to `less`.

tleonhardt:
Dramatically changed the approach to this PR

Changes include:
- ``pager`` is no longer a settable parameter which can be changed by the end user at runtime
- The ``cmd2.Cmd`` class now has two different attributes:
    - ``pager``
        - contains command-line command to use to display wrapped text using a pager
        - By default set to **more** on Windows or **less -RXF** on POSIX
    - ``pager_chop``
        - contains command-line command to use to display chopped (truncated) text using a pager
        - By default set to **more** on Windows or **less -SRXF** on POSIX
- The ``cmd2.Cmd.ppaged()`` method now has an additional optional argument ``chop`` which defaults to ``False``
    - If ``chop`` is ``True``, then ``self.pager_chop`` gets used as the pager
    - If ``chop`` is ``False``, then ``self.pager`` gets used as the pager

Here I have intentionally leaned towards giving the developer power to ensure a consistent experience for end users.  Along those lines, I removed the use of the PAGER environment variable in setting ``pager`` and also removed it as a runtime settable parameter.  But I also moved towards having both ``pager`` and ``pager_chop`` be instance attributes of the ``cmd2.Cmd`` class so that developers can easily override them in their class which derives from ``cmd2.Cmd`` to encompass whatever pagination behavior they desire.

I couldn't think of any single solution which would somehow magically make all stakeholders perfectly happy.  Hopefully this is a reasonable enough compromise that everyone feels that they can get what they need without too much pain ;-)

This closes #393

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
     * Fixed issue where piping and redirecting did not work correctly with paths that had spaces
 * Enhancements
     * Added ability to print a header above tab-completion suggestions using `completion_header` member
+    * **pager** is now a settable parameter which controls pager command used by ``cmd2.Cmd.ppaged`` method
+        * **pager** looks for *PAGER* environment variable if present or uses sane defaults if not
 
 ## 0.8.8 (TBD, 2018)
 * Bug Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,8 @@
 * Enhancements
     * Added ability to print a header above tab-completion suggestions using `completion_header` member
     * Added ``pager`` and ``pager_chop`` attributes to the ``cmd2.Cmd`` class
-        * ``pager`` looks for *PAGER* environment variable if present or uses sane defaults if not
-        * ``pager_chop`` appends a **-S** flag if ``pager`` starts with **less**
+        * ``pager`` defaults to **less -RXF** on POSIX and **more** on Windows
+        * ``pager_chop`` defaults to **less -SRXF** on POSIX and **more** on Windows
     * Added ``chop`` argument to ``cmd2.Cmd.ppaged()`` method for displaying output using a pager
         * If ``chop`` is ``False``, then ``self.pager`` is used as the pager
         * Otherwise ``self.pager_chop`` is used as the pager
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,12 @@
     * Fixed issue where piping and redirecting did not work correctly with paths that had spaces
 * Enhancements
     * Added ability to print a header above tab-completion suggestions using `completion_header` member
-    * **pager** is now a settable parameter which controls pager command used by ``cmd2.Cmd.ppaged`` method
-        * **pager** looks for *PAGER* environment variable if present or uses sane defaults if not
+    * Added ``pager`` and ``pager_chop`` attributes to the ``cmd2.Cmd`` class
+        * ``pager`` looks for *PAGER* environment variable if present or uses sane defaults if not
+        * ``pager_chop`` appends a **-S** flag if ``pager`` starts with **less**
+    * Added ``chop`` argument to ``cmd2.Cmd.ppaged()`` method for displaying output using a pager
+        * If ``chop`` is ``False``, then ``self.pager`` is used as the pager
+        * Otherwise ``self.pager_chop`` is used as the pager
 
 ## 0.8.8 (TBD, 2018)
 * Bug Fixes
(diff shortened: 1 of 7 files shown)


tleonhardt:
@kotfu Let me know if this meets what you were looking for with Issue #393.

@kmvanbrunt and/or @anselor Please let me know if this change works for your applications.

kmvanbrunt:
The original pager code was written to achieve specific results especially when paging data that wraps. If a user has a pager environment variable, there is a risk that paged data will be presented in a way that the developer did not intend and may even be unreadable. A user may then assume a problem exists with the cmd2 application.

This PR forces developers that want the default behavior to write custom code to enforce it in all cases. Therefore, I would rather we preserve the original code. If a developer wants different behavior, they are free to override the pager.

tleonhardt:
@kmvanbrunt 
You bring up a valid concern.

What would you think if the default reverted to always be what it was before but the cmd2.Cmd.__init__ method accepted an additional optional Boolean parameter which could be manually set to allow the PAGER environment variable to override the default if present?

@kotfu Do you think going to the approach I mention above is worth the slight added complexity?

kmvanbrunt:
@tleonhardt 
I guess I don't understand the use case enough to support adding another flag to **cmd2**. I prefer to keep the code as simple as possible and let users extend the code as needed.

kotfu:
I was the one who opened the issue originally. I did so because when I saw the `ppaged()` method I noticed that it had a hardcoded `less` with a set of options that truncate long lines. It seemed a bit unfriendly to have a default implementation which truncated long lines with no way for the end user to change it.

On *nix-ish systems, there is a long standing tradition that the user can set their desired pager and options using the `PAGER` environment variable. So I thought this might be a more useful default.

Any developer of `cmd2` based applications can create their own implementation of `ppaged()`, or create a new method to buffer output to the screen. If the developer wants to tightly control the options to the pager to get the desired line wrapping behavior, let them do it.

In this case we need to decide whether we want to be developer friendly by default, or user friendly by default. If we chose developer friendly, then we wouldn't add a way for the user to override the pager. We would choose a sensible default behavior for `ppaged()` and if developers want different, they can write their own implementation.

If we chose user friendly, then we would do something like @tleonhardt has implemented here, and if developers want it differently, they can hide the setting and make their own implementation of `ppaged()`.

I have no preference either way, but I don't think truncating long lines of output is a good default behavior.

kotfu:
I forgot to add, I don't think we should add a new parameter to `cmd2.Cmd.init()`, if developers want a different behavior, let them implement the desired behavior in `ppaged()` or write their own output method and use that instead.

tleonhardt:
If I am understanding the different points of view correctly, then one of the big questions comes down to whether or not the pager should truncate long lines by default or not.  From my perspective, the answer is "it depends".  In particular it depends on the type of data being presented.  For displaying tabular data from things like database tables, there is no question in my mind that truncation is the correct answer since that will preserve a single row per line along with column formatting and a good pager like **less** will have a visual indicator that the text has been truncated and allow you to use the left and right arrow keys to scroll to see the truncated text.  However, for other types of data such as long textual prose, truncation would undoubtedly be the preferable alternative.  There may conceivably be applications which wish to display paged output in both of those ways at different times.

So my question is, what is the best way to offer that flexibility and what is the "sanest" default?  

Should we have two separate methods for truncated versus non-truncated paged output?  Such as (names subject to change):
- ``ppaged``
- ``ppaged_trunc``
Where each of those are backed by an attribute within the ``Cmd`` class which sets the actual pager command such as:
- ``pager``
- ``pager_trunc``

Or should we have the single method backed by a single attribute in the ``Cmd`` class and allow the developer to temporarily override and restore that attribute if/when they want to alter the default behavior?

I might have gone too far in making the ``pager`` attribute a runtime user settable parameter.  I definitely wanted to make it a class attribute to make it easier for developers to change.  I would be happy to get rid of the settable parameter if you don't see a strong use case for it.

Speaking to the philosophical point of view, my first priority is to make ``cmd2`` developer friendly - a tool where developers can quickly and easily create feature-rich interactive command-line applications.  But my second priority is to also make it user friendly.  I also imagine that in most of the communities where CLI apps are being developed today that the typical end users might be developers themselves, at least at the scripting level.  If there are cases where developer and user friendliness conflict, then I would prefer to side with the developers, but in the majority of cases I believe we should strive for both.

tleonhardt:
@kmvanbrunt @kotfu I completely changed the approach I was taking towards this PR based on feedback from both of you.  

Please let me know what you think about the current approach.  I tried my best to reach a reasonable compromise which will be capable of meeting everyone's needs without too much inconvenience on anyone.

kmvanbrunt:
I like the final solution. It allows for both vertical and side scrolling.

Can you add a comment to the docstring explaining that chopping is ideal for displaying wide tabular data as is done in utilities such as pgcli?

That way developers will understand the full potential of the `ppaged()` function.

kotfu:
I like this PR now, I think it strikes a good balance to the issues we have discussed.

PR #447, opened 2018-06-22
Corrected how we force a redisplay in pyreadline

kmvanbrunt:
The code that called this was working fine due to other factors, but this is the proper implementation of the function for pyreadline.

### DIFF
--- a/cmd2/rl_utils.py
+++ b/cmd2/rl_utils.py
@@ -75,16 +75,17 @@ elif 'gnureadline' in sys.modules or 'readline' in sys.modules:
         readline_lib = ctypes.CDLL(readline.__file__)
 
 
+# noinspection PyProtectedMember
 def rl_force_redisplay() -> None:
     """
-    Causes readline to redraw prompt and input line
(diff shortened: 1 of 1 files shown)


cmd2/rl_utils.py
```diff
@@ -75,22 +75,24 @@ def pyreadline_remove_history_item(pos: int) -> None:
...
 
         # After manually updating the display, readline asks that rl_display_fixed be set to 1 for efficiency
         display_fixed = ctypes.c_int.in_dll(readline_lib, "rl_display_fixed")
         display_fixed.value = 1
 
     elif rl_type == RlType.PYREADLINE:  # pragma: no cover
-        # noinspection PyProtectedMember
+        # Call _print_prompt() first to set the new location of the prompt
         readline.rl.mode._print_prompt()
+        readline.rl.mode._update_line()
```
tleonhardt: This change looks good to me.  

If it fixes a bug then you should update the ChangeLog and consider back porting the equivalent fix to the python2 branch.

PR #451, opened 2018-06-23
Deprecated CmdResult helper class and promoted CommandResult

### LINKED ISSUE #449: Replace legacy CmdResult class with new CommandResult class
The ``CmdResult`` class in **cmd2.py** should probably be considered legacy and replaced with the ``CommandResult`` class in **pyscript_bridge.py**.  

As long as this wouldn't create an undue burnden on anyone in terms of backwards compatibility, then we should remove all reference to ``CmdResult`` and replace it with reference to ``CommandResult``.

Would anyone object to this?

tleonhardt:
These classes are subtly different, particularly in terms of their truthiness.

CmdResult
- attributes:  out, err, war
- truthy:  if err is falsy

CommandResult
- attributes:  stdout, stderr, data
- truthy:  if err is falsy AND data is not None

So CmdResult was oriented to provide essentially info, error, and warning messages to the user (typically as stirngs), whereas CommandResult is geared towards providing info and error messages to the user as strings in addition to data to the user in a command-specific format which is arbitrary other than it should never be None if the command succeeds.

This closes #449

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@
     * Added ``chop`` argument to ``cmd2.Cmd.ppaged()`` method for displaying output using a pager
         * If ``chop`` is ``False``, then ``self.pager`` is used as the pager
         * Otherwise ``self.pager_chop`` is used as the pager
+* Deprecations
+    * The ``CmdResult`` helper class is *deprecated* and replaced by the improved ``CommandResult`` class
+        * ``CommandResult`` has the following attributes: **stdout**, **stderr**, and **data**
+            * ``CmdResult`` had attributes of: **out**, **err**, **war**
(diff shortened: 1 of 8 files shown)


tleonhardt:
I would appreciate any feedback and opinions.  These two classes are slightly different, so it isn't exactly like the new one is clearly 100% better than the now deprecated one in all cases, though I do think it is better in general.  But they exist primarily in helper roles and the old one isn't used within cmd2 itself, so it kind of feels awkward to leave it around.

As long as everyone thinks this is a good path forward I'm happy with the changes.

tleonhardt:
@kmvanbrunt @kotfu @anselor 
Any opinions on this PR?

kmvanbrunt:
@tleonhardt 
I'm OK merging it.

PR #456, opened 2018-06-26
table_display.py example now uses tableformatter instead of tabulate

### LINKED ISSUE #450: Modify the table_display.py example to use new tableformatter module
The new [tableformatter](https://github.com/python-tableformatter/tableformatter) module has excellent synergy with ``cmd2`` in terms of displaying tabular data in an aesthetically pleasing fashion.

We should update the existing [table_display.py](https://github.com/python-cmd2/cmd2/blob/master/examples/table_display.py) example to use ``tableformatter`` instead of ``tabulate`` and make sure we really display the power of this combination.

tleonhardt:
Changed the table_display.py example to use the tableformatter module which resuls in nicer looking tables.

This closes #450


tleonhardt:
@anselor Please feel free to make any changes you would recommend.  I'm just trying to demonstrate the basics of how to make a nice looking table within ``cmd2``.

tleonhardt:
I pushed some unicode updates to make the unicode features very obvious.  Unfortunately, my improvements appear to have unearthed a bug in the ``wcwidth`` module:
https://github.com/jquast/wcwidth/issues/25

anselor:
I can see why it would return 9 instead of 5. It's probably assuming each of those Hindi characters renders at 2 characters wide.

tleonhardt:
The Hindi discrepancy appears to occur on macOS but not Linux

tleonhardt:
I switched to get the Hindi translation of Mumbai based on Google Translate instead of a Google search which gave a different result and the width is working correctly now.

tleonhardt:
@anselor I added the ability to choose the grid style based on argparse arguments to the commands.  

I am now done with all of the changes I wanted to make to this.  As long as you are happy with it, I think it is good to merge.

If we can figure out how to auto-color text in a row red if the population density is crazy high, then that would be gravy.

PR #458, opened 2018-06-26
Fix bug for startup script with spaces

### LINKED ISSUE #457: The startup_script feature ignores spaces
I tried to add a `startup_script` using the following code:
```python
class MyCmd(Cmd):
    def __init__():
        Cmd.__init__(self, startup_script='/home/ubuntu/my folder/start.rc')
        ...
```

However, after debugging the code, it seems that cmd2 is considering that path as two separated elements: `/home/ubuntu/my` and `folder/start.rc', and then of course the parsing fails when trying to see if that's a real file, as I get the following error:

`ERROR: /home/ubuntu/my does not exist or is not a file`

Any help?

tleonhardt:
This closes #457

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -452,7 +452,7 @@ class Cmd(cmd.Cmd):
         if startup_script is not None:
             startup_script = os.path.expanduser(startup_script)
             if os.path.exists(startup_script) and os.path.getsize(startup_script) > 0:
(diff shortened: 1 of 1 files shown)


tleonhardt:
This is a trivial 1-line obvious bug fix so I am going to merge it in without co-contributor review (though we did have effective review by the reporter of this bug).

PR #460, opened 2018-06-27
Added some explanation of CompletionItem usage in code

anselor:


cmd2/argparse_completer.py
```diff
@@ -81,6 +81,33 @@ def my_completer(text: str, line: str, begidx: int, endidx:int, extra_param: str
 
 
 class CompletionItem(str):
+    """
```
tleonhardt: Thanks for the added comments

PR #462, opened 2018-07-08
Add methods to register hooks, for better plugin support

### LINKED ISSUE #410: Improve hooks to better support plugins

### LINKED ISSUE #417: preparse() and postparse() never get called

### LINKED ISSUE #472: Fix test_which_editor_bad unit test on WSL
``cmd2`` runs fine on Windows Subsystem for Linux (WSL), however there is one unit test failure in the **test_which_editor_bad** test.

This is because it is a test which is skipped on Windows and checks for the supposedly non-existent editor *notepad.exe*.  However, on WSL, that exists!  So this should be a very easy fix. 

One way is to add an extra conditional to this test to detect if this is running on a Linux OS in WSL using something like the following:
```Python
@pytest.mark.skipif(sys.platform.startswith('win'),
                    reason="utils.which function only used on Mac and Linux")
def test_which_editor_bad():
    import platform
    editor = 'notepad.exe'
    path = utils.which(editor)

    # Detect when running under Windows Subsystem for Linux (WSL)
    if 'Microsoft' in platform.release():
        # This is a weird case where notepad.exe IS available on a Linux OS ...
        assert path is not None:
    else:
        # Assert that the editor wasn't found because no notepad.exe on non-Windows systems ;-)
        assert path is None
```

kotfu:
Add methods which allow multiple application and command processing lifecycle hooks to be registered. All registered hooks are called at the appropriate point of the application or command processing lifecycle.

- Closes #410
- Closes #417
- Closes #472 (unrelated)

The primary driver for this change is better support for plugins. Prior to this PR, if a plugin implements an application or command processing hook, the plugin will break if the `cmd2.Cmd` application also implements the same hook.

This PR deprecates all of the current application and command processing lifecycle hooks, except those required to retain backwards compatibility with `cmd.Cmd`. Complete documentation of the new system, including the now deprecated methods can be found in `hooks.rst`.

I have created a new plugin which implements the recently removed abbreviation behavior. The plugin can be found at https://github.com/python-cmd2/cmd2-abbrev. The plugin utilizes the hook methods in this PR.

I have also created a plugin template project (https://github.com/python-cmd2/cmd2-plugin-template) which implements a sample plugin and includes best practices and recommendations for plugin developers.

**TODO**:
- [x] Make sure ``preparse`` gets called for backwards compatibility reasons
- [x] Also update documentation in **hooks.rst** to show when this gets called
- [x] Consider adding a list of hooks analogous to ``preparse``
- [x] Updated documentation in **integrating.rst**


tleonhardt:
@kotfu This is a pretty major change.  I think I will slowly review it over a few days as I try to take time and make sure I understand everything going on.

As it appears to also be a change which breaks backward compatibility with the previous hooks API, it is important that the CHANGELOG gets updated appropriately.

kotfu:
While this is a meaty change, it is fully backward compatible. This PR deprecates (in as much as we deprecate, i.e. wrote it into the docs as deprecated) some of the current hook methods, but they are still fully functional in this PR. No changes were made to existing hook tests (which don't cover the existing hooks very robustly).

Once everyone has had time to weigh in on this approach and we make any necessary changes (or decide to reject it outright), I'll be happy to update the CHANGELOG.

tleonhardt:
I'm continuing to digest the content of this PR.  So far so good.  

My one major gripe is that the docs said that ``preloop`` and ``postloop`` are deprecated, but those are defined by ``cmd``.  So for compatibility with ``cmd``, we can't deprecate those two along with ``precmd`` and ``postcmd`` which you had already indicated.

kotfu:
This is an error in the documentation. I just verified that the code still calls `preloop()` and `postloop()`, but it also has a way to register preload and post loop hooks, so you can have multiple functions called before the command loop begins. I’ll fix the documentation up in the next day or two.

tleonhardt:
@kotfu Need to merge master into this branch and resolve conflicts.

@kmvanbrunt Made some recent changes which improve error checking related to persistent readline history files which conflict with changes made here.

I suspect the conflicts are relatively minor.

tleonhardt:
Dammit.  Sometimes I hate the GitHub review system.  I accidentally deleted all of my review comments right when I was done reviewing.

cmd2/cmd2.py
```diff
@@ -3123,8 +3185,134 @@ def cmdloop(self, intro: Optional[str]=None) -> None:
...
+    ###
+    #
+    # plugin related functions
+    #
+    ###
+    def _initialize_plugin_system(self):
+        """Initialize the plugin system"""
+        self._preloop_hooks = []
+        self._postloop_hooks = []
+        self._postparsing_hooks = []
```
tleonhardt: It seems asymmetric that there is no ``self._preparsing_hooks``.  

Worse, ``preparse`` is currently never called.  The main reason that bug ticket #417 was created is that in the port from a parser based on ``pyparsing`` to one based on ``shlex`` in the 0.9.0 release, we accidentally stopped calling ``preparse``.  Even if we decide to deprecate it, it should be called now for backwards compatibility.

One possible place to call it would be at the beginning of ``_complete_statement``:
```Python
    def _complete_statement(self, line: str) -> Statement:
        """Keep accepting lines of input until the command is complete.

        There is some pretty hacky code here to handle some quirks of
        self.pseudo_raw_input(). It returns a literal 'eof' if the input
        pipe runs out. We can't refactor it because we need to retain
        backwards compatibility with the standard library version of cmd.
        """
        statement = self.statement_parser.parse(self.preparse(line))
```

cmd2/plugin.py
```diff
@@ -0,0 +1,27 @@
+#
+# coding=utf-8
+"""Classes for the cmd2 plugin system"""
+import attr
```
tleonhardt: At some point did Python incorporate the 3rd-party ``attrs`` module into the standard library as built-in module ``attr``??

If so, are we sure it is properly available in Python 3.4 without any additional dependencies?

docs/hooks.rst
```diff
@@ -1,43 +1,312 @@
...
+  changed based on application state and/or command results
+
+
+Command Processing Loop
+-----------------------
+
+When you call `.cmdloop()`, the following sequence of events are repeated until
+the application exits:
+
+1. Output the prompt
```
tleonhardt: Documentation which is similar to this, but now out of date, exists in **integrating.rst** when discussing the behavior of ``onecmd_plus_hooks``.  That file should be updated to be consistent with the new code in this PR.

docs/hooks.rst
```diff
@@ -1,43 +1,312 @@
...
+2. Accept user input
+3. Parse user input into `Statement` object
```
tleonhardt: ``preparse`` should be added here before parsing

docs/hooks.rst
```diff
@@ -1,43 +1,312 @@
...
+- no error message will be displayed to the user
+
+If a hook raises any other exception:
+- no more hooks (except command finalization hooks) of any kind will be called
+- if the command has not yet been executed, it will not be executed
+- the exception message will be displayed for the user.
+
+Specific types of hook methods have additional options as described below.
+
+Postparsing Hooks
```
tleonhardt: If you decide to add a list of preparsing hooks, then add a section for that before this.

docs/hooks.rst
```diff
@@ -1,43 +1,312 @@
...
+
+Your hook will be passed a ``PostcommandData`` object, which has a ``statement``
+attribute that describes the command which was executed. If your postcommand
+hook method gets called, you are guaranteed that the command method was called,
+and that it didn't raise an exception.
+
+If any postcommand hook raises an exception, the exception will be displayed to the user, and no further postcommand hook methods
+will be called. Command finalization hooks, if any, will be called.
+
+After all registered precommand hooks have been called,
```
tleonhardt: This line should probably say **postcommand**, not **precommand**

tleonhardt:
@kmvanbrunt If you get a chance, I would very much appreciate your review of this PR.

cmd2/parsing.py
```diff
@@ -90,11 +121,11 @@ class StatementParser:
...
-            aliases=None,
-            shortcuts=None,
+            allow_redirection: bool = True,
+            terminators: List[str] = None,
+            multiline_commands: List[str] = None,
+            aliases: Dict[str, str] = None,
+            shortcuts: List[Tuple[str, str]] = None,
     ):
         self.allow_redirection = allow_redirection
         if terminators is None:
```
tleonhardt: Given a recent bug fix PR #463, it would be beneficial to have a unit test which covers the case when ``terminators`` is ``None``.

cmd2/cmd2.py
kotfu: `preparse()` is now called by `_complete_statement()`. It is asymmetric that there are no `self._preparsing_hooks`. Here's why.

If it's a multiline command, `preparse()` may not get all the user input. `_complete_statement()` really does two things: a) parse the user input, and b) accept more input in case it's a multiline command the passed string doesn't have a terminator. If we call `preparse()` at the beginning (which is now what it does), we call `preparse()` before we parse the input, but if it's a multiline command and there is no terminator, `preparse()` won't have all the input because it hasn't been entered yet. That's undesirable and misleading.

If we call `preparse()` after we are sure we have all the input, then by definition it can't be before we have parsed the input, because we have to parse it to know if we have it all. It's also misleading to have a hook called `preparse()` which doesn't get called until after everything is parsed.

The good news is that parsing (finally) is fast. If a developer wants to change the user input, they can write a post parsing hook, which gets passed the complete user input, modify it as desired, and reparse it.

By not having a `self._preparsing_hooks` we just make it less confusing. Postparsing hooks have all the input, and can choose to modify it if necessary before the statement is executed.

cmd2/plugin.py
kotfu: No, but I forgot to add it to `setup.py`. Fixed. Primary interest in this library is to make (almost) immutable `Statement` object.

docs/hooks.rst
kotfu: Updated.

docs/hooks.rst
kotfu: Updated.

docs/hooks.rst
kotfu: See above rationale for not having preparsing hooks.

docs/hooks.rst
kotfu: Yup. Fixed.

cmd2/parsing.py
kotfu: Agreed. Added.

cmd2/cmd2.py
tleonhardt: That sounds like a reasonable explanation to me for why it is asymmetric.

setup.py
```diff
@@ -60,7 +60,7 @@
 Topic :: Software Development :: Libraries :: Python Modules
 """.splitlines())))
 
-INSTALL_REQUIRES = ['pyperclip >= 1.5.27', 'colorama']
+INSTALL_REQUIRES = ['pyperclip >= 1.5.27', 'colorama', 'attrs']
```
tleonhardt: Anytime we add a dependency we should updated the README.md accordingly since we have a section which discusses the required dependencies.

tests/test_parsing.py
```diff
@@ -26,10 +26,37 @@ def parser():
     )
     return parser
 
+@pytest.fixture
+def default_parser():
+    parser = StatementParser()
+    return parser
+
+def test_parse_empty_string_default(default_parser):
```
tleonhardt: Thanks for adding the extra unit tests

cmd2/cmd2.py
```diff
@@ -1780,7 +1850,7 @@ def _complete_statement(self, line: str) -> Statement:
         pipe runs out. We can't refactor it because we need to retain
         backwards compatibility with the standard library version of cmd.
         """
-        statement = self.statement_parser.parse(line)
+        statement = self.statement_parser.parse(self.preparse(line))
```
tleonhardt: Thanks for making sure ``preparse`` now properly gets called

cmd2/cmd2.py
```diff
@@ -1635,10 +1642,20 @@ def precmd(self, statement: Statement) -> Statement:
 
     # noinspection PyMethodMayBeStatic
     def preparse(self, raw: str) -> str:
-        """Hook method executed just before the command line is interpreted, but after the input prompt is generated.
+        """Hook method executed before user input is parsed.
+
+        WARNING: If it's a multiline command, `preparse()` may not get all the
```
tleonhardt: Good comment

tleonhardt:
@kotfu 
Before merging, you should update the README to mention the new dependency on ``attrs`` and the CHANGELOG to mention the overall changes.

@kmvanbrunt I know you are pretty busy at work, but if you get a chance, I would value your feedback for this PR.

tleonhardt:
I stared updating the CHANGELOG for you and bumped the version to 0.9.4.

I also updated the README and Sphinx docs to reflect the new dependency on ``attrs``.

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## 0.9.4 (TBD, 2018)
+* Bug Fixes
+    * Fixed bug where ``preparse`` wasn't getting called
+* Enhancements
```
tleonhardt: Please fill in a better description of your changes

CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
...
+    * Improved implementation of lifecycle hooks to to support a plugin
+      framework, see ``docs/hooks.rst`` for details.
```
tleonhardt: This description looks good.  Thanks for updating.

README.md
```diff
@@ -3,6 +3,7 @@ cmd2: a tool for building interactive command line apps
 [![Latest Version](https://img.shields.io/pypi/v/cmd2.svg?style=flat-square&label=latest%20stable%20version)](https://pypi.python.org/pypi/cmd2/)
 [![Build status](https://img.shields.io/travis/python-cmd2/cmd2.svg?style=flat-square&label=unix%20build)](https://travis-ci.org/python-cmd2/cmd2)
 [![Appveyor build status](https://img.shields.io/appveyor/ci/FedericoCeratto/cmd2.svg?style=flat-square&label=windows%20build)](https://ci.appveyor.com/project/FedericoCeratto/cmd2)
+[![VSTS Build status](https://python-cmd2.visualstudio.com/cmd2/_apis/build/status/cmd2-Python%20package-CI?branch=master)](https://python-cmd2.visualstudio.com/cmd2/_build/latest?definitionId=1&branch=master)
```
tleonhardt: The cmd2 PR build/test CI system is now also integrated with Visual Studio Team Services (VSTS).  Currently it is doing builds for Python 3.6 and 3.7 on Linux VMs on a Windows host which is redundant with the TravisCI testing.  But the idea is that eventually we will use VSTS to do our macOS testing once Microsoft's support for that feature is more complete.

tleonhardt:
@kmvanbrunt @anselor
Do either of you have any feedback before this PR gets merged in?

tleonhardt:
Ok @kmvanbrunt wouldn’t have time to review this for another week and a half, so I’m just going to merge this PR.

PR #463, opened 2018-07-11
Use instance attribute for terminators

ksunden:
If the `terminators` is the default, `None`, the function fails because the local variable was not updated.


cmd2/parsing.py
```diff
@@ -163,7 +163,7 @@ def __init__(
         invalid_command_chars = []
         invalid_command_chars.extend(constants.QUOTES)
         invalid_command_chars.extend(constants.REDIRECTION_CHARS)
-        invalid_command_chars.extend(terminators)
+        invalid_command_chars.extend(self.terminators)
```
tleonhardt: @ksunden Thanks for the bug fix.  It would never have made sense to extend that list of strs with None

PR #478, opened 2018-07-21
Add hook example

### LINKED ISSUE #477: Create example showing how hooks work

kotfu:
Closes #477.

### DIFF
--- a/examples/hooks.py
+++ b/examples/hooks.py
@@ -1,0 +1,115 @@
+#!/usr/bin/env python
+# coding=utf-8
+"""
+A sample application for cmd2 demonstrating how to use hooks.
+
+This application shows how to use postparsing hooks to allow case insensitive
+command names, abbreviated commands, as well as allowing numeric arguments to
+follow a command without any intervening whitespace.
+
+"""
+
+import re
+
+from typing import List
+
+import cmd2
+
+
+class CmdLineApp(cmd2.Cmd):
+    """Example cmd2 application demonstrating the use of hooks.
+
+    This simple application has one command, `list` which generates a list
+    of 10 numbers. This command takes one optional argument, which is the
+    number to start on.
+
+    We have three postparsing hooks, which allow the user to enter:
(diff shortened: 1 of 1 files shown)


examples/hooks.py
```diff
@@ -0,0 +1,115 @@
...
+            try:
+                first = int(first)
+            except ValueError:
+                first = 1
+        else:
+            first = 1
+        last = first + 10
+
+        for x in range(first, last):
+            self.poutput(x)
```
tleonhardt: Recommend converting x from an int to a str first since ``poutput`` is intended to receive a ``str`` (even though this will work):
```Python
self.poutput(str(x))
```

examples/hooks.py
```diff
@@ -0,0 +1,115 @@
...
+        data.statement = self.statement_parser.parse("{} {}".format(
+            command,
+            '' if data.statement.args is None else data.statement.args
+        ))
+        return data
+
+    def abbrev_hook(self, data: cmd2.plugin.PostparsingData) -> cmd2.plugin.PostparsingData:
+        """Accept unique abbreviated commands"""
+        target = 'do_' + data.statement.command
+        if not target in dir(self):
```
tleonhardt: Recommend testing for membership using:
```Python
if target not in dir(self):
```

examples/hooks.py
```diff
@@ -0,0 +1,115 @@
...
+    def __init__(self, *args, **kwargs):
+        # sneakily remove the cmd2.Cmd command called load
+        # this lets a user enter a command like "l5" and allows it to
+        # be unambiguous
+        delattr(cmd2.Cmd, "do_load")
+
+        super().__init__(*args, **kwargs)
+
+        # register three hooks
+        self.register_postparsing_hook(self.add_whitespace_hook)
```
tleonhardt: Interesting use case for multiple postparsing hooks.  I like the example, thanks for creating it.

Issue #470, opened 2018-07-12
Version number should be in a single place


tleonhardt:
Currently when we change the version number, this has to be done in 4 places:
- setup.py
- cmd2/cmd2.py
- tests/test_cmd2.py
- docs/conf.py

We should re-architect how the version is done so it only needs to be changed in a single place.  Or worst-case possibly in two places.  Having to change it in 4 places is painful and potentially prone to error.

kotfu:
How about zero places? If we use https://github.com/pypa/setuptools_scm, the version number it automagically is set to the git tag. With this integrated into `cmd2`, all we have to do is tag and upload to pypi.

This approach does mean that there is no meaningful way to "test" the version number, like we do now in `tests/test_cmd2.py`, but I'm not so sure that's helpful anyway.

tleonhardt:
I'd be fine with zero places.  I wasn't aware of [setuptools_scm](https://github.com/pypa/setuptools_scm).

The only reason we have a test for the version number was to make sure we changed it in least the two places associated with that test (on the theory that if you remembered to change it in those two, chances are good you remembered to change it in the other two).  If the version number is in either zero or one place, then I don't see any good reason to "test" that version number.

tleonhardt:
This was closed by #481

PR #481, opened 2018-07-30
Use setuptools_scm for version number

kotfu:
For issue #470, remove hard coded version numbers. Instead use setuptools_scm, which derives the version number from git tags.

`setuptools_scm` has a couple of drawbacks:

- you can't set the version number in any other way than by a git tag. If you want to increment the version number without tagging, you are out of luck.
- `setuptools_scm` is opinionated. If you follow semantic versioning, you must tag minor releases with the a patch level of '0', i.e. '2.0.0', not '2.0'
- if HEAD is not pointing at a tag, you get a long version number, i.e. `0.9.4.dev73+g4da2d37.d20180729`. This is only visible to developers (because only developers should be running untagged versions of `cmd2`), but can be modified if we really want to change it.

The good news is that this PR takes away much of the pain of version increments.


kotfu:
This PR is currently failing on the VSTS build. I think it might be a problem with the build configuration. The test that is failing does:
`
assert cmd2.__version__
`

The only way that `__version__` can generates an `AttributeError` (which is what is happening) is if the `cmd2` package is not installed. I think that on the VSTS build, it's testing the source code in place, not installing the `cmd2` package and testing the installed package.

tleonhardt:
@kotfu I configured the VSTS macOS build to simply use pytest directly, so yeah it is just testing the code in place.  My bad.

In the long run I’m going to switch the VSTS build to be Yaml based so the configuration is in our GitHub repo, but that feature is in preview so I didn’t want to rely on it yet.

In the meantime, see if you can create a free account at visualstudio.com and send me an email with your account name and I’ll add you as an admin to the cmd2 project there so you can tweak the build configuration.

cmd2/__init__.py
```diff
@@ -1,6 +1,14 @@
 #
 # -*- coding: utf-8 -*-
 """This simply imports certain things for backwards compatibility."""
-from .cmd2 import __version__, Cmd, Statement, EmptyStatement, categorize
+
+from pkg_resources import get_distribution, DistributionNotFound
+try:
+    __version__ = get_distribution(__name__).version
```
tleonhardt: So this is finally going to make me abandon my PYTHONPATH hack way of doing ``cmd2`` development and switch to the **pip** way instead using:
```sh
$ cd ~/src/cmd2
$ pip install -e .
```

But that is probably a good thing ...

tleonhardt:
@kotfu I modified the VSTS build so that it first uses **pip** to install ``cmd2`` via:
```sh
$ pip install -e .
```

You should have gotten an email invite to join that VSTS account/project.  

The VSTS macOS build capability is in somewhat of a preview state.  A little down the road we will switch to using a YAML file to define the build to make it obvious what is going on and we will probably test on a couple versions of Python on macOS instead of just one.  But I figured some automated build on macOS is better than no automated build on macOS ...

PR #493, opened 2018-08-01
Fixed case where extra slash was printing when tab completing users on Windows

kmvanbrunt:
I have fixed this same bug in the Python 3 version. That will be in a later pull request.


CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
-## 0.8.8 (TBD, 2018)
+## 0.8.9 (TBD 2018)
```
tleonhardt: Recommend putting this as (August TBD, 2018) since we officially abandon support for Python 2.7 on August 31, 2018.

Also, since this is prepping for the 0.8.9 release for Python 2.7, you should bump the version number to 0.8.9 in this PR in all 4 required places:
* setup.py
* cmd2.py
* docs/conf.py
* tests/test_cmd2.py

NOTE: On *master* now, we have completely changed the version scheme so there is no explicit version number anymore (it is based on the Git Tag).

cmd2.py
```diff
@@ -1694,12 +1694,15 @@ def complete_users():
...
 
             # Windows lacks the pwd module so we can't get a list of users.
-            # Instead we will add a slash once the user enters text that
+            # Instead we will return a result once the user enters text that
             # resolves to an existing home directory.
             if sys.platform.startswith('win'):
                 expanded_path = os.path.expanduser(text)
                 if os.path.isdir(expanded_path):
-                    users.append(text + os.path.sep)
+                    user = text
```
tleonhardt: This looks reasonable

tleonhardt:
Ignore the VSTS build for this PR - it isn't setup to install ``mock`` since it is currently only building for Python 3.7 where that isn't needed.

tleonhardt:
Consider if it would make sense to add a unit test for the new code

cmd2.py
```diff
@@ -229,7 +229,7 @@ def pyreadline_remove_history_item(pos):
         pass
 
 
-__version__ = '0.8.8'
+__version__ = '0.8.9'
```
tleonhardt: Looks good to me

PR #494, opened 2018-08-03
Matches sorted

### LINKED ISSUE #492: Add ability to custom sort tab-completion matches
Currently the `complete()` method sorts tab-completion matches in all cases where there is more than one match. There are cases where a user has a custom sort order and already returns the matches to `complete()` in that preferred order. Add a flag like `self.matches_already_sorted` that overrides the sorting in `complete()`.

kmvanbrunt:
Supporting custom sorting of tab-completion matches. 

This closes #492

### DIFF
--- a/cmd2/utils.py
+++ b/cmd2/utils.py
@@ -157,3 +157,12 @@ def remove_duplicates(list_to_prune: List) -> List:
         temp_dict[item] = None
 
     return list(temp_dict.keys())
+
+
+def alphabetical_sort(list_to_sort: List[str]) -> List[str]:
+    """
+    Sorts a list of strings alphabetically
+    :param list_to_sort: the list being sorted
+    :return: the sorted list
+    """
+    return sorted(list_to_sort, key=str.casefold)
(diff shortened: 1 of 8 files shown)


cmd2/utils.py
```diff
@@ -144,3 +144,25 @@ def is_text_file(file_path: str) -> bool:
...
+    return list(temp_dict.keys())
+
+
+def alphabetical_sort(list_to_sort: List[str]) -> List[str]:
+    """
+    Sorts a list of strings alphabetically
+    :param list_to_sort: the list being sorted
+    :return: the sorted list
+    """
+    return sorted(list_to_sort, key=str.casefold)
```
tleonhardt: Consider doing full-blown Unicode normalization and casefolding here.  It is significantly more complex, but this StackOverflow article has a good explanation of how exactly it should be done: https://stackoverflow.com/questions/40348174/should-i-use-python-casefold

The associated utility functions would likely be useful elsewhere as well.

PR #499, opened 2018-08-21
Copied and reduced the tab_autocompletion example to specifically exe…

anselor:
…rcise updating

the argparse object during init instead of during the class construction.


examples/tab_autocomp_dynamic.py
```diff
@@ -0,0 +1,240 @@
+#!/usr/bin/env python3
+# coding=utf-8
+"""
+A example usage of AutoCompleter with delayed initialization of the argparse object
```
tleonhardt: Thanks for adding this relevant example

PR #501, opened 2018-08-21
Make sure get_all_commands() returns functions

kmvanbrunt:
Previously ``get_all_commands()`` returned any attribute that started with **do_** - it didn't confirm that the attribute was actually callable.

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -2470,7 +2470,7 @@ Usage:  Usage: unalias [-a] name [name ...]
         self._should_quit = True
         return self._STOP_AND_EXIT
 
-    def select(self, opts: Union[str, List[str], List[Tuple[str, Optional[str]]]], prompt: str='Your choice? ') -> str:
+    def select(self, opts: Union[str, List[str], List[Tuple[Any, Optional[str]]]], prompt: str='Your choice? ') -> str:
         """Presents a numbered menu to the user.  Modelled after
            the bash shell's SELECT.  Returns the item chosen.
 
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -1553,7 +1553,8 @@ class Cmd(cmd.Cmd):
 
     def get_all_commands(self) -> List[str]:
         """Returns a list of all commands."""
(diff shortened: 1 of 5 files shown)


cmd2/cmd2.py
```diff
@@ -2470,7 +2472,7 @@ def do_quit(self, _: str) -> bool:
         self._should_quit = True
         return self._STOP_AND_EXIT
 
-    def select(self, opts: Union[str, List[str], List[Tuple[str, Optional[str]]]], prompt: str='Your choice? ') -> str:
+    def select(self, opts: Union[str, List[str], List[Tuple[Any, Optional[str]]]], prompt: str='Your choice? ') -> str:
```
tleonhardt: This fix to the type hinting here is actually breaking the unit test of the build for the Sphinx documentation.  This needs to be fixed one way or the other.  

Our ReadTheDocs configuration currently builds our Sphinx documentation on Linux using Python 3.5, so that is the platform that should be tested on.  The **invoke** options should have one which builds the Sphinx documentation for testing.

tleonhardt:
I updated the CHANGELOG and added specific unit tests for ``get_all_commands()`` and ``get_help_topcis()``.

The Sphinx documentation build unit test is still failing.  I tested it on my Mac using Python 3.5.4 and it passed.  Travis CI is running it on Python 3.5.6 on Linux and it is failing.  I'll try some further testing to replicate the issue.

tleonhardt:
A Google search revealed the error to be a bug in Sphinx 1.7.7 and newer:
https://github.com/sphinx-doc/sphinx/issues/5322

It is scheduled to be fixed in Sphinx 1.7.8.

Perhaps the best approach is to tie the Sphinx unit test to run with the exact version of Sphinx being used by ReadTheDocs.

tox.ini
```diff
@@ -12,7 +12,7 @@ setenv =
 [testenv:docs]
 basepython = python3.5
 deps =
-  sphinx
+  sphinx<1.7.7
```
tleonhardt: For now I am restricting the version of Sphinx used for the documentation unit test because of a bug in the latest version of Sphinx.

PR #505, opened 2018-08-23
Fixed bug where alias was dropping quotes

### LINKED ISSUE #504: Alias command is dropping quotes of arguments
Because `@with_argument_list` returns a list of args in their unquoted state, the `alias` command isn't saving quotes.

Ex: `alias fake !cat "out file.txt"` becomes `alias fake out file.txt`

kmvanbrunt:
Fixes #504 

[kotfu] I think we need to do the following before merging in this PR:

- [x]  as suggested by @tleonhardt, use attrs on Statement object to do immutability
- [x]  make `Statement.arg_list` include the command as the first list element, so it mirrors `Statement.argv` !wontfix. @kmvanbrunt made a convincing argument that `Statement.arg_list` should me more like `Statement.args`, not `Statement.argv`. Documentation updated appropriately.
- [x] Fix documentation and unit tests for `StatementParser.parse_command_only()`. Our changes in this PR infer that `parse_command_only` sets more attributes on the `Statement` object than the documentation says it does, and the unit tests do not check for what is and is not set by that method.

[kotfu] I'll work on all three of these items and get them completed.

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -2214,8 +2214,7 @@ class Cmd(cmd.Cmd):
 
             return stop
 
-    @with_argument_list
-    def do_alias(self, arglist: List[str]) -> None:
+    def do_alias(self, statement: Statement) -> None:
         """Define or display aliases
 
 Usage:  Usage: alias [name] | [<name> <value>]
@@ -2234,22 +2233,27 @@ Usage:  Usage: alias [name] | [<name> <value>]
 
     Example: alias ls !ls -lF
 
-    If you want to use redirection or pipes in the alias, then either quote the tokens with these
-    characters or quote the entire alias value.
+    If you want to use redirection or pipes in the alias, then quote them to avoid
+    the alias command itself from being redirected
 
     Examples:
         alias save_results print_results ">" out.txt
-        alias save_results print_results "> out.txt"
-        alias save_results "print_results > out.txt"
+        alias save_results print_results '>' out.txt
 """
+        # Get alias arguments as a list with quotes preserved
+        alias_arg_list = statement.arg_list
+
         # If no args were given, then print a list of current aliases
-        if not arglist:
+        if not alias_arg_list:
             for cur_alias in self.aliases:
                 self.poutput("alias {} {}".format(cur_alias, self.aliases[cur_alias]))
+            return
+
+        # Get the alias name
+        name = alias_arg_list[0]
 
         # The user is looking up an alias
-        elif len(arglist) == 1:
-            name = arglist[0]
+        if len(alias_arg_list) == 1:
             if name in self.aliases:
                 self.poutput("alias {} {}".format(name, self.aliases[name]))
             else:
@@ -2257,8 +2261,16 @@ Usage:  Usage: alias [name] | [<name> <value>]
 
         # The user is creating an alias
         else:
-            name = arglist[0]
-            value = ' '.join(arglist[1:])
+            # Unquote redirection and pipes
+            index = 1
+            while index < len(alias_arg_list):
+                unquoted_arg = utils.strip_quotes(alias_arg_list[index])
+                if unquoted_arg in constants.REDIRECTION_TOKENS:
+                    alias_arg_list[index] = unquoted_arg
+                index += 1
+
+            # Build the alias value string
+            value = ' '.join(alias_arg_list[1:])
 
             # Validate the alias to ensure it doesn't include weird characters
             # like terminators, output redirection, or whitespace
@@ -2594,24 +2606,20 @@ Usage:  Usage: unalias [-a] name [name ...]
                 param = args.settable[0]
             self.show(args, param)
 
-    def do_shell(self, command: str) -> None:
+    def do_shell(self, statement: Statement) -> None:
         """Execute a command as if at the OS prompt.
 
     Usage:  shell <command> [arguments]"""
-
         import subprocess
-        try:
-            # Use non-POSIX parsing to keep the quotes around the tokens
-            tokens = shlex.split(command, posix=False)
-        except ValueError as err:
-            self.perror(err, traceback_war=False)
-            return
+
+        # Get list of arguments to shell with quotes preserved
+        tokens = statement.arg_list
 
         # Support expanding ~ in quoted paths
         for index, _ in enumerate(tokens):
             if tokens[index]:
-                # Check if the token is quoted. Since shlex.split() passed, there isn't
-                # an unclosed quote, so we only need to check the first character.
+                # Check if the token is quoted. Since parsing already passed, there isn't
+                # an unclosed quote. So we only need to check the first character.
                 first_char = tokens[index][0]
                 if first_char in constants.QUOTES:
                     tokens[index] = utils.strip_quotes(tokens[index])
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -1720,7 +1720,7 @@ class Cmd(cmd.Cmd):
         :return: tuple containing (command, args, line)
         """
         statement = self.statement_parser.parse_command_only(line)
-        return statement.command, statement, statement.command_and_args
+        return statement.command, statement.args, statement.command_and_args
 
     def onecmd_plus_hooks(self, line: str) -> bool:
         """Top-level function called by cmdloop() to handle parsing a line and running the command and all of its hooks.
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -692,8 +692,7 @@ class Cmd(cmd.Cmd):
         elif rl_type == RlType.PYREADLINE:
             readline.rl.mode._display_completions = self._display_matches_pyreadline
 
-    def tokens_for_completion(self, line: str, begidx: int, endidx: int) -> Tuple[Optional[List[str]],
-                                                                                  Optional[List[str]]]:
+    def tokens_for_completion(self, line: str, begidx: int, endidx: int) -> Tuple[List[str], List[str]]:
         """
         Used by tab completion functions to get all tokens through the one being completed
         :param line: the current input line with leading whitespace removed
@@ -710,7 +709,7 @@ class Cmd(cmd.Cmd):
                      The last item in both lists is the token being tab completed
 
                  On Failure
-                    Both items are None
+                    Both items are empty
         """
         import copy
         unclosed_quote = ''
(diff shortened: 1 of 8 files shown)


cmd2/parsing.py
```diff
@@ -105,6 +105,17 @@ def command_and_args(self):
             rtn = None
         return rtn
 
+    @property
+    def arg_list(self) -> Union[List[str], None]:
```
tleonhardt: I think this should always return a list, possibly an empty one but never None.  Doing so would require extra checks in things that call it such as in do-shell now.

tests/test_cmd2.py
```diff
@@ -1761,6 +1761,15 @@ def test_alias(base_app, capsys):
     out = run_cmd(base_app, 'alias fake')
     assert out == normalize('alias fake pyscript')
 
+def test_alias_with_quotes(base_app, capsys):
```
tleonhardt: I recommend creating another unit test where the thing quoted isn’t a redirector.  Keep this one too.

kmvanbrunt: 2 things are quoted in the unit test
alias fake help ">" "out file.txt"

cmd2/parsing.py
kmvanbrunt: Done

tests/test_cmd2.py
tleonhardt: Oops didn’t see that.  Thanks for pointing it out.

cmd2/parsing.py
```diff
@@ -105,6 +105,17 @@ def command_and_args(self):
...
+    @property
+    def arg_list(self) -> List[str]:
+        """
+        Returns a list of the arguments to the command, not including any output
+        redirection or terminators. quoted arguments remain quoted.
+        """
+        if self.args is None:
+            return []
+
+        return self.args.split()
```
tleonhardt: Should this be spit using shlex?

kmvanbrunt:
@kotfu I altered the Statement class in a way that I'd like your approval first. Just to make sure I didn't break anything.

kotfu:
The reason *Statement.args* existed in the first place was because the old pyparsing object had an args attribute. I kept it around to make the transition easier. I also remember thinking that it would be good (I have no reasons why it would be good, but it seemed like a good idea) to have the contents of the string (Statement subclasses str) available as an attribute. I can go either way on whether we think we need *Statement.args*

I like the changes in this PR, and it has made me think about these two attributes on the *Statement* object:

*argv* - the command and arguments as a list; redirection tokens excluded; quotes removed; aliases and shortcuts expanded
*arg_list* - no command, a list of arguments, redirection tokens excluded, quotes not removed, aliases and shortcuts expanded

The last two items are different in two ways: *arg_list* includes quotes, and it doesn't have the command as the first element in the list.

What if we made *arg_list* include the command as the first element. Then the only difference between *argv* and *arg_list* would be whether we stripped the quotes off.

This longish comment yields two questions for us to consider:

1. Do we leave the *args* attribute, with it containing the same value as the *str* the Statement object subclasses?
2. Do we add the command as the first element to *arg_list*?

cmd2/cmd2.py
```diff
@@ -1717,7 +1717,7 @@ def parseline(self, line: str) -> Tuple[str, str, str]:
         :return: tuple containing (command, args, line)
         """
         statement = self.statement_parser.parse_command_only(line)
-        return statement.command, statement.args, statement.command_and_args
+        return statement.command, statement, statement.command_and_args
```
tleonhardt: If I'm reading this correctly, the type hint for the return type of this function should change like so:
```Python3
def parseline(self, line: str) -> Tuple[str, Statement, str]:
```
Basically the middle item in the tuple should change from ``str`` to ``Statement``.

kotfu: This is an interesting question to consider. **parseline** is an override of a standard library `cmd` method. The original method returns a **str**. **Statement** is a subclass of **str**. Should:

- we return a **Statement** object or a **str**?
- If we decide to return a **Statement**, should the type hint match the method we are overriding, or the object we are returning?

kmvanbrunt:
Is there a reason statement.args could be None?

kmvanbrunt:
What about changing Statement.args to a property?

kotfu:
I can't remember. I just went and looked at the unit tests and there are some like
```
assert not statement.args
```
which implies it doesn't matter if it's `None` or an empty string.

But there are some other unit tests which say
```
assert statement.args is None
```
Which implies there is a good reason for it to be None.

So either I'm a crappy unit test writer, and they should all say
```
assert not statement.args
```
or there is a reason it should be None, and I just can't remember what that reason is.

Lemme keep digging.

kmvanbrunt:
@kotfu Yes to your question as to why I would make it a property.
For backward compatibility and to match the command_and_args() property, it would be capable of returning None. Does that seem correct?

tleonhardt:
You should update the CHANGELOG to add a blurb about this bug fix  

Also, please consider my and kotfu's comments above regarding the ``parseline`` type hint and/or return value.

cmd2/parsing.py
```diff
@@ -61,7 +61,7 @@ def __new__(cls,
                 *,
                 raw: str = None,
                 command: str = None,
-                args: str = None,
+                arg_list: List[str] = None,
```
tleonhardt: To me it seems like it would be more for the ``Statement`` class to initialize the ``str`` attributes to an empty string instead of ``None``.  For the ``List`` attributes, it seems like it better to check if they are ``None`` in the ``__new__`` body and if so set them to an empty list.

@kotfu What would you think about making this change?

cmd2/parsing.py
```diff
@@ -29,46 +29,41 @@ class Statement(str):
...
-                            command, otherwise None
-    :type command:          str or None
-    :var args:              the arguments to the command, not including any output
+                            command, otherwise empty
+    :type command:          str
+    :var arg_list:          list of arguments to the command, not including any output
                             redirection or terminators. quoted arguments remain
                             quoted.
-    :type args:             str or None
-    :var: argv:             a list of arguments a la sys.argv. Quotes, if any, are removed
```
tleonhardt: Ok, so it looks like you converted ``argv`` into a property, so it shouldn't have any effect on end users, but perhaps we have reduced storage of redundant data.

cmd2/parsing.py
```diff
@@ -29,46 +29,41 @@ class Statement(str):
...
     :var command:           the command, i.e. the first whitespace delimited word
-    :type command:          str or None
+    :type command:          str
     :var multiline_command: if the command is a multiline command, the name of the
-                            command, otherwise None
-    :type command:          str or None
-    :var args:              the arguments to the command, not including any output
+                            command, otherwise empty
+    :type command:          str
+    :var arg_list:          list of arguments to the command, not including any output
```
tleonhardt: Ok, and ``arg_list`` got added

cmd2/parsing.py
```diff
@@ -29,46 +29,41 @@ class Statement(str):
...
-                pipe_to: str = None,
-                output: str = None,
-                output_to: str = None
+                raw: str = '',
+                command: str = '',
+                arg_list: List[str] = None,
+                multiline_command: str = '',
+                terminator: str = '',
+                suffix: str = '',
+                pipe_to: List[str] = None,
```
tleonhardt: @kmvanbrunt Why did the type of ``pipe_to`` change from ``str`` to ``List[str]``?

kotfu:
I’d like to look at this all again before we merge it. I’ll be able to do it on Monday.

cmd2/parsing.py
kmvanbrunt: I believe I fixed a typo. pipe_to is a list.

cmd2/cmd2.py
```diff
@@ -2237,31 +2236,44 @@ def do_alias(self, arglist: List[str]) -> None:
 
     Example: alias ls !ls -lF
 
-    If you want to use redirection or pipes in the alias, then either quote the tokens with these
-    characters or quote the entire alias value.
+    If you want to use redirection or pipes in the alias, then quote them to prevent
```
tleonhardt: Changes look good to me

cmd2/parsing.py
```diff
@@ -25,91 +25,106 @@ class Statement(str):
...
-        if argv is None:
-            argv = []
-        object.__setattr__(stmt, "argv", argv)
-        object.__setattr__(stmt, "multiline_command", multiline_command)
-        object.__setattr__(stmt, "terminator", terminator)
-        object.__setattr__(stmt, "suffix", suffix)
-        object.__setattr__(stmt, "pipe_to", pipe_to)
-        object.__setattr__(stmt, "output", output)
-        object.__setattr__(stmt, "output_to", output_to)
+        stmt = super().__new__(cls, value)
```
tleonhardt: You are taking a different and interesting approach to immutability which is subtly different than the original, but I think should behave identically in practice.

However, since ``cmd2`` already has a required dependency on ``attrs``, I would recommend considering using the ``attrs`` support for [immutability](http://www.attrs.org/en/stable/examples.html#immutability) for a couple reasons:
1. It is very succinct and readable
1. ``attrs`` is used by a ton of people, so it should be well tested and proven

cmd2/parsing.py
kotfu: Yup, it's better as empty strings and lists, instead of `None`.

cmd2/parsing.py
kotfu: Agreed. I'll make the changes to `Statement` so it uses `attrs` for immutability.

kmvanbrunt:
@kotfu Thanks for your help with the ticket.

I'm not sure if `arg_list` should have `command` as the first element. I understand your idea to have it mirror `argv`, but I don't see `argv` and `arg_list` as being connected. Instead I see `arg_list` being connected to `args` which does not include the command.

Also `argv` has a known meaning because of its name. People would expect command to be included in a list called `argv`. Whereas the name `arg_list` implies it will be a list of just the arguments.

Thoughts?

cmd2/parsing.py
```diff
@@ -25,93 +28,76 @@ class Statement(str):
...
-    :type suffix:           str or None
-    :var pipe_to:           if output was piped to a shell command, the shell command
-                            as a list of tokens
-    :type pipe_to:          list
-    :var output:            if output was redirected, the redirection token, i.e. '>>'
-    :type output:           str or None
-    :var output_to:         if output was redirected, the destination file
-    :type output_to:        str or None
+    # string containing exactly what we input by the user
+    raw = attr.ib(default='', validator=attr.validators.instance_of(str), type=str)
```
tleonhardt: In a perfect world, I would make nearly all of these keyword-only arguments; however ``attrs`` didn't add the **kw_only** argument until the most recent version (18.2.0) which just got released on Sept 1, 2018.  I'd rather not force a required 3rd-party dependency on something that has only been available for about a week because this could make it inconvenient to use ``cmd2`` on older platforms.  However, we may want to consider refactoring this a year or so down the road.

kotfu:
Finally. I think we have everything done. Assuming the unit tests all pass, I think we are ready to merge it.

PR #509, opened 2018-08-24
Added a tag invoke task for adding a Git tag and pushing it to origin

### LINKED ISSUE #469: Add tag creation and commit to invoke task for publishing to PyPI
The **pypi** *invoke* task for publishing a ``cmd2`` release to [PyPi](https://pypi.org/project/cmd2/) currently just publishes to PyPI.  It doesn't create or commit a tag.

This task should also automatically create a tag based on the version number and should commit and push this tag.

tleonhardt:
Added ``tag`` **invoke** task to add a new Git tag and push it to origin.

This is a pretty simple task, but provides a small amount of automation in case a developer doesn't remember the correct Git syntax offhand.

This closes #469 

Given that we now use a version based off of Git tag instead of a hard-coded version in the code, I decided not to make the tag creation part of the ``pypi`` task.  That way a developer can create the appropriate Git tag however she wishes prior to publishing a release to pypi and isn't limited to creating it via **invoke**.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -173,6 +173,15 @@ def clean_all(context):
     pass
 namespace_clean.add_task(clean_all, 'all')
 
+@invoke.task
+def tag(context, name='', message=''):
+    "Add a Git tag and push it to origin"
+    # If a tag was provided on the command-line, then add a Git tag and push it to origin
+    if name:
+        context.run('git tag -a {} -m {!r}'.format(name, message))
+        context.run('git push origin {}'.format(name))
+namespace.add_task(tag)
+
 @invoke.task(pre=[clean_all])
 def sdist(context):
     "Create a source distribution"


tasks.py
```diff
@@ -173,6 +175,34 @@ def clean_all(context):
...
+
+@invoke.task()
+def validatetag(context):
+    "Check to make sure that a tag exists for the current HEAD and it looks like a valid version number"
+    # Validate that a Git tag exists for the current commit HEAD
+    result = context.run("git describe --exact-match --tags $(git log -n1 --pretty='%h')")
+    tag = result.stdout.rstrip()
+
+    # Validate that the Git tag appears to be a valid version number
+    ver_regex = re.compile('(\d+)\.(\d+)\.(\d+)')
```
tleonhardt: Do you think we should be more or less strict when it comes to validating that the Git Tag appears to be a valid version number for publishing to PyPI?

PR #511, opened 2018-09-05
Reduce redundant definition of dependencies

### LINKED ISSUE #486: Reduce or eliminate redundant definitions of dependencies
We currently have dependencies specified in multiple files. Ideally they would only be specified in a single place. I think the right place for these is `setup.py`, but we could/should have a discussion about the single place.

`tox.ini` is the big offender here. Each environment specifies a list of dependencies to install in order to test the package. Instead of:

`deps = 
  mock
  pytest
  pytest-cov
  pytest-mock
`

We should have:

`extras = dev`

which instructs tox to install all the dev dependencies from `setup.py`. `setup.py` is already platform and python version smart about which dependencies are actually needed.

There are some other potential files to examine, but they just install `tox`, so maybe they could be forgiven, but if we can figure out a way...

- appveyor.yml
- travis.yml

tleonhardt:
Modified **setup.py** to include a definition of the extra dependencies required for running unit tests.
This allowed a significant simplification of **tox.ini**.

This closes #486

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -71,7 +71,7 @@ EXTRAS_REQUIRE = {
     # Python 3.4 and earlier require contextlib2 for temporarily redirecting stderr and stdout
     ":python_version<'3.5'": ['contextlib2', 'typing'],
     # Extra dependencies for running unit tests
-    'test': ["argcomplete ; sys_platform!='win32",  # include argcomplete tests where available
+    'test': ["argcomplete ; sys_platform!='win32'",  # include argcomplete tests where available
              "mock ; python_version<'3.6'",  # for python 3.5 and earlier we need the third party mock module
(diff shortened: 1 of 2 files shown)


tleonhardt:
@kotfu Feel free to make further improvements in terms of reducing more redundancy in tox.ini if you wish.  But this seems like a good step in the right direction.

tleonhardt:
This slows the overall testing down slightly because it installs all test tools (including code coverage ones) on all test platforms and runs code coverage analysis on all of them.

However, the upside is it greatly simplifies the tox.ini file.

tleonhardt:
I'd like to understand why it looks like this reduces code coverage, because it should not.

tleonhardt:
This seems to be working well now. I'm just going to merge it and we can tweak it later if we desire.

PR #532, opened 2018-09-21
Fixed case where hints with no text printed

kmvanbrunt:


cmd2/argparse_completer.py
```diff
@@ -654,12 +654,19 @@ def _print_action_help(self, action: argparse.Action) -> None:
             else:
                 prefix = ''
 
+        if action.help is None:
```
tleonhardt: If possible, it would be nice to add a unit test or two to help cover the new code

PR #534, opened 2018-09-23
Fixed several hack classes built to simulate file descriptors

### LINKED ISSUE #497: Fix several hack classes built to simulate file descriptors
We have several hack classes built to simulate regular file descriptors. There are at least:

- transcript.py::OutputTrap
- conftest.py::StdOut
- pyscript_bridge.py::CopyStream

These should really be replaced with `StringIO` objects.

tleonhardt:
Now there is a single class, ``StdSim`` in utils.py, which is intended to simulate stdout and stderr file objects.

This class replaced the following:
- pyscript_bridge.py::CopyStream
- transcript.py::OutputTrap
- conftest.py::StdOut

The new class should also fix a bug related to a ``bytes`` object being written to stdout versus a ``str``.

This closes #497 

TODO:
------
- [x] Add unit test(s) related to sending bytes to a ``StdSim``

### DIFF
--- a/cmd2/utils.py
+++ b/cmd2/utils.py
@@ -246,3 +246,57 @@ def natural_sort(list_to_sort: List[str]) -> List[str]:
     :return: the list sorted naturally
     """
     return sorted(list_to_sort, key=natural_keys)
+
+
+class StdSim(object):
+    """Class to simulate behavior of sys.stdout or sys.stderr.
+
+    Stores contents in internal buffer and optionally echos to the inner stream it is simulating.
+    """
+    class ByteBuf(object):
+        """Inner class which stores an actual bytes buffer and does the actual output if echo is enabled."""
+        def __init__(self, inner_stream, echo: bool = False) -> None:
+            self.byte_buf = b''
+            self.inner_stream = inner_stream
+            self.echo = echo
+
+        def write(self, b: bytes) -> None:
+            """Add bytes to internal bytes buffer and if echo is True, echo contents to inner stream."""
+            self.byte_buf += b
+            if self.echo:
+                self.inner_stream.buffer.write(b)
+
+    def __init__(self, inner_stream, echo: bool = False) -> None:
+        self.buffer = self.ByteBuf(inner_stream, echo)
+        self.inner_stream = inner_stream
+
+    def write(self, s: str) -> None:
+        """Add str to internal bytes buffer and if echo is True, echo contents to inner stream."""
+        b = s.encode()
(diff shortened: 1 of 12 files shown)


tleonhardt:
I would appreciate constructive feedback regarding the design of the ``StdSim`` class itself in **utils.py**.

cmd2/utils.py
```diff
@@ -246,3 +246,57 @@ def natural_sort(list_to_sort: List[str]) -> List[str]:
...
+    """
+    class ByteBuf(object):
+        """Inner class which stores an actual bytes buffer and does the actual output if echo is enabled."""
+        def __init__(self, inner_stream, echo: bool = False) -> None:
+            self.byte_buf = b''
+            self.inner_stream = inner_stream
+            self.echo = echo
+
+        def write(self, b: bytes) -> None:
+            """Add bytes to internal bytes buffer and if echo is True, echo contents to inner stream."""
```
kmvanbrunt: To match stdout, this should raise a TypeError if bytes are not passed in.
Here are actual errors I received from sys.stdout.buffer.write().

- `sys.stdout.buffer.write('test string')`

> TypeError: a bytes-like object is required, not 'str'

- `sys.stdout.buffer.write(5)`

> TypeError: a bytes-like object is required, not 'int'

cmd2/utils.py
```diff
@@ -246,3 +246,57 @@ def natural_sort(list_to_sort: List[str]) -> List[str]:
...
+            self.byte_buf += b
+            if self.echo:
+                self.inner_stream.buffer.write(b)
+
+    def __init__(self, inner_stream, echo: bool = False) -> None:
+        self.buffer = self.ByteBuf(inner_stream, echo)
+        self.inner_stream = inner_stream
+
+    def write(self, s: str) -> None:
+        """Add str to internal bytes buffer and if echo is True, echo contents to inner stream."""
```
kmvanbrunt: To match stdout, this should raise a TypeError if a string is not passed in.
Here are actual errors I received from sys.stdout.write().

- `sys.stdout.write(4)`

> TypeError: write() argument must be str, not int

- `sys.stdout.write(b'test bytes')`

> TypeError: write() argument must be str, not bytes

cmd2/utils.py
tleonhardt: StdSim now raises a TypeError exception if either write() method is passed the wrong type

tleonhardt:
Now that I have addressed the review comments, I'm going to merge this in.

Issue #484, opened 2018-07-30
Add video or animated GIF demonstrating cmd2 capabilities


tleonhardt:
``cmd2`` has some wonderful features, but some of them are more discoverable than others and it is impossible to show how useful of a tool it is for developers with a short code snippet or static screenshot.

The "front page" README.md would benefit from replacing the static screenshot with a short video or animated GIF.

MarkLalor:
This could use some better compression (100MB. Actually, skipping some frames would probably be more effective), but here's a starting point!

Too big to upload directly here but here's an imgur link https://i.imgur.com/8gIQXWq.mp4

tleonhardt:
@MarkLalor Thanks for the video!  I think it is a good starting point which is much nicer than the static image we have on the front page README.md currently.  

Could you possibly submit a PR where the video shows up on the front-page README as a static image with a play button and allows playing it with a click (without having to separately jump to YouTube)?

In the long run it would be nice to have a video which shows off some of the advanced tab-completion features as well, but this is a nice start.

MarkLalor:
@tleonhardt I'm not sure how to pull off a play button with Github markdown. I uploaded a Youtube video just for discussion right now since the gif was over 10MB and Github doesn't allow over 10MB in the comments.

I can improve the video before submitting the PR though. You mentioned some of the more advanced tab-completion features? Which features exactly, and how do you think I can demo them? If there's any other features as well I can include them.

tleonhardt:
Apparently there isn't a built-in way on GitHub or in Markdown to add that play button.  One possibility would be to leave the existing static image (or replace it with a similar one) and to modify the link when it is clicked to go to the video.  Then we would probably have add some text above the image saying something like "Click the image below to watch a brief video demonstrating the capabilities of cmd2".

cmd2 provides good tab-completion wherever it makes sense.  For example, when using shell commands via "!", path completion is automatic.  Another example, when using the **set** command, setting names are completed.  But in general, when using one of cmd2's argparse decorators along with cmd2's ACArgumentParser class, lots of things get automatically completed, such as flags, hints get provided, choices can get filled in, etc.  There are also some helper methods for basic completion, index-based completion, and flag-based completion.  These examples might be good ones to look at for some of these features:

* [tab_completion.py](https://github.com/python-cmd2/cmd2/blob/master/examples/tab_completion.py)
[...]

PR #535, opened 2018-09-24
Replace static image link with link to YouTube video

tleonhardt:
This partly addresses #484 

This PR leaves the same static image on the front-page README, but replaces the embedded link with a link to the YouTube video made by @MarkLalor.  Some text added above the image tells the user to click the image to watch a video.


tleonhardt:
I'm going to merge this in for now since it is safe.  We can work on improving this in the future if we like.

PR #536, opened 2018-09-24
Prevent :param and :return type lines from showing in help summaries

kmvanbrunt:


cmd2/cmd2.py
```diff
@@ -2459,9 +2459,12 @@ def _print_topics(self, header: str, cmds: List[str], verbose: bool) -> None:
                     doc_block = []
                     found_first = False
                     for doc_line in doc.splitlines():
-                        str(doc_line).strip()
-                        if len(doc_line.strip()) > 0:
-                            doc_block.append(doc_line.strip())
+                        stripped_line = doc_line.strip()
+                        if stripped_line.startswith(':'):
+                            # Stop since we've now hit the :param type lines
+                            break
```
tleonhardt: What if someone put relevant documentation after the :param lines?

tleonhardt:
Please consider adding a unit test or two to cover the new code.

cmd2/cmd2.py
kmvanbrunt: How about this?

```Python
# Don't include :param type lines
if stripped_line.startswith(':'):
    if found_first:
        break
elif stripped_line:
    doc_block.append(stripped_line)
    found_first = True
elif found_first:
    break
```

tleonhardt: Yeah, I think that would work fine.  That way it will retain comments either before the :param block or after, but if both are present it will only retain the comments before.  Correct?

kmvanbrunt: Yes, if comments appear before, then found_first is True. So it should break the moment it hits a :param line.

cmd2/cmd2.py
```diff
@@ -2459,13 +2459,17 @@ def _print_topics(self, header: str, cmds: List[str], verbose: bool) -> None:
                     doc_block = []
                     found_first = False
                     for doc_line in doc.splitlines():
-                        str(doc_line).strip()
-                        if len(doc_line.strip()) > 0:
-                            doc_block.append(doc_line.strip())
-                            found_first = True
-                        else:
+                        stripped_line = doc_line.strip()
```
tleonhardt: Looks OK.  Consider adding a unit test or two.

tests/test_cmd2.py
```diff
@@ -46,6 +46,11 @@ def test_base_help_verbose(base_app):
     expected = normalize(BASE_HELP_VERBOSE)
     assert out == expected
 
+    # Make sure :param type lines are filtered out of help summary
+    help_doc = base_app.do_help.__func__.__doc__
```
tleonhardt: Looks good.  Thanks for adding a unit test.

PR #538, opened 2018-09-25
Fix pyscript related unit tests

### LINKED ISSUE #537: Manual unit test failures on some Linux systems
I just manually ran unit tests on a Debian 9.5 system with Python 3.6 installed using conda and encountered some unit test failures in the pyscript tests.

I suspect the changes may relate to the recent StdSim PR where the new StdSim class replaced the old CopyStream class used by pyscript_bridge.py.

I do **NOT** see manual unit test failures on any of the following systems:
* macOS
* Windows
* Ubuntu 18.04
* Ubuntu 18.04 via Windows Subsystem for Linux (WSL)

So this issue is isolated to some particular versions of Linux and/or some particular versions of Python on Linux.

tleonhardt:
This fixes some unit tests which were failing on Debian 9.5  

Actually the unit tests were fundamentally flawed and shouldn't have passed on several systems.  They have been significantly improved.

This closes #537

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -2695,9 +2695,11 @@ Usage:  Usage: unalias [-a] name [name ...]
         Non-python commands can be issued with ``pyscript_name("your command")``.
         Run python code from external script files with ``run("script.py")``
         """
-        from .pyscript_bridge import PyscriptBridge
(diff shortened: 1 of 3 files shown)


tleonhardt:
I'm merging this in now since it fixes an annoying bug.  We can try to improve further in the future if we like.

PR #542, opened 2018-09-25
Fixed issue where only first alias was working

### LINKED ISSUE #541: alias command broken in some circumstances
The ``alias`` command is currently broken on the *master* branch.

A good way to reproduce one case in which it is broken is to use the [alias_startup.py](https://github.com/python-cmd2/cmd2/blob/master/examples/alias_startup.py) example and attempt to run the **pwd** alias as defined in the [.cmd2rc](https://github.com/python-cmd2/cmd2/blob/master/examples/.cmd2rc) startup script.

I believe that the changes in the works on the *macro* branch will fix the current problem.

kmvanbrunt:
This closes #541

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -2160,7 +2160,7 @@ class Cmd(cmd.Cmd):
     def do_alias(self, statement: Statement) -> None:
         """Define or display aliases
 
-Usage:  Usage: alias [name] | [<name> <value>]
+    Usage: alias [name] | [<name> <value>]
     Where:
         name - name of the alias being looked up, added, or replaced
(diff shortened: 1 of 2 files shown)


cmd2/cmd2.py
```diff
@@ -2188,7 +2188,8 @@ def do_alias(self, statement: Statement) -> None:
 
         # If no args were given, then print a list of current aliases
         if not alias_arg_list:
-            for cur_alias in self.aliases:
+            sorted_aliases = utils.alphabetical_sort(list(self.aliases))
```
tleonhardt: Thanks for fixing what I accidentally broke ;-)

PR #543, opened 2018-09-25
Removed dependence on -h when printing argparse command help

### LINKED ISSUE #515: Help does not work with all argparse wrapped functions
For an `argparse` command, the help function adds '-h' to the command to print help text. For most cases, this approach works. However, there are cases in which '-h' won't be available because the **ArgumentParser** has `add_help` set to **False**.

The help function needs to be updated to do something similar to the **argparse_completer** module. It's able to process a line of text and figure out what parser or subparser is active. When it identifies the parser it can just call `parser.print_help()`.

kmvanbrunt:
Fixed #515


cmd2/argparse_completer.py
```diff
@@ -510,6 +510,18 @@ def complete_command_help(self, tokens: List[str], text: str, line: str, begidx:
                             return self.basic_complete(text, line, begidx, endidx, completers.keys())
         return []
 
+    def format_help(self, tokens: List[str]) -> str:
```
tleonhardt: This looks OK to me.  Consider updating the CHANGELOG and/or adding a unit test for this method.

PR #550, opened 2018-09-27
Added macros

kmvanbrunt:
Macros are similar to aliases, but can take arguments when called. These are useful when you need to run a complex command frequently using different arguments.

Arguments are expressed when creating a macro using {#} notation where {1} means the first argument.

The following creates a macro called my_macro that expects two arguments:

> macro create my_macro make_dinner -meat {1} -veggie {2}

When the macro is called, the provided arguments are resolved and the assembled command is run. For example:

> my_macro beef broccoli ---> make_dinner -meat beef -veggie broccoli

Other unrelated changes:
* ``unalias`` command deleted
* ``alias`` command now uses *create*, *delete*, and *list* sub-commands


README.md
```diff
@@ -31,6 +31,7 @@ Main Features
 - Multi-line commands
 - Special-character command shortcuts (beyond cmd's `?` and `!`)
 - Command aliasing similar to bash `alias` command
+- Macros, which are similar to aliases, but can take arguments when called
```
tleonhardt: Pseudo-random train of thought ... we also have a similar feature list in **setup.py** which is what gets published to PyPI that I don't think we have updated in a while.  You should update that as well and make sure it is consistent.

cmd2/argparse_completer.py
```diff
@@ -799,13 +799,15 @@ def get_lines(parts, indent, prefix=None):
...
                     indent = ' ' * (len(prefix) + len(prog) + 1)
                     # Begin cmd2 customization
-                    if opt_parts:
-                        lines = get_lines([prog] + pos_parts, indent, prefix)
-                        lines.extend(get_lines(req_parts, indent))
+                    if req_parts:
+                        lines = get_lines([prog] + req_parts, indent, prefix)
                         lines.extend(get_lines(opt_parts, indent))
+                        lines.extend(get_lines(pos_parts, indent))
+                    elif opt_parts:
```
tleonhardt: Consider adding a unit test to cover the ``elif`` clauses

cmd2/cmd2.py
```diff
@@ -2176,120 +2239,352 @@ def _cmdloop(self) -> bool:
...
-            {
-                1: alias_names,
-                2: list(alias_names | visible_commands)
-            }
-        return self.index_based_complete(text, line, begidx, endidx, index_dict, self.path_complete)
+        # Find all normal arguments
+        arg_list = []
+        normal_matches = re.finditer(MacroArg.macro_normal_arg_pattern, value)
+        max_arg_num = 0
+        num_set = set()
```
lancestarr: A better name would be `arg_nums`

cmd2/cmd2.py
```diff
@@ -2176,120 +2239,352 @@ def _cmdloop(self) -> bool:
...
-        """Unsets aliases
+        while True:
+            try:
+                cur_match = normal_matches.__next__()
 
-    Usage: unalias [-a] name [name ...]
-    Where:
-        name - name of the alias being unset
+                # Get the number between the braces
+                cur_num = int(re.findall(MacroArg.digit_pattern, cur_match.group())[0])
```
lancestarr: A better name would be `arg_num`, in my opinion, although your disdain for variable names that differ by a single letter would then be violated if you took my previous suggestion (so I'd suggest adding `ber` to one of them (as in `number`, not bit error rate))

cmd2/cmd2.py
```diff
@@ -2176,120 +2239,352 @@ def _cmdloop(self) -> bool:
...
+        escaped_matches = re.finditer(MacroArg.macro_escaped_arg_pattern, value)
+
+        while True:
+            try:
+                cur_match = escaped_matches.__next__()
+
+                # Get the number between the braces
+                cur_num = int(re.findall(MacroArg.digit_pattern, cur_match.group())[0])
+
+                arg_list.append(MacroArg(start_index=cur_match.start(), number=cur_num, is_escaped=True))
```
lancestarr: It wasn't immediately obvious to me that we're using `start_index=cur_match.start()` to preserve argument order; a comment might be helpful

cmd2/cmd2.py
```diff
@@ -2176,120 +2239,352 @@ def _cmdloop(self) -> bool:
...
+        """ Deletes macros """
+        if args.all:
+            self.macros.clear()
+            self.poutput("All macros deleted")
+        elif not args.name:
+            self.do_help(['macro', 'delete'])
         else:
-            # Get rid of duplicates
-            arglist = utils.remove_duplicates(arglist)
+            # Get rid of duplicates and strip quotes since the argparse decorator for alias command preserves them
```
lancestarr: Consider replacing _alias_ with _this_; applies elsewhere

cmd2/cmd2.py
```diff
@@ -2176,120 +2239,352 @@ def _cmdloop(self) -> bool:
...
+        """ Lists some or all macros """
+        if args.name:
+            # Get rid of duplicates and strip quotes since the argparse decorator for alias command preserves them
+            names_to_view = [utils.strip_quotes(cur_name) for cur_name in utils.remove_duplicates(args.name)]
+
+            for cur_name in names_to_view:
+                if cur_name in self.macros:
+                    self.poutput("macro create {} {}".format(cur_name, self.macros[cur_name].value))
+                else:
+                    self.perror("Macro {!r} not found".format(cur_name), traceback_war=False)
```
lancestarr: I'm not following why we're using `!r` here; applies to other places where it's just a string

README.md
kmvanbrunt: Done

cmd2/cmd2.py
```diff
@@ -2176,120 +2239,352 @@ def _cmdloop(self) -> bool:
...
+                           "  To use the literal string {1} in your command, escape it this way: {{1}}.\n"
+                           "\n"
+                           "  An argument number can be repeated in a macro. In the following example the\n"
+                           "  first argument will populate both {1} instances.\n"
+                           "\n"
+                           "    macro create ft file_taxes -p {1} -q {2} -r {1}\n"
+                           "\n"
+                           "  To quote an argument in the resolved command, quote it during creation.\n"
+                           "\n"
+                           "    macro create backup !cp \"{1}\" \"{1}.orig\"\n"
```
lancestarr: This would be a good example to use when commenting on why we preserve quotes in the `@with_argparse` statements for alias and macro

cmd2/cmd2.py
```diff
@@ -2176,120 +2239,352 @@ def _cmdloop(self) -> bool:
...
+                           "\n"
+                           "  Macros can resolve into commands, aliases, and macros. Thus it is possible\n"
```
lancestarr: Having "thus" and "so" is bugging me. How about:

```Python
        "  Be careful! Since macros can resolve into commands, aliases, and macros,\n" 
        "  it is possible to create a macro that results in infinite recursion.\n" 
```

cmd2/cmd2.py
```diff
@@ -2176,120 +2239,352 @@ def _cmdloop(self) -> bool:
...
+    macro_create_parser.set_defaults(func=macro_create)
+
+    # macro -> delete
+    macro_delete_help = "delete macros"
+    macro_delete_description = "Delete specified macros or all macros if --all is used"
+    macro_delete_parser = macro_subparsers.add_parser('delete', help=macro_delete_help,
+                                                      description=macro_delete_description)
+    setattr(macro_delete_parser.add_argument('name', nargs='*', help='macro to delete'),
+            ACTION_ARG_CHOICES, get_macro_names)
+    macro_delete_parser.add_argument('-a', '--all', action='store_true', help="all macros will be deleted")
```
lancestarr: Consider `help="delete all macros"`

cmd2/parsing.py
```diff
@@ -247,23 +296,32 @@ def __init__(
...
-        separated string of characters that can not appear in a command.
+        If word is not a valid command, return False and error text
         This string is suitable for inclusion in an error message of your
         choice:
 
-        valid, invalidchars = statement_parser.is_valid_command('>')
+        valid, errmsg = statement_parser.is_valid_command('>')
         if not valid:
-            errmsg = "Aliases can not contain: {}".format(invalidchars)
+            errmsg = "Aliases {}".format(errmsg)
```
lancestarr: Consider _Alias_

cmd2/parsing.py
```diff
@@ -247,23 +296,32 @@ def __init__(
...
         """
         valid = False
 
-        errmsg = 'whitespace, quotes, '
+        if not word:
+            return False, 'cannot be an empty string'
+
+        errmsg = 'cannot start with a shortcut: '
+        errmsg += ', '.join(shortcut for (shortcut, expansion) in self.shortcuts)
```
lancestarr: These two lines are premature; just do them inside the `if` below

cmd2/utils.py
```diff
@@ -304,3 +304,15 @@ def __getattr__(self, item: str):
             return self.__dict__[item]
         else:
             return getattr(self.inner_stream, item)
+
+
+def unquote_redirection_tokens(args: List[str]) -> None:
+    """
+    Used to unquote redirection tokens in a list of command line arguments
```
lancestarr: Consider "Unquote redirection tokens in a list of command-line arguments"

cmd2/cmd2.py
```diff
@@ -2176,120 +2239,352 @@ def _cmdloop(self) -> bool:
...
+            return
 
-    With two or more arguments, 'alias' creates or replaces an alias.
+        utils.unquote_redirection_tokens(args.command_args)
 
-    Example: alias ls !ls -lF
+        # Build the alias value string
+        value = args.command
+        for cur_arg in args.command_args:
+            value += ' ' + cur_arg
```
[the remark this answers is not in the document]
kmvanbrunt: Won't that result in a space after args.command if args.command_args is empty?

cmd2/cmd2.py
```diff
@@ -2006,24 +2028,65 @@ def onecmd(self, statement: Union[Statement, str]) -> Optional[bool]:
...
+        :param statement: the parsed statement from the command line
+        :return: a flag indicating whether the interpretation of commands should stop
+        """
+        try:
+            macro = self.macros[statement.command]
+        except KeyError:
+            raise KeyError("{} is not a macro".format(statement.command))
+
+        # Confirm we have the correct number of arguments
+        if len(statement.arg_list) != macro.required_arg_count:
```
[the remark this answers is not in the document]
kmvanbrunt: Currently I require the exact number of expected arguments, since its like calling a function.
Should I allow extra arguments and just append them to the end of the resolved command?

cmd2/cmd2.py
```diff
@@ -2176,120 +2239,352 @@ def _cmdloop(self) -> bool:
...
+                           "Examples:\n"
+                           "  alias ls !ls -lF\n"
+                           "  alias create show_log !cat \"log file.txt\"\n"
+                           "  alias create save_results print_results \">\" out.txt\n")
+
+    alias_create_parser = alias_subparsers.add_parser('create', help=alias_create_help,
+                                                      description=alias_create_description,
+                                                      epilog=alias_create_epilog)
+    setattr(alias_create_parser.add_argument('name', help='name of this alias'),
+            ACTION_ARG_CHOICES, get_commands_aliases_and_macros_for_completion)
```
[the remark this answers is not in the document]
kmvanbrunt: This should just be aliases and commands. I am fixing it.

cmd2/cmd2.py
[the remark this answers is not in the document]
lancestarr: No, I think your approach is good. Just add a comment.

```Python
    # For macros, every argument must be provided and there can be no extra arguments.
```

cmd2/cmd2.py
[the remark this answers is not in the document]
lancestarr: Good point.

```Python
value = args.command + (' ' + ' '.join(args.command_args) if args.command_args else '')
```

or

```Python
value = args.command

if args.command_args:
    value +=' ' + ' '.join(args.command_args)
```

cmd2/cmd2.py
lancestarr: Then consider replacing _alias_ with _the top-level_ or _the macro_

cmd2/cmd2.py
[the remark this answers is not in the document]
lancestarr: So I could do `alias create alias !ls` and disable my `alias` command?

cmd2/cmd2.py
```diff
@@ -1618,6 +1618,12 @@ def get_macro_names(self) -> List[str]:
         """Return a list of macro names."""
         return list(self.macros)
 
+    def get_commands_and_aliases_for_completion(self) -> List[str]:
+        """Return a list of visible commands and aliases for tab completion"""
+        visible_commands = set(self.get_visible_commands())
+        alias_names = set(self.get_alias_names())
+        return list(visible_commands | alias_names | macro_names)
```
lancestarr: Consider removing ` | macro_names`

cmd2/cmd2.py
```diff
@@ -2176,120 +2245,355 @@ def _cmdloop(self) -> bool:
...
+                           "Examples:\n"
+                           "  alias ls !ls -lF\n"
+                           "  alias create show_log !cat \"log file.txt\"\n"
+                           "  alias create save_results print_results \">\" out.txt\n")
+
+    alias_create_parser = alias_subparsers.add_parser('create', help=alias_create_help,
+                                                      description=alias_create_description,
+                                                      epilog=alias_create_epilog)
+    setattr(alias_create_parser.add_argument('name', help='name of this alias'),
+            ACTION_ARG_CHOICES, get_commands_and_aliases_for_completion)
```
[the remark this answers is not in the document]
lancestarr: I suggest no tab completion for macro and alias names

README.md
tleonhardt: I'm not aware of a way to link them when using Markdown other than writing Python code in setup.py to parse the README.md file and pull out the lines we want.  Actually, this is probably totally do-able and a good idea.

If using reStructuredText, there are some other things you can do since it is more powerful and flexible than Markdown, but I would strongly prefer to keep the README in Markdown format due to the simplicity and lower learning curve.

cmd2/cmd2.py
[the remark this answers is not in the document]
tleonhardt: If we aren't already doing a check, I think we should not allow users to create aliases or macros with the same name as a command; otherwise this leads to potentially pathological situations like the one Lance mentioned.

main.py
```diff
@@ -7,6 +7,6 @@
 
     # Set "use_ipython" to True to include the ipy command if IPython is installed, which supports advanced interactive
     # debugging of your application via introspection on self.
-    app = cmd2.Cmd(use_ipython=True)
+    app = cmd2.Cmd(use_ipython=True, persistent_history_file='cmd2_history.txt')
```
tleonhardt: FYI, there is an Issue that exists which recommends moving **main.py** underneath the **cmd2** directory and renaming it.  If you are making changes to it anyways, you might want to consider knocking that one out at the same time.

cmd2/parsing.py
```diff
@@ -13,6 +13,55 @@
...
+    # to reproduce this string later
+    number_str = attr.ib(validator=attr.validators.instance_of(str))
+
+    # Tells if this argument is escaped and therefore needs to be unescaped
+    is_escaped = attr.ib(validator=attr.validators.instance_of(bool))
+
+    # Pattern used to find normal argument
+    # Digits surrounded by exactly 1 brace on a side and 1 or more braces on the opposite side
+    # Match strings like: {5}, {{{{{4}, {2}}}}}
+    macro_normal_arg_pattern = re.compile(r'(?<!\{)\{\d+\}|\{\d+\}(?!\})')
```
tleonhardt: Does "{" or "}" need to be escaped in a RegEx?  I don't think so.  This applies to the regex a few lines below as well.

cmd2/parsing.py
```diff
@@ -247,23 +296,32 @@ def __init__(
...
-            errmsg = "Aliases can not contain: {}".format(invalidchars)
+            errmsg = "Alias {}".format(errmsg)
         """
         valid = False
 
-        errmsg = 'whitespace, quotes, '
+        if not word:
+            return False, 'cannot be an empty string'
+
+        for (shortcut, expansion) in self.shortcuts:
```
tleonhardt: ``expansion`` appears unused and if that is accurate could be replaced with "_"

cmd2/parsing.py
```diff
@@ -247,23 +296,32 @@ def __init__(
...
+            if word.startswith(shortcut):
+                errmsg = 'cannot start with a shortcut: '
+                errmsg += ', '.join(shortcut for (shortcut, expansion) in self.shortcuts)
```
tleonhardt: What is this line supposed to do?  It is confusing and doesn't make sense to me.  We are already inside a for loop that loops over (shortcut, expansion) in self.shortcuts.

examples/arg_print.py
```diff
@@ -2,13 +2,12 @@
 # coding=utf-8
 """A simple example demonstrating the following:
     1) How arguments and options get parsed and passed to commands
-    2) How to change what syntax get parsed as a comment and stripped from
-       the arguments
+    2) How to change what syntax get parsed as a comment and stripped from the arguments
```
lancestarr: _gets_?

tests/conftest.py
```diff
@@ -160,11 +160,9 @@ def get_begidx():
     def get_endidx():
         return endidx
 
-    first_match = None
     with mock.patch.object(readline, 'get_line_buffer', get_line):
         with mock.patch.object(readline, 'get_begidx', get_begidx):
             with mock.patch.object(readline, 'get_endidx', get_endidx):
                 # Run the readline tab-completion function with readline mocks in place
                 first_match = app.complete(text, 0)
```
lancestarr: This temporary variable appears unnecessary

tests/test_completion.py
```diff
@@ -113,6 +113,23 @@ def test_complete_bogus_command(cmd2_app):
...
 
+def test_complete_macro(base_app, request):
+    # Create the macro
+    out = run_cmd(base_app, 'macro create fake pyscript {1}')
+    assert out == normalize("Macro 'fake' created")
+
+    # Macros do path completion
+    test_dir = os.path.dirname(request.module.__file__)
+
+    text = os.path.join(test_dir, 'script.py')
```
lancestarr: If I'm understanding this right, this unit test assumes script.py is in the test directory and verifies that if the user typed a macro name, then a complete filename, then pressed tab, the tab completion would append a space. If that's right, it seems underwhelming. Why not show that it can actually complete a filename?

You've done a ton of work here and I don't want to seem ungrateful, though. I hope I'm misinterpreting the code or there's a good reason you've done it this way.

cmd2/cmd2.py
```diff
@@ -2003,13 +2002,23 @@ def _restore_output(self, statement: Statement) -> None:
...
-        """Gets the method name associated with a given command.
+    def _cmd_func(self, command: str) -> Optional[Callable]:
+        """
+        Get the function for a command
+        :param arg: the name of the command
+        """
+        func_name = self._cmd_func_name(command)
+        if func_name:
+            return getattr(self, func_name)
+        return None
```
lancestarr: Isn't this implicit?

cmd2/cmd2.py
```diff
@@ -2003,13 +2002,23 @@ def _restore_output(self, statement: Statement) -> None:
...
+
+    def _cmd_func_name(self, command: str) -> str:
+        """Get the method name associated with a given command.
 
         :param arg: command to look up method name which implements it
```
lancestarr: Consider changing _arg_ to _command_

cmd2/cmd2.py
```diff
@@ -2588,20 +2596,18 @@ def do_help(self, arglist: List[str]) -> None:
...
+                    completer = AutoCompleter(getattr(cmd_func, 'argparser'), cmd2_app=self)
                     self.poutput(completer.format_help(arglist))
                 else:
                     # No special behavior needed, delegate to cmd base class do_help()
-                    cmd.Cmd.do_help(self, funcname[3:])
+                    super().do_help(arglist[0])
             else:
                 # This could be a help topic
-                cmd.Cmd.do_help(self, arglist[0])
+                super().do_help(arglist[0])
```
lancestarr: This repeats a line above. Could we instead:

```Python
if cmd_func and hasattr(cmd_func, 'arg_parser'):
    completer = AutoCompleter(getattr(cmd_func, 'argparser'), cmd2_app=self)
    self.poutput(completer.format_help(arglist))
else:
    # No special behavior needed
    super().do_help(arglist[0])
```

cmd2/parsing.py
lancestarr: The lines to build `errmsg` used to be before the loop, but that assignment to `errmsg` was unused if the condition inside the loop never evaluated to `True`. At my suggestion, @kmvanbrunt moved these lines into the loop. Yes, it looks a little strange at first, but the intent is to show the user the shortcut characters, since they can't be used at the beginning of an alias. I believe doing this any other way would result in needless string concatenation for a string that wouldn't be used in the happy-day scenario.

CHANGELOG.md
```diff
@@ -21,8 +21,11 @@
             if the output is a pipe or a file the escape sequences are stripped
             * Always - output methods **never** strip ANSI escape sequences, regardless of the output destination
             * Never - output methods strip all ANSI escape sequences
+    * Added ``macro`` command to create macros, which are similar to aliases, but can take arguments when called
+    * ``alias`` is now an argparse command with subcommands to create, list, and delete aliases
```
kotfu: I think we should clarify that the alias command has changed syntax. All current alias commands in startup scripts or transcripts will break with this release.

CHANGELOG.md
```diff
@@ -21,8 +21,11 @@
             if the output is a pipe or a file the escape sequences are stripped
             * Always - output methods **never** strip ANSI escape sequences, regardless of the output destination
             * Never - output methods strip all ANSI escape sequences
+    * Added ``macro`` command to create macros, which are similar to aliases, but can take arguments when called
+    * ``alias`` is now an argparse command with subcommands to create, list, and delete aliases 
 * Deprecations
-    * Deprecated the builtin ``cmd2`` suport for colors including ``Cmd.colorize()`` and ``Cmd._colorcodes``
+    * Deprecated the built-in ``cmd2`` support for colors including ``Cmd.colorize()`` and ``Cmd._colorcodes``
+    * `unalias` is no longer a command since ``alias delete`` replaced it
```
kotfu: Move the note about `unalias` from the deprecations section to the deletions section.

cmd2/cmd2.py
```diff
@@ -122,6 +122,9 @@ def __subclasshook__(cls, C):
 HELP_CATEGORY = 'help_category'
 HELP_SUMMARY = 'help_summary'
 
+# All command functions start with this
+COMMAND_PREFIX = 'do_'
```
kotfu: Consider moving to constants.py

PR #553, opened 2018-09-28
Argparse conversion

kmvanbrunt:
I converted all commands to use **argparse** with `ACArgumentParsers`. This gives a consistent look and feel for each command and usage statement. I figure this adds polish before **cmd2** 1.0 comes out.


CHANGELOG.md
```diff
@@ -22,7 +22,7 @@
             * Always - output methods **never** strip ANSI escape sequences, regardless of the output destination
             * Never - output methods strip all ANSI escape sequences
     * Added ``macro`` command to create macros, which are similar to aliases, but can take arguments when called
-    * ``alias`` is now an argparse command with subcommands to create, list, and delete aliases
```
lancestarr: Is this a merge mistake?

tleonhardt: I view this as a fix because there were duplicated lines for this on master - one here and one below under the Deletions (possibly breaking changes) section.

cmd2/cmd2.py
```diff
@@ -2585,23 +2553,73 @@ def do_macro(self, args: argparse.Namespace):
...
+        return self.basic_complete(text, line, begidx, endidx, strs_to_match)
+
+    def complete_help_subcommand(self, text: str, line: str, begidx: int, endidx: int) -> List[str]:
+        """Completes the subcommand argument of help"""
+
+        # Get all tokens through the one being completed
+        tokens, _ = self.tokens_for_completion(line, begidx, endidx)
+
+        if not tokens:
+            return []
```
tleonhardt: This method has a few uncovered lines in the various conditional cases.  If it isn't too difficult, it would be good to add unit tests to fully cover it.

cmd2/cmd2.py
```diff
@@ -2865,14 +2886,20 @@ def do_set(self, args: argparse.Namespace) -> None:
             if onchange_hook is not None:
                 onchange_hook(old=current_value, new=value)
```
tleonhardt: I don't think this line is covered.  If it isn't too difficult, it would be nice to cover.

cmd2/cmd2.py
```diff
@@ -3087,64 +3105,47 @@ def quit():
...
         # Restore command line arguments to original state
         sys.argv = orig_args
 
-    def complete_pyscript(self, text: str, line: str, begidx: int, endidx: int) -> List[str]:
-        """Enable tab-completion for pyscript command."""
-        index_dict = {1: self.path_complete}
-        return self.index_based_complete(text, line, begidx, endidx, index_dict)
-
     # Only include the do_ipy() method if IPython is available on the system
     if ipython_available:
```
tleonhardt: None of the code under here is covered with unit tests.  Since for consistency purposes we purposely don't install ipython on any of the CI test systems, we might just want to put in a #pramga to say this shouldn't count towards code coverage.

cmd2/cmd2.py
```diff
@@ -2585,23 +2553,73 @@ def do_macro(self, args: argparse.Namespace):
...
+    setattr(help_parser.add_argument('command', help="command to retrieve help for", nargs="?"),
+            ACTION_ARG_CHOICES, ('complete_help_command',))
+    setattr(help_parser.add_argument('subcommand', help="subcommand to retrieve help for",
+                                     nargs=argparse.REMAINDER),
+            ACTION_ARG_CHOICES, ('complete_help_subcommand',))
+    help_parser.add_argument('-v', '--verbose', action='store_true',
+                             help="print a list of all commands with descriptions of each")
+
+    @with_argparser(help_parser)
+    def do_help(self, args: argparse.Namespace) -> None:
```
tleonhardt: Something needs to change with this type hint and/or how it is being called.  There are a number of cases now where this is being called with a simple ``str``.  Take the following snippet from the implementation for the **alias delete** sub-command as one example:
```Python3
elif not args.name:
    self.do_help('alias delete')
```

cmd2/cmd2.py
```diff
@@ -3087,64 +3105,47 @@ def quit():
...
-            End with ``Ctrl-D`` (Unix) / ``Ctrl-Z`` (Windows), ``quit()``, '`exit()``.
-            """
+        @with_argparser(ACArgumentParser())
+        def do_ipy(self, _: argparse.Namespace) -> None:
+            """Enter an interactive IPython shell"""
             from .pyscript_bridge import PyscriptBridge
             bridge = PyscriptBridge(self)
 
+            banner = 'Entering an embedded IPython shell. Type quit() or <Ctrl>-d to exit.\n'
+            banner += 'Run Python code from external files with: run filename.py\n'
```
lancestarr: Typically we're now doing these in single statements (no more `+=`)

cmd2/cmd2.py
kmvanbrunt: I am calling it correctly since do_help() is a wrapped function whose wrapper expects a string.
Should we eliminate type hinting on the command functions?

tleonhardt: No, this appears to be a PyCharm bug.  I'm not going to worry about it.

PR #559, opened 2018-10-01
Remove self from pystate if locals_in_py is False

### LINKED ISSUE #558: py command doesn't properly respect locals_in_py=False if it was ever True
Currently the **py** command will have ``self`` accessible forevermore if the ``locals_in_py`` attribute ever gets set to True - changing it back to False will have no effect.

Either setting it to False should make ``self`` no longer accessible within **py** OR ``locals_in_py`` should not be a runtime settable parameter.

kmvanbrunt:
Fixes #558

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -2982,6 +2982,8 @@ class Cmd(cmd.Cmd):
 
             if self.locals_in_py:
                 self.pystate['self'] = self
+            elif 'self' in self.pystate:
+                del self.pystate['self']
 
             localvars = self.pystate
             from code import InteractiveConsole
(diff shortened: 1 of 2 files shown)


cmd2/cmd2.py
```diff
@@ -2982,6 +2982,8 @@ def run(filename: str):
 
             if self.locals_in_py:
                 self.pystate['self'] = self
+            elif 'self' in self.pystate:
```
tleonhardt: Looks good to me.

tests/test_cmd2.py
```diff
@@ -216,18 +216,35 @@ def test_base_shell(base_app, monkeypatch):
...
     run_cmd(base_app, 'py print(qqq)')
     out, err = capsys.readouterr()
     assert out.rstrip() == '3'
 
+    # Add a more complex statement
     run_cmd(base_app, 'py print("spaces" + " in this " + "command")')
     out, err = capsys.readouterr()
     assert out.rstrip() == 'spaces in this command'
 
+    # Set locals_in_py to True and make sure we see self
```
tleonhardt: Thanks for adding a test.

PR #568, opened 2018-10-05
Using sub-command instead of subcommand where possible to be consistent with argparse

kmvanbrunt:


cmd2/argparse_completer.py
```diff
@@ -917,6 +917,13 @@ def set_custom_message(self, custom_message: str='') -> None:
         self._custom_error_message = custom_message
     # End cmd2 customization
 
+    def add_subparsers(self, **kwargs):
```
tleonhardt: Smart way of handling a sane default

PR #569, opened 2018-10-05
No longer using stderr and self.data together to determine truthiness of a CommandResult.

### LINKED ISSUE #566: CommandResult is False for cmd2 built-ins
Since no built-in **cmd2** command fills out `self._last_result`, `CommandResult` is always **False** when running these commands in `pyscript`.

kmvanbrunt:
Either self.data is used or if that's not possible, then self.stderr is used.
This provided applications the ability to print to stderr even if an error didn't occur.

Fixes #566

### DIFF
--- a/cmd2/pyscript_bridge.py
+++ b/cmd2/pyscript_bridge.py
@@ -38,7 +38,7 @@ class CommandResult(namedtuple_with_defaults('CommandResult', ['stdout', 'stderr
 
         # If data has a __bool__ method, then call it to determine success of command
         if self.data is not None and callable(getattr(self.data, '__bool__', None)):
-            return True if self.data else False
+            return bool(self.data)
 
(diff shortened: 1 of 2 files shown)


cmd2/pyscript_bridge.py
```diff
@@ -33,9 +33,16 @@ class CommandResult(namedtuple_with_defaults('CommandResult', ['stdout', 'stderr
 
     NOTE: Named tuples are immutable.  So the contents are there for access, not for modification.
     """
-    def __bool__(self):
-        """If stderr is None and data is not None the command is considered a success"""
-        return not self.stderr and self.data is not None
+    def __bool__(self) -> bool:
```
tleonhardt: Looks good to me.  Maybe check if the Argument Processing Sphinx docs should be updated?

PR #570, opened 2018-10-05
set_window_title() will no longer write to stderr unless self._terminal_lock can be acquired

kmvanbrunt:
This is to avoid a command's stderr data being corrupted with strings used to set the window title.

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -548,8 +548,8 @@ class Cmd(cmd.Cmd):
         self.exit_code = None
 
         # This lock should be acquired before doing any asynchronous changes to the terminal to
-        # ensure the updates to the terminal don't interfere with the input being typed. It can be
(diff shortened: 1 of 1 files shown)


cmd2/cmd2.py
```diff
@@ -3525,34 +3526,48 @@ def async_alert(self, alert_msg: str, new_prompt: Optional[str] = None) -> None:
...
             return
 
-        import colorama.ansi as ansi
-        try:
-            sys.stderr.write(ansi.set_title(title))
-        except AttributeError:
-            # Debugging in Pycharm has issues with setting terminal title
-            pass
+        # Sanity check that can't fail if self.terminal_lock was acquired before calling this function
+        if self.terminal_lock.acquire(blocking=False):
```
tleonhardt: This seems like a prudent change.

PR #588, opened 2018-10-23
Fix redundant environments in tox.ini

### LINKED ISSUE #519: tox.ini has redundant pythons
**tox.ini** has too many python environments. When I run `tox` on macOS, it runs the test twice in python 3.5, one for the environment called "py35" and one for the environment called "py35-win". Not sure whether/how this might impact CI testing, but it should be investigated.

tleonhardt:
This closes #519

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -4,4 +4,4 @@ install:
 build: off
 
 test_script:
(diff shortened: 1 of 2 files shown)


tleonhardt:
This is a pretty trivial fix so I am just going to merge it in

PR #599, opened 2018-11-18
Undefined names: from colorama import Fore, Back

cclauss:
__Fore__ and __Back__ are used on line 128 but they are never defined or imported.  That means that they are _undefined names_  which have the potential to raise NameError at runtime.

[flake8](http://flake8.pycqa.org) testing of https://github.com/python-cmd2/cmd2 on Python 3.7.1

$ __flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics__
```
./examples/plumbum_colors.py:127:21: F821 undefined name 'Fore'
        color_off = Fore.RESET + Back.RESET
                    ^
./examples/plumbum_colors.py:127:34: F821 undefined name 'Back'
        color_off = Fore.RESET + Back.RESET
                                 ^
```

### DIFF
--- a/examples/plumbum_colors.py
+++ b/examples/plumbum_colors.py
@@ -31,6 +31,7 @@ import random
 import argparse
 
 import cmd2
+from colorama import Fore, Back
 from plumbum.colors import fg, bg, reset
 
 FG_COLORS = {


tleonhardt:
@cclauss Thanks for the PR.

It looks like we accidentally have a cut-and-paste error from when we copied this example from another one.  The intent is that this example should purely use [plumbum](https://plumbum.readthedocs.io/en/latest/) instead of [colorama](https://github.com/tartley/colorama).  So the ``Fore`` and ``Back`` shouldn't be used at all.  But we can fix that in another PR.

Thanks for finding and fixing this bug!

PR #600, opened 2018-11-18
Travis CI: Use flake8 to find syntax errors and undefined names

cclauss:
Currently __app__ and __self__ are _undefined names_.  Is there a way to use __global__ or __import__ or __# noqa__ to resolve this?

__E901,E999,F821,F822,F823__ are the "_showstopper_" [flake8](http://flake8.pycqa.org) issues that can halt the runtime with a SyntaxError, NameError, etc. Most other flake8 issues are merely "style violations" -- useful for readability but they do not effect runtime safety.
* F821: undefined name `name`
* F822: undefined name `name` in `__all__`
* F823: local variable name referenced before assignment
* E901: SyntaxError or IndentationError
* E999: SyntaxError -- failed to compile a file into an Abstract Syntax Tree

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,8 +46,8 @@ before_script:
   # stop the build if there are Python syntax errors or undefined names
   # exit-zero treats all errors as warnings.  The GitHub editor is 127 chars wide
   if [[ $TOXENV == py37 ]]; then
-    flake8 . --count --exclude=./examples,./tests --select=E901,E999,F821,F822,F823 --show-source --statistics ;
-    flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics ;
+    flake8 . --count --exclude=./.,./examples,./tests --select=E901,E999,F821,F822,F823 --show-source --statistics ;
+    flake8 . --count --exclude=./. --exit-zero --max-complexity=10 --max-line-length=127 --statistics ;
   fi
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ before_script:
   # stop the build if there are Python syntax errors or undefined names
   # exit-zero treats all errors as warnings.  The GitHub editor is 127 chars wide
   if [[ $TOXENV == py37 ]]; then
-    flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics ;
+    flake8 . --count --exclude=./examples,./tests --select=E901,E999,F821,F822,F823 --show-source --statistics ;
     flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics ;
   fi
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,8 +46,8 @@ before_script:
   # stop the build if there are Python syntax errors or undefined names
   # exit-zero treats all errors as warnings.  The GitHub editor is 127 chars wide
   if [[ $TOXENV == py37 ]]; then
-    flake8 . --count --exclude=./examples,./tests --select=E901,E999,F821,F822,F823 --show-source --statistics ;
-    flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics ;
+    flake8 . --count --exclude=./.*,./examples,./tests --select=E901,E999,F821,F822,F823 --show-source --statistics ;
+    flake8 . --count --exclude=./.* --exit-zero --max-complexity=10 --max-line-length=127 --statistics ;
   fi
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,8 +46,8 @@ before_script:
   # stop the build if there are Python syntax errors or undefined names
   # exit-zero treats all errors as warnings.  The GitHub editor is 127 chars wide
   if [[ $TOXENV == py37 ]]; then
-    flake8 . --count --exclude=./.,./examples,./tests --select=E901,E999,F821,F822,F823 --show-source --statistics ;
-    flake8 . --count --exclude=./. --exit-zero --max-complexity=10 --max-line-length=127 --statistics ;
+    flake8 . --count --exclude=./examples,./tests --select=E901,E999,F821,F822,F823 --show-source --statistics ;
+    flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics ;
   fi
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ matrix:
 #            - BREW_INSTALL=python3
 
 install:
-  - pip install tox
+  - pip install flake8 tox
 #  - |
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -33,15 +33,22 @@ matrix:
...
 #        brew install "$BREW_INSTALL"
 #      fi
 #    fi
-#    pip install tox
+
+before_script:
+  # stop the build if there are Python syntax errors or undefined names
+  # exit-zero treats all errors as warnings.  The GitHub editor is 127 chars wide
+  if [[ $TOXENV == py37 ]]; then
+    flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics ;
```
tleonhardt: Adding a check for these "*showstopper*" issues is a nice improvement.  However, I would prefer if we only check the actual deliverable ``cmd2`` source code in the **cmd2** directory.  

As you pointed out, if we also apply these checks to code in the **examples** or **tests** directories, we will run into issues with ``app`` and ``self`` being undefined names.  This is a side-effect of the somewhat unconventional way in which ``cmd2`` supports Python scripting within ``cmd2`` applications.  These Python scripts which use ``app`` and ``self`` are ONLY intended to ever be run inside ``cmd2``'s scripting engine and not within a generic Python interpreter.

@cclauss What do you think?

@kmvanbrunt @kotfu @anselor Do any of you have any bright ideas how to make a small modification to the ``cmd2`` code so that ``flake8`` doesn't warn about ``app`` and/or ``self`` being undefined names in our example scripts and tests?

.travis.yml
```diff
@@ -33,15 +33,22 @@ matrix:
...
+    flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics ;
```
tleonhardt: This one I'm not in favor of, at least not yet because it is too strict.  But I like the idea of creating a new Issue to clean up some of the ``flake8`` warnings.

.travis.yml
```diff
@@ -33,15 +33,22 @@ matrix:
...
 #        brew install "$BREW_INSTALL"
 #      fi
 #    fi
-#    pip install tox
+
+before_script:
+  # stop the build if there are Python syntax errors or undefined names
+  # exit-zero treats all errors as warnings.  The GitHub editor is 127 chars wide
+  if [[ $TOXENV == py37 ]]; then
+    flake8 . --count --exclude=./examples,./tests --select=E901,E999,F821,F822,F823 --show-source --statistics ;
```
tleonhardt: If we go the route of excluding specified directories instead of just including the **cmd2** directory, then for performance reasons when running locally, it is very important that we include "**./.***" so it doesn't try to parse directories such as:
* .cache
* .git
* .idea
* .pytest_cache
* .tox
* .vscode

I'm fine with this blacklisting approach, but the performance on my dev machine without excluding "**./.***" is orders of magnitude worse than running flake8 without any exclusions or running flake8 on  the cmd2 directory.

.travis.yml
tleonhardt: Ok, as long as it is only producing warnings, I'm OK with this raising visibility for developers.  Thanks for pointing out that this won't fail the build.

.travis.yml
```diff
@@ -33,15 +33,22 @@ matrix:
...
 #        brew install "$BREW_INSTALL"
 #      fi
 #    fi
-#    pip install tox
+
+before_script:
+  # stop the build if there are Python syntax errors or undefined names
+  # exit-zero treats all errors as warnings.  The GitHub editor is 127 chars wide
+  if [[ $TOXENV == py37 ]]; then
+    flake8 . --count --exclude=./.,./examples,./tests --select=E901,E999,F821,F822,F823 --show-source --statistics ;
```
tleonhardt: I'm not sure how the exclude parser for ``flake8`` works, but the way you have it now actually excludes **EVERYTHING**.  The pattern needs to be written with a trailing asterisk "./.*" like so:
```sh
flake8 . --count --exclude=./.*,./examples,./tests --select=E901,E999,F821,F822,F823 --show-source --statistics ;
```

Same comment applies to the line below.

cclauss: I agree.

Undoing the last change.

tleonhardt: @cclauss The last change was a good change, it just needed an asterisk after the dot in the first part of the excludes.

Without that exclude it runs for an extremely long time on my system - many minutes.

.travis.yml
tleonhardt: On my Mac, the timing is as follows:
* 100 secs for **time flake8 --exclude=./examples,./tests**
*     2 secs for **time flake8 --exclude=./.*./examples,./tests**
*     2 secs for **time flake8 .**

.travis.yml
cclauss: ???

$ __time flake8 .__
```
real	0m3.379s
user	0m7.075s
sys	0m0.268s
```
$ __time flake8 . --exclude=./.*__
```
real	0m3.161s
user	0m7.070s
sys	0m0.224s
```
What is your machine?

.travis.yml
cclauss: Put a . in the first one.

.travis.yml
tleonhardt: Try a **time flake8 --exclude=./examples,./tests**.

The timing is fine if you don't have any excludes - I'm assuming it has sane default exclusions if you don't specify --exclude.  But as soon as you specify --exclude, you need to do a thorough job of it - it looks like those sane defaults go away.

tleonhardt:
Full data on timing ...
```sh
$ flake8 --version
3.5.0 (mccabe: 0.6.1, pycodestyle: 2.4.0, pyflakes: 1.6.0) CPython 3.7.0 on Darwin

18:35 $ time flake8 . --count --exclude=./.*,./examples,./tests --select=E901,E999,F821,F822,F823
real	0m1.273s
user	0m2.227s
sys	0m0.188s

$ time flake8 . --count --select=E901,E999,F821,F822,F823
real	0m1.995s
user	0m5.413s
sys	0m0.241s

18:35 $ time flake8 . --count --exclude=./examples,./tests --select=E901,E999,F821,F822,F823
real	1m35.387s
user	9m51.663s
sys	0m3.212s
```

.travis.yml
```diff
@@ -33,15 +33,22 @@ matrix:
...
 #        brew install "$BREW_INSTALL"
 #      fi
 #    fi
-#    pip install tox
+
+before_script:
+  # stop the build if there are Python syntax errors or undefined names
+  # exit-zero treats all errors as warnings.  The GitHub editor is 127 chars wide
+  if [[ $TOXENV == py37 ]]; then
+    flake8 . --count --exclude=./.*,./examples,./tests --select=E901,E999,F821,F822,F823 --show-source --statistics ;
```
tleonhardt: Thanks for working with me to get this working well.

Sorry for all of the back and forth.

Issue #545, opened 2018-09-26
What should show in cmd2 history?


kmvanbrunt:
@kotfu @tleonhardt 
This question pertains to cmd2's history and not readline's.

Currently we put what the user typed in history. (`statement.raw`) Is there any reason why can't put what was run instead? (`statement.command_and_args`)

That would make it clearer when things like shortcuts and aliases are entered. You will have a record of what actually occurred.

kotfu:
I had this same thought last time I fixed a transcript bug. It's confusing now because you can't see what actually got run when you type an alias. If we switched it it would be confusing because when you   use the history command it won't show what you typed.

What do you think about a developer setting (not a user available setting) which would allow the developer to choose one of these behaviors?

tleonhardt:
As long as it doesn't present too big of a maintenance headache, I like the idea of a developer setting which would allow ability to choose one of these behaviors; because as you discussed there is a potential confusion either way.

kmvanbrunt:
What would the default be?

kotfu:
I think the default should be to put what the user typed into the history. That's what cmd.cmd does, and that's what cmd2 has done thus far. 

We need to proceed carefully here. A quick look at the code indicates that we could easily break transcripts by messing with this. `statement.command_and_args` excludes any output redirection clauses the user typed. Our current unit tests don't cover these types of scenarios very well, so if we proceed to make a change, I suggest we create more robust history and transcript tests, and then once those are committed and passing with no code changes to `cmd2.Cmd`, then we can start to implement this new setting.

A completely different alternative might be to add a new command line switch to the history command, which outputs the commands with shortcuts and aliases expanded. This would make it explicit, available to the user, and wouldn't break any existing functionality.

kmvanbrunt:
What about adding a member to `HistoryItem` that's the resolved command? The history output could somehow print both if the `raw` is different than `command_and_args`.

kmvanbrunt:
Even if we add flags to the `history` command, I believe we still need to store both the `raw` and `command_and_args` values in `HistoryItem`. That way we only have to resolve the command once.

Plus, a user's parsing hooks might be dependent on state and resolve a command differently if run again. Therefore we need to store what was resolved at the time the user ran it.

tleonhardt:
Excellent conversation and suggestion!  So here is my summary and proposal of what we want to do moving forward:

* By default the **history** command will print what the user typed, just like it does now
[...]

PR #604, opened 2018-11-30
Made it so default_to_shell results in do_shell being called so that output can be captured

kmvanbrunt:
This partially addresses #545

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -1998,25 +1998,16 @@ class Cmd(cmd.Cmd):
         if statement.command in self.macros:
             stop = self._run_macro(statement)
         else:
-            command = statement.command
-            func = self.cmd_func(command)
-            func_arg = statement.args
-
-            if not func and self.default_to_shell:
-                command = 'shell'
-                func = self.cmd_func(command)
-                func_arg = statement.command_and_args
-
+            func = self.cmd_func(statement.command)
             if func:
-                # Check if this command should be stored in the history
-                if command not in self.exclude_from_history:
+                # Since we have a valid command store it in the history
+                if statement.command not in self.exclude_from_history:
                     self.history.append(statement.raw)
 
-                stop = func(func_arg)
+                stop = func(statement)
 
             else:
-                self.default(statement)
-                stop = False
+                stop = self.default(statement)
 
         if stop is None:
             stop = False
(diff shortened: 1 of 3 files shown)


cmd2/cmd2.py
```diff
@@ -632,7 +632,7 @@ def perror(self, err: Union[str, Exception], traceback_war: bool=True, err_color
         err_msg = err_color + err_msg + Fore.RESET
         self.decolorized_write(sys.stderr, err_msg)
 
-        if traceback_war:
+        if traceback_war and not self.debug:
```
tleonhardt: Nice little fix.  It didn't make sense to print this if ``self.debug`` is ``True``.

cmd2/cmd2.py
```diff
@@ -1893,6 +1893,8 @@ def _redirect_output(self, statement: Statement) -> None:
             try:
                 self.pipe_proc = subprocess.Popen(statement.pipe_to, stdin=subproc_stdin)
             except Exception as ex:
+                self.perror('Not piping because - {}'.format(ex), traceback_war=False)
```
tleonhardt: Yeah, this seems like a good idea to inform the user - good call.

cmd2/cmd2.py
```diff
@@ -1998,18 +1998,29 @@ def onecmd(self, statement: Union[Statement, str]) -> bool:
         if statement.command in self.macros:
             stop = self._run_macro(statement)
         else:
-            func = self.cmd_func(statement.command)
+            command = statement.command
+            func = self.cmd_func(command)
+            func_arg = statement.args
+
+            if not func and self.default_to_shell:
```
tleonhardt: I very much like changing the logic of how ``default_to_shell`` behavior works so that it directly uses the ``do_shell`` method.  However, I'm not sure I like moving that logic out of the ``do_default`` method.  

The benefit of having it within the ``do_default`` method is that it allows end users to override ``do_default`` however they choose and get reasonable expected behavior.

cmd2/cmd2.py
```diff
@@ -2057,18 +2068,11 @@ def _run_macro(self, statement: Statement) -> bool:
...
         """
-        if self.default_to_shell:
-            result = os.system(statement.command_and_args)
-            # If os.system() succeeded, then don't print warning about unknown command
-            if not result:
-                return
-
-        # Print out a message stating this is an unknown command
-        self.poutput('*** Unknown syntax: {}\n'.format(statement.command_and_args))
+        self.poutput('*** {} is not a recognized command, alias, or macro\n'.format(statement.command))
```
tleonhardt: Nice improvement, we should have always done this.

cmd2/cmd2.py
```diff
@@ -2803,7 +2807,7 @@ def show(self, args: argparse.Namespace, parameter: str='') -> None:
         :param args: argparse parsed arguments from the set command
         :param parameter: optional search parameter
         """
-        param = parameter.strip().lower()
+        param = utils.norm_fold(parameter.strip())
```
tleonhardt: Good call.  We should have been doing proper unicode normalization and case folding

cmd2/cmd2.py
```diff
@@ -3156,15 +3160,19 @@ def load_ipy(app):
     history_parser_group.add_argument('-e', '--edit', action='store_true',
                                       help='edit and then run selected history items')
     history_parser_group.add_argument('-s', '--script', action='store_true', help='output commands in script format')
-    history_parser_group.add_argument('-o', '--output-file', metavar='FILE', help='output commands to a script file')
-    history_parser_group.add_argument('-t', '--transcript', help='output commands and results to a transcript file')
+    setattr(history_parser_group.add_argument('-o', '--output-file', metavar='FILE',
```
tleonhardt: Nice tab-completion behavior improvements

Issue #548, opened 2018-09-27
Consider converting cmd2.Cmd class attributes to instance attributes


tleonhardt:
Many of the current class attributes within the ``cmd2.Cmd`` class should actually be instance attributes instead.  

Consider these on a case-by-case basis and convert them accordingly along with any affected unit tests and examples.

Having some things as class attributes when they should be instance attributes recently caused a problem with unit tests not being isolated from each other because the class itself was getting modified in memory in-between test fixture invocations which were creating instances of the class.  So we should be very careful to make sure nothing is a class attribute unless we really intend it to be.

tleonhardt:
So I think we need to categorize ``cmd2.Cmd`` attributes into the following categories:
* class attributes (all of these should probably be immutable)
    * shared by all members of the class - constants set in class code and never changed
* immutable instance attributes
    * set within ``cmd2.Cmd.__init__`` during instance initialization and never changed
* mutable instance attributes
    * set within ``cmd2.Cmd.__init__``, but also updatable by other methods

The trick will be deciding what makes sense for which attributes.

Also, should we try to enforce immutability using something like ``attrs``?

PR #605, opened 2018-12-04
Moved some class specific code out of __init__

kmvanbrunt:
This partially addresses #548

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -394,10 +394,6 @@ class Cmd(cmd.Cmd):
         # Call super class constructor
         super().__init__(completekey=completekey, stdin=stdin, stdout=stdout)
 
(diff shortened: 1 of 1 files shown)


cmd2/cmd2.py
```diff
@@ -2584,6 +2580,10 @@ def complete_help_subcommand(self, text: str, line: str, begidx: int, endidx: in
     help_parser.add_argument('-v', '--verbose', action='store_true',
                              help="print a list of all commands with descriptions of each")
 
+    # Get rid of cmd's complete_help() functions so AutoCompleter will complete the help command
+    if getattr(cmd.Cmd, 'complete_help', None) is not None:
```
tleonhardt: Yup, this looks like a better place for this

PR #607, opened 2018-12-06
Fix flake8 issues

### LINKED ISSUE #601: Fix flake8 issues in the code
https://travis-ci.org/python-cmd2/cmd2/jobs/456554031#L1339 provides a list of > 700 flake8 issues in the codebase.  Some are important and others less so but PRs that fix these issues would be of interest.

tleonhardt:
This commit contains a very large number of trivial changes in order to fix flake8 errors and warnings.  Predominantly these are whitespace changes.

Additionally, the build for Python 3.7 on TravisCI has been tweaked to fail if there are any flake8 errors using the following commandline:
* flake8 . --count --ignore=E252 --max-complexity=31 --max-line-length=127 --show-source --statistics
NOTE: In the future the max cyclomatic complexity should be lowered, but some improvements need to be made first.

One flake8 error is being ignored entirely:
* E252 missing whitespace around parameter equals
    * ignored because it doesn't correctly deal with default argument values after a type hint

A few flake8 errors are being selectively ignored in certain files:
* C901 fuction is too complex
    * ignored in argparse_completer.py because the complex code is an override of argparse complexity
* E302 expected 2 blank lines after ...
    * ignored in all unit test files for convenience
* F401 module imported but unused
    * ignored in cmd2/__init__.py because imports are for convenience of cmd2 developers and backwards compatibility
* F821 undefined name
    * ignored in cmd2 script files which are intended to run only within cmd2 applications via pyscript where "app" and "cmd" are defined

This closes #601


tleonhardt:
The masking of F821's is only done for test and example files which are not deliverable as part of the ``cmd2`` package itself and thus will not affect code being exercised by end users.

PR #609, opened 2018-12-06
Removed ERROR: text from perror output

kmvanbrunt:


cmd2/cmd2.py
```diff
@@ -624,7 +624,7 @@ def perror(self, err: Union[str, Exception], traceback_war: bool=True, err_color
         if isinstance(err, Exception):
             err_msg = "EXCEPTION of type '{}' occurred with message: '{}'\n".format(type(err).__name__, err)
         else:
-            err_msg = "ERROR: {}\n".format(err)
+            err_msg = "{}\n".format(err)
```
tleonhardt: Make sure you add something to the CHANGELOG about a potentially breaking change since this could possibly break unit tests for applications which use ``cmd2``.

PR #610, opened 2018-12-07
Fix pytest deprecation warnings

### LINKED ISSUE #606: Fix pytest warnings
``pytest`` 4.x has warnings that should be addressed, including **mock** being deprecated and replaced by **mocker**.

kotfu:
Fix #606.

### DIFF
--- a/tests/test_bashcompletion.py
+++ b/tests/test_bashcompletion.py
@@ -124,13 +124,13 @@ def my_fdopen(fd, mode, *args):
 
 # noinspection PyShadowingNames
 @pytest.mark.skipif(skip_no_argcomplete or skip_windows, reason=skip_reason)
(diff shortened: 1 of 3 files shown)


tests/test_bashcompletion.py
```diff
@@ -124,13 +124,13 @@ def my_fdopen(fd, mode, *args):
 
 # noinspection PyShadowingNames
 @pytest.mark.skipif(skip_no_argcomplete or skip_windows, reason=skip_reason)
-def test_invalid_ifs(parser1, mock):
+def test_invalid_ifs(parser1, mocker):
```
tleonhardt: Thanks for fixing the ``pytest`` warnings.  I also have no idea why they show up some of the time and not others!

PR #612, opened 2018-12-14
Fixed bug in cmd2.Cmd.select() when user enters 0 or negative number

### LINKED ISSUE #611: The select() method needs to check for a user response of 0.
The select() method presents a numbered list, starting with "1", for the user to choose from. When a 0 is entered, the last item in the list is automatically selected. Depending on the list size, some negative numbers are also accepted. I would suggest changing the code as follows to flag anything less then 1 as invalid:

```Python3
choice = int(response)
if choice < 1:
     self.poutput("{!r} isn't a valid choice. Pick a number between 1 and {}:\n".format(response, len(fulloptions)))
     continue
result = fulloptions[choice - 1][0]
```

tleonhardt:
This closes #611

### DIFF
--- a/tests/test_cmd2.py
+++ b/tests/test_cmd2.py
@@ -1344,7 +1344,7 @@ def test_select_options(select_app):
     # And verify the expected output to stdout
     assert out == expected
 
-def test_select_invalid_option(select_app):
+def test_select_invalid_option_too_big(select_app):
     # Mock out the input call so we don't actually wait for a user's response on stdin
     m = mock.MagicMock(name='input')
     # If side_effect is an iterable then each call to the mock will return the next value from the iterable.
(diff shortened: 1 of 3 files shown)


tests/test_cmd2.py
```diff
@@ -1368,6 +1368,30 @@ def test_select_invalid_option(select_app):
     # And verify the expected output to stdout
     assert out == expected
 
+def test_select_invalid_option_too_small(select_app):
+    # Mock out the input call so we don't actually wait for a user's response on stdin
+    m = mock.MagicMock(name='input')
+    # If side_effect is an iterable then each call to the mock will return the next value from the iterable.
+    m.side_effect = ['0', '1']  # First pass and invalid selection, then pass a valid one
```
kmvanbrunt: Change to "First pass an"
Same with line 1351

tleonhardt: Done

PR #619, opened 2019-01-24
Fixed buffering issue when echoing StdSim data

kmvanbrunt:
Echoing string data directly to inner_stream instead of inner_stream.buffer to make sure line buffering is respected

### DIFF
--- a/cmd2/utils.py
+++ b/cmd2/utils.py
@@ -281,10 +281,7 @@ class StdSim(object):
                 raise TypeError('a bytes-like object is required, not {}'.format(type(b)))
             self.byte_buf += b
             if self.echo:
-                if hasattr(self.inner_stream, 'buffer'):
-                    self.inner_stream.buffer.write(b)
-                else:
-                    self.inner_stream.write(b.decode(encoding=self.encoding, errors=self.errors))
+                self.inner_stream.buffer.write(b)
 
     def __init__(self, inner_stream, echo: bool = False,
                  encoding: str='utf-8', errors: str='replace') -> None:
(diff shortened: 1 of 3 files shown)


cmd2/utils.py
```diff
@@ -281,10 +281,7 @@ def write(self, b: bytes) -> None:
                 raise TypeError('a bytes-like object is required, not {}'.format(type(b)))
             self.byte_buf += b
             if self.echo:
-                if hasattr(self.inner_stream, 'buffer'):
-                    self.inner_stream.buffer.write(b)
-                else:
-                    self.inner_stream.write(b.decode(encoding=self.encoding, errors=self.errors))
+                self.inner_stream.buffer.write(b)
```
tleonhardt: I don't think this is a 100% safe change.  I vaguely recall when manually testing this class in a variety of ways that there are some small minority of things that could be used as an inner_stream which do **NOT** have a buffer attribute. 

Thus, I would recommend leaving this code block as is.

cmd2/utils.py
```diff
@@ -297,15 +294,17 @@ def __init__(self, inner_stream, echo: bool = False,
...
         self.encoding = encoding
         self.errors = errors
 
     def write(self, s: str) -> None:
         """Add str to internal bytes buffer and if echo is True, echo contents to inner stream"""
         if not isinstance(s, str):
             raise TypeError('write() argument must be str, not {}'.format(type(s)))
-        b = s.encode(encoding=self.encoding, errors=self.errors)
-        self.buffer.write(b)
+        self.buffer.byte_buf += s.encode(encoding=self.encoding, errors=self.errors)
```
tleonhardt: This part looks OK I think

cmd2/utils.py
tleonhardt: Actually now that you refactored the part that was calling this, I think this is ok

PR #621, opened 2019-02-06
Made quit() and exit() available to pyscripts

### LINKED ISSUE #620: Add quit() and exit() to pyscript environment
`quit()` and `exit()` only exist in the interactive Python console. If you try to run these functions within a pyscript, they close the whole console application.

These functions should be made available to **pyscript** environments so scripts can return back to the console.

kmvanbrunt:
This fixes #620

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -2959,165 +2959,160 @@ class Cmd(cmd.Cmd):
             return False
         self._in_py = True
 
-        # noinspection PyBroadException
-        try:
-            # Support the run command even if called prior to invoking an interactive interpreter
-            def run(filename: str):
-                """Run a Python script file in the interactive console.
+        # Support the run command even if called prior to invoking an interactive interpreter
+        def py_run(filename: str):
+            """Run a Python script file in the interactive console.
+            :param filename: filename of *.py script file to run
+            """
+            expanded_filename = os.path.expanduser(filename)
 
-                :param filename: filename of *.py script file to run
-                """
-                expanded_filename = os.path.expanduser(filename)
-
-                # cmd_echo defaults to False for scripts. The user can always toggle this value in their script.
-                bridge.cmd_echo = False
-
-                try:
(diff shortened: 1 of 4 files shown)


cmd2/cmd2.py
```diff
@@ -2979,9 +2979,17 @@ def run(filename: str):
...
+            def quit():
+                """Function callable from the interactive Python console to exit that environment"""
+                raise EmbeddedConsoleExit
+
+            # Set up Python environment
             bridge = PyscriptBridge(self)
-            self.pystate['run'] = run
             self.pystate[self.pyscript_name] = bridge
+            self.pystate['run'] = run
+            self.pystate['quit'] = quit
```
tleonhardt: Looks good.  Would it be reasonable to add a couple new unit tests to cover the added functionality?

cmd2/cmd2.py
```diff
@@ -3113,7 +3117,7 @@ def quit():
             pass
         finally:
             self._in_py = False
-        return self._should_quit
+            return self._should_quit
```
tleonhardt: This interesting whitespace change seems innocuous.  But is there a good reason for it?

cmd2/cmd2.py
tleonhardt: If it isn't easy to add a unit test, perhaps add an example?

cmd2/cmd2.py
```diff
@@ -2959,160 +2959,159 @@ def do_py(self, args: argparse.Namespace) -> bool:
...
-                                    else:
-                                        sys.modules['readline'] = saved_readline
+                            if 'gnureadline' in sys.modules:
+                                # Restore what the readline module pointed to
+                                if saved_readline is None:
+                                    del(sys.modules['readline'])
+                                else:
+                                    sys.modules['readline'] = saved_readline
 
-        except Exception:
```
tleonhardt: We need something like this to handle a KeyboardInterrupt exception (Ctrl-C)

PR #623, opened 2019-02-06
Fixed issue where calling exit() or quit() from a pyscript would close the whole console

### LINKED ISSUE #620: Add quit() and exit() to pyscript environment

kmvanbrunt:
This fixes #620

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -2959,160 +2959,159 @@ class Cmd(cmd.Cmd):
             return False
         self._in_py = True
 
-        # noinspection PyBroadException
-        try:
-            # Support the run command even if called prior to invoking an interactive interpreter
-            def run(filename: str):
-                """Run a Python script file in the interactive console.
+        # Support the run command even if called prior to invoking an interactive interpreter
+        def py_run(filename: str):
+            """Run a Python script file in the interactive console.
+            :param filename: filename of *.py script file to run
+            """
+            expanded_filename = os.path.expanduser(filename)
 
-                :param filename: filename of *.py script file to run
-                """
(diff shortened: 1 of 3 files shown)


cmd2/cmd2.py
```diff
@@ -2959,160 +2959,159 @@ def do_py(self, args: argparse.Namespace) -> bool:
...
-                                    else:
-                                        sys.modules['readline'] = saved_readline
+                            if 'gnureadline' in sys.modules:
+                                # Restore what the readline module pointed to
+                                if saved_readline is None:
+                                    del(sys.modules['readline'])
+                                else:
+                                    sys.modules['readline'] = saved_readline
 
-        except Exception:
```
tleonhardt: I think we need something like this to catch a KeyboardInterrupt exception (Ctrl-C) and let that gracefully exit the interactive Python session and drop us back to the console

kmvanbrunt: I allow the interactive Python session to handle their own exceptions. I did however add KeyboardInterrupt suppression to py and pyscript for when the interactive console isn't running.

cmd2/cmd2.py
```diff
@@ -3109,10 +3115,12 @@ def quit():
                                     else:
                                         sys.modules['readline'] = saved_readline
 
-        except Exception:
+        except KeyboardInterrupt:
```
tleonhardt: Yeah I guess this is good.

Issue #613, opened 2018-12-21
Redirecting pyscript with app() commands closes cmd2


kmvanbrunt:
I have a pyscript with 1 line:
`app('help')`

If the script is run like: `pyscript script.py > out.txt', then cmd2 closes.

If a second line is added as shown below and the script run as before, then an exception prints before cmd2 exits.
`app('help')`
`app('help')`

Exception that prints:
> Traceback (most recent call last):
>   File "/home/user/PycharmProjects/cmd2/cmd2/cmd2.py", line 1716, in onecmd_plus_hooks
>     stop = self.onecmd(statement)
>   File "/home/user/PycharmProjects/cmd2/cmd2/cmd2.py", line 2003, in onecmd
>     stop = func(statement)
>   File "/home/user/PycharmProjects/cmd2/cmd2/cmd2.py", line 259, in cmd_wrapper
>     return func(instance, args)
>   File "/home/user/PycharmProjects/cmd2/cmd2/cmd2.py", line 2591, in do_help
>     self._help_menu(args.verbose)
>   File "/home/user/PycharmProjects/cmd2/cmd2/cmd2.py", line 2640, in _help_menu
>     self.poutput("{}\n".format(str(self.doc_leader)))
>   File "/home/user/PycharmProjects/cmd2/cmd2/cmd2.py", line 601, in poutput
>     self.decolorized_write(self.stdout, msg_str)
>   File "/home/user/PycharmProjects/cmd2/cmd2/cmd2.py", line 579, in decolorized_write
>     (self.colors.lower() == constants.COLORS_TERMINAL.lower() and not fileobj.isatty()):
> ValueError: I/O operation on closed file
> EXCEPTION of type 'ValueError' occurred with message: 'I/O operation on closed file'

tleonhardt:
I did a little regression testing and this bug was not present in the 0.8.9 release and is present in the 0.9.0 release.  So I believe it was introduced when [pyscript_bridge.py](https://github.com/python-cmd2/cmd2/blob/master/cmd2/pyscript_bridge.py) was created.

This bug isn't specific to the ``pyscript`` command, it is reproducible whenever a ``cmd2`` command is called with ``py`` and redirected; thus the following also quits the ``cmd2`` app:
```sh
(Cmd) py app('help') > out.txt
(py37) ✔ ~/src/cmd2/examples [master|…2]
```

However, running a non-cmd2 command with ``py`` does NOT exit the ``cmd2`` app:
```sh
[...]

PR #630, opened 2019-03-01
Pyscript overhaul

### LINKED ISSUE #586: py app.command() doesn't work if command has sub-commands
`py app.help()` produces output.
`py app.alias()` does **not** produce output.

The difference between `help()` and `alias()` is that `alias()` has sub-commands. This is why
`py app.alias.create()` produces output.

kmvanbrunt:
Removed ability to call commands in ``pyscript`` as if they were functions (e.g ``app.help()``) in favor of only supporting one ``pyscript`` interface. This simplifies future maintenance.

This closes #586 since the feature will no longer exist.


CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 * Deprecations
     * Deprecated support for bash completion since this feature had slow performance. Also it relied on
     ``AutoCompleter`` which has since developed a dependency on ``cmd2`` methods. 
+    * Removed ability to call commands in ``pyscript`` as if they were functions (e.g ``app.help()``) in favor
```
tleonhardt: This should be under the **Potentially breaking changes** section since it could break a user's Python scripts.

cmd2/argparse_completer.py
```diff
@@ -988,7 +988,7 @@ def __init__(self, *args, **kwargs) -> None:
         self._custom_error_message = ''
 
     # Begin cmd2 customization
-    def set_custom_message(self, custom_message: str='') -> None:
+    def set_custom_message(self, custom_message: str = '') -> None:
```
tleonhardt: I think this aspect of PEP8 predates optional type hinting and shouldn't be applied.  That being said, PyCharm and/or pylint complain, so it is probably good to just do it this way for maintainability.

cmd2/pyscript_bridge.py
```diff
@@ -7,12 +7,10 @@
 Released under MIT license, see LICENSE file
 """
 
-import argparse
 import sys
-from typing import List, Optional
+from typing import Optional
```
tleonhardt: There are major changes to the way the **pyscript** command works beyond what we can catch in our unit tests.  Have you performed sufficient manual testing with existing scripts to verify correct behavior?

tests/pyscript/bar1.py
```diff
@@ -1,3 +0,0 @@
-# flake8: noqa F821
```
tleonhardt: Is deleting the majority of our **pyscript** unit tests the right thing to do?  Or would it be better to convert them to using the other syntax given that is the only supported one now?

tleonhardt:
@kmvanbrunt Would it be reasonable to address #613 in this PR too?

cmd2/argparse_completer.py
kmvanbrunt: See https://www.python.org/dev/peps/pep-0526/
Look for: `def __init__(self, damage: int, captain: str = None):`

tests/pyscript/bar1.py
kmvanbrunt: The deleted tests only exercised the `ArgparseFunctor` stuff.

The remaining interface to `pyscript` is simple because it does no translation before sending the command to our parser. It is pretty well covered by existing unit tests.

cmd2/pyscript_bridge.py
kmvanbrunt: The existing scripts were not affected since they did not use `ArgparseFunctor` or `PyscriptBridge.__getattr__()`. But yes, I did test them.

CHANGELOG.md
kmvanbrunt: True, but it's also a deprecation which are always "Potentially breaking changes".
Still want it moved?

tleonhardt: I think in the past we have separated the categories as follows;
* Deprecations
    * This is where we recommend users stop using a feature because we plan to delete it in a future release
* Deletions (potentially breaking changes)
    * This is where a feature was deleted in the release and thus things which used it will break

cmd2/argparse_completer.py
tleonhardt: Ok, I guess its official then.  I shall attempt to assimilate my thinking to the Borg.

cmd2/pyscript_bridge.py
tleonhardt: Ok, sounds good enough.

tests/pyscript/bar1.py
tleonhardt: Maybe take a quick look at code coverage within ``do_pyscript`` and ``pyscript_bridge``.  But I think your current approach looks OK overall.

CHANGELOG.md
kmvanbrunt: OK, I'll fix that

PR #634, opened 2019-03-04
Fixed exception caused by using preserve_quotes argument of with_argument_list decorator

### LINKED ISSUE #633: with_argument_list breaks when using preserve_quotes argument
The following function causes this exception
`TypeError: with_argument_list() missing 1 required positional argument: 'func'`

```Python
@with_argument_list(preserve_quotes=True)
    def do_blah(self, args):
        print(args)
```

kmvanbrunt:
Fixes #633

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -175,25 +175,26 @@ def with_category(category: str) -> Callable:
     return cat_decorator
 
 
-def with_argument_list(func: Callable[[Statement], Optional[bool]],
-                       preserve_quotes: bool = False) -> Callable[[List], Optional[bool]]:
+def with_argument_list(preserve_quotes: bool = False) -> Callable[[List], Optional[bool]]:
     """A decorator to alter the arguments passed to a do_* cmd2 method. Default passes a string of whatever the user
     typed. With this decorator, the decorated method will receive a list of arguments parsed from user input using
     shlex.split().
 
-    :param func: do_* method this decorator is wrapping
     :param preserve_quotes: if True, then argument quotes will not be stripped
     :return: function that gets passed a list of argument strings
     """
     import functools
 
-    @functools.wraps(func)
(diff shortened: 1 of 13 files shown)


cmd2/cmd2.py
```diff
@@ -175,25 +175,26 @@ def cat_decorator(func):
     return cat_decorator
 
 
-def with_argument_list(func: Callable[[Statement], Optional[bool]],
-                       preserve_quotes: bool = False) -> Callable[[List], Optional[bool]]:
+def with_argument_list(preserve_quotes: bool = False) -> Callable[[List], Optional[bool]]:
```
tleonhardt: Now that I think about it, it seems like True would be a much better default value for this.  What do you think.  Honestly in my memory I didn’t know this argument existed and I thought we always preserved quotes.

docs/argument_processing.rst
```diff
@@ -371,7 +371,7 @@ argument list instead of a string::
             # cmdline contains a string
             pass
 
-        @with_argument_list
+        @with_argument_list()
```
tleonhardt: I find the required parentheses without any argument slightly awkward, but not a big deal.  Just a thought - is it with having the preserve_quotes argument at all?  Are there important use cases we would miss by just always preserving quotes when this decorator is used?

docs/freefeatures.rst
```diff
@@ -217,19 +217,6 @@ of using ``pyscript`` is shown below  along with the **examples/arg_printer.py**
     arg 2: 'bar'
     arg 3: 'baz'
 
-.. note::
```
tleonhardt: I would recommend keeping this note and just updating it to show the required parentheses.

tleonhardt:
See this StackOverflow post for a possible way we could use the decorator with or without parents:  https://stackoverflow.com/questions/35572663/using-python-decorator-with-or-without-parentheses

PR #636, opened 2019-03-05
Fix bug in with_argument_list decorator

### LINKED ISSUE #633: with_argument_list breaks when using preserve_quotes argument

tleonhardt:
Fix bug when the ``with_argument_list decorator`` is called with the optional ``preserve_quotes`` argument.

This closes #633


docs/freefeatures.rst
```diff
@@ -222,12 +222,11 @@ of using ``pyscript`` is shown below  along with the **examples/arg_printer.py**
     If you want to be able to pass arguments with spaces to scripts, then we strongly recommend using one of the decorators,
```
kmvanbrunt: This note seems out of place. It says: "If you want to be able to pass arguments with spaces to scripts..."
But `lprint` is not a script.
Maybe just delete this note since the `with_argument_list` decorator is already documented elsewhere.

cmd2/cmd2.py
```diff
@@ -175,25 +175,30 @@ def cat_decorator(func):
     return cat_decorator
 
 
-def with_argument_list(func: Callable[[Statement], Optional[bool]],
-                       preserve_quotes: bool = False) -> Callable[[List], Optional[bool]]:
+def with_argument_list(*args: List[Callable], preserve_quotes: bool = False) -> Callable[[List], Optional[bool]]:
```
kmvanbrunt: I realize this works, but I'm confused. Why is args a list of Callables?

docs/freefeatures.rst
tleonhardt: Thanks for pointing that out.  I will reword it.

cmd2/cmd2.py
tleonhardt: Here I am making use of the ability to gather all positional arguments into a list, sort of like a varargs in C/C++ combined with the feature that any arguments after that are keyword-only arguments.

kmvanbrunt: Wow

tleonhardt: Here is a better and more complete explanation:  https://www.digitalocean.com/community/tutorials/how-to-use-args-and-kwargs-in-python-3

PR #637, opened 2019-03-05
Using StatementParser instead of shlex within command method wrappers

kmvanbrunt:
Since `StatementParser` didn't exist when the wrapper decorators were written, those methods were still relying of `parse_quoted_string()` to parse out the args into a list. I removed that function and replaced it with one that uses a `StatementParser` to parse the arguments.

@kotfu I removed a unit test called `test_arglist_decorator_twice`. Is that an OK change? I wasn't sure the use case of having two `with_argument_list` decorators on one method. Let me know if I need to put it back and make it work with the current code.


kmvanbrunt:
Nevermind, this needs work.

PR #638, opened 2019-03-05
Added a shlex.split() wrapper to have a common way of calling it.

kmvanbrunt:
This is pretty much a refactor that provides a common way of calling `shlex.split()` to ensure the same arguments are always used for consistent parsing.

I also replaced `parse_quoted_string()` with `_get_command_arg_list()`. I believe the name and documentation of this function better communicates its purpose. This function also doesn't bother calling `shlex.split()` on an already parsed `Statement` object.

@kotfu I removed a unit test called `test_arglist_decorator_twice`. Is that an OK change? I wasn't sure the use case of having two `with_argument_list` decorators on one method. Let me know if I need to put it back and make the new code work with it. It would be a trivial fix for me.

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -35,7 +35,6 @@ import glob
 import inspect
 import os
 import re
-import shlex
 import sys
 import threading
 from typing import Any, Callable, Dict, List, Mapping, Optional, Tuple, Type, Union, IO
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -158,7 +158,7 @@ def with_category(category: str) -> Callable:
     return cat_decorator
 
 
-def _get_command_arg_list(to_parse: Union[str, Statement], preserve_quotes: bool) -> List[str]:
+def _get_command_arg_list(to_parse: Union[Statement, str], preserve_quotes: bool) -> List[str]:
     """
     Called by the argument_list and argparse wrappers to retrieve just the arguments being
     passed to their do_* methods as a list.
@@ -198,7 +198,7 @@ def with_argument_list(*args: List[Callable], preserve_quotes: bool = False) -> 
 
     def arg_decorator(func: Callable):
         @functools.wraps(func)
-        def cmd_wrapper(cmd2_instance, statement: Union[str, Statement]):
+        def cmd_wrapper(cmd2_instance, statement: Union[Statement, str]):
             parsed_arglist = _get_command_arg_list(statement, preserve_quotes)
             return func(cmd2_instance, parsed_arglist)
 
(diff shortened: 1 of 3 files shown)


cmd2/cmd2.py
```diff
@@ -176,10 +158,37 @@ def cat_decorator(func):
...
+
+    :param to_parse: what is being passed to the do_* method. It can be one of two types:
+                     1. An already parsed Statement
+                     2. An argument string in cases where a do_* method is explicitly called
+                        e.g.: Calling do_help('alias create') would cause to_parse to be 'alias create'
+
+    :param preserve_quotes: if True, then quotes will not be stripped from the arguments
+    :return: the arguments in a list
+    """
+    if isinstance(to_parse, Statement):
```
tleonhardt: You can add an initial if to check if it is an instance of list and if so, then just return it as-is.  Doing this will make the decorator idempotent so that you can call it twice and leave in the unit test.

cmd2/cmd2.py
```diff
@@ -150,24 +150,6 @@ def categorize(func: Union[Callable, Iterable], category: str) -> None:
...
-    Parse a quoted string into a list of arguments
-    :param string: the string being parsed
-    :param preserve_quotes: if True, then quotes will not be stripped
-    """
-    if isinstance(string, list):
-        # arguments are already a list, return the list we were passed
-        lexed_arglist = string
-    else:
-        # Use shlex to split the command line into a list of arguments based on shell rules
-        lexed_arglist = shlex.split(string, comments=False, posix=False)
```
tleonhardt: I don’t think this file uses shlex anymore and the unused import is causing a flake8 unit test failure.

cmd2/cmd2.py
```diff
@@ -189,9 +198,9 @@ def with_argument_list(*args: List[Callable], preserve_quotes: bool = False) ->
 
     def arg_decorator(func: Callable):
         @functools.wraps(func)
-        def cmd_wrapper(self, cmdline):
-            lexed_arglist = parse_quoted_string(cmdline, preserve_quotes)
-            return func(self, lexed_arglist)
+        def cmd_wrapper(cmd2_instance, statement: Union[Statement, str]):
```
tleonhardt: I like the improved variable names

cmd2/parsing.py
```diff
@@ -607,6 +607,16 @@ def _command_and_args(tokens: List[str]) -> Tuple[str, str]:
 
         return command, args
 
+    @staticmethod
+    def shlex_split(str_to_split: str) -> List[str]:
```
tleonhardt: This doesn’t quite feel right to me as a static method in StatementParser.  What would you think about making it a standalone function in either parsing.py (same file)?

cmd2/cmd2.py
```diff
@@ -176,10 +158,37 @@ def cat_decorator(func):
     return cat_decorator
 
 
+def _get_command_arg_list(to_parse: Union[Statement, str], preserve_quotes: bool) -> List[str]:
```
tleonhardt: I recommend moving this helper function to be a standalone function in parsing.py

cmd2/cmd2.py
tleonhardt: That being said, I don't really see a use case for calling this decorator twice ...

kotfu:
Now that it's merged, here's my comment on the question @kmvanbrunt asked above, i.e. what's the use case for duplicate argils decorators on a single method. It was just an application of Postel's law, which is: be conservative in what you do, be liberal in what you accept from others. I have no use case for why you would need two decorators on the method, but since it was trivial to ensure that our decorator didn't break or give unexpected behavior if someone did apply it twice, I tested for it and make sure it worked.

PR #645, opened 2019-03-10
History improvements

### LINKED ISSUE #640: Update history Sphinx documentation to help avoid confusion
The [history documentation](https://cmd2.readthedocs.io/en/latest/freefeatures.html?highlight=history#searchable-command-history) could use an extensive update/re-write to better explain:
1. That there are two logically different types of history within ``cmd2`` - **readline history** and the **history command** history
1. How and why the two differ
1. Much more info on the history command and how to use it

See #639 for more info on one case of user confusion.

### LINKED ISSUE #641: Add support for negative indices to history command
The built-in **history** command has a very powerful and flexible indexing/slicing syntax.  One thing which it doesn't currently support is negative indices.  It would be much more Pythonic to support negative indexing, so we should do it.

Make sure we update the documentation when this feature is added.

kotfu:
Improve history and associated documentation as discussed in #640 and #641.

Closes #640 
Closes #641


cmd2/history.py
```diff
@@ -60,108 +60,154 @@ def pr(self, script=False, expanded=False, verbose=False) -> str:
 
 
 class History(list):
-    """ A list of HistoryItems that knows how to respond to user requests. """
+    """A list of HistoryItems that knows how to respond to user requests.
```
tleonhardt: Nice job improving comments throughout

docs/freefeatures.rst
```diff
@@ -253,9 +253,198 @@ the readline history.
...
+
+You can use negative numbers as either the first or second number of the range (but not both). If
+you want to display the last three commands entered::
+
+    (Cmd) history -- -3:
+        2  alias create two !echo two
+        3  alias create three !echo three
+        4  alias create four !echo four
+
+Notice the double dashes. These are required because the history command uses `argparse` to parse
```
tleonhardt: Ah, good documentation.  It explains the weirdness I was experiencing.  I guess it is treating the "-" as a dash and attempting to match a flag.

tests/test_history.py
```diff
@@ -63,26 +63,62 @@ def hist():
...
+    assert hist.span('2..') == ['second', 'third', 'fourth']
+    assert hist.span('2:') == ['second', 'third', 'fourth']
+
+    assert hist.span('-2..') == ['third', 'fourth']
+    assert hist.span('-2:') == ['third', 'fourth']
+
+    assert hist.span('1..3') == ['first', 'second', 'third']
+    assert hist.span('1:3') == ['first', 'second', 'third']
+
+    assert hist.span(':-2') == ['first', 'second']
```
tleonhardt: This unit test works as intended and as I would expect, however when I use it with the history command in a ``cmd2`` app, it appears to treat the trailing negative index as exclusive whereas as positive one is treated as inclusive.

docs/freefeatures.rst
kmvanbrunt: @kotfu @tleonhardt 
This is the official description of '--' from [argparse documentation](https://docs.python.org/3/library/argparse.html).

> If you have positional arguments that must begin with - and don’t look like negative numbers, you can insert the pseudo-argument '--' which tells parse_args() that everything after that is a positional argument

Maybe we can put a summary of this in freefeatures.rst.

docs/freefeatures.rst
```diff
@@ -306,25 +310,26 @@ will continue to the end::
         3  alias create three !echo three
         4  alias create four !echo four
 
-You can use negative numbers as either the first or second number of the range (but not both). If
-you want to display the last three commands entered::
+You can use negative numbers as either the first or second number of the range
+(but not both). If you want to display the last three commands entered::
```
tleonhardt: This is a weird limitation.  But if is well documented so no complaints.

tleonhardt:
Make surge a single ending negative index in a span works correctly.  Other than that looks good I think.

cmd2/cmd2.py
```diff
@@ -3231,15 +3231,22 @@ def do_history(self, args: argparse.Namespace) -> None:
             # If a character indicating a slice is present, retrieve
             # a slice of the history
             arg = args.arg
+            arg_is_int = False
+            try:
+                _ = int(arg)
```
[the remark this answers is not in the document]
kotfu: `history 1:-1` returns everything from the first command to the next to last command, inclusive. If you only have two commands in your history, the first command and the next to last command are the same command, so you will only get one result. If you have three commands in your history, `history 1:-1` will get the first two.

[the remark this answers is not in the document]
kotfu: I get no flake error on my system when using `$ invoke flake8`. However, removing the assignment on the int cast is cleaner anyway. I'll commit a change.

docs/freefeatures.rst
kotfu: Added.

docs/freefeatures.rst
kotfu: Fixed this so it works if both indices are negative.

kotfu:
@tleonhardt, as I understand your feedback, an ending negative index should include the referenced item. If I had four items in the history, and typed `history 1:-1` you would expect all four items to be returned.

Here was my thinking when I implemented the current behavior. Python's native zero-based indices have `0` as the first element of the array, and `-1` as the last element of the array, like so:
```
 0  1  2  3  4  5  6  7  8 
 S  O  M  E  W  O  R  D  S
-9 -8 -7 -6 -5 -4 -3 -2 -1
```

If we shift the positive indices to the left to make them 1 based, should we not also shift the negative indices to the left on my lame ASCII chart as well? If we don't, then the size of the slice returned by `history 1:-1` changes. With native zero based indices (on a list with 3 or more elements) `list[1:-1]` returns a list two shorter than the original list. If I implement your feedback, `history 1:-1` returns a list only one shorter than the original list.

I am open to either solution, but just wanted to explain my thinking. Lemme know if you still think I should make the requested modification, and I'll be happy to do it.

tleonhardt:
@kotfu 
We are in really unfamiliar territory here since with the **history** command, we are doing something significantly different than either ``Python`` or ``Bash history``.

In Python, list indexing is 0-based and list slicing is inclusive of the start index and exclusive of the end index.  In Bash, the **history** command there is no slicing or range extraction - ``history`` by itself returns the entire history and ``history 5`` returns the last 5 elements of the history.

In the ``cmd2`` **history** command, ``history 5`` returns the 5th element of the history and ``history 1:3`` returns the first three history elements - so it is 1-based indexing and slicing is inclusive of both start and end index.  If we say that ``history -1`` returns the last element in the history, then I would think that slicing should still be inclusive of both indices.  Adjusting negative indices to be "1-based" would result in the crazy situation that index 0 would refer to the last element - which I'm sure none of us want ;-)

I'm pretty flexible to how we do things, I just want to make sure we are internally logically consistent and well documented.   Three possibilities I see include:
1. Keep things mostly the way they are now and just make  slicing always inclusive of both the start and end index, even when the end index is negative
    * This is the least work and most backward compatible solution
1. Move to an entirely Pythonic list indexing and slicing syntax where it is 0-based and exclusive of the end index
    * This breaks backwards compatibility, but would potentially be good if we expect end users of ``cmd2`` applications to be familiar with Python
1. Keep things mostly the way they are now, but make slicing always exclusive of the end index like Python.
    * This mildly breaks backward compatibility, but honestly I don't see a real advantage here

What do you folks think?

kotfu:
After thinking about it for a day, I think I like option #1. It's much easier to explain to a regular person that "1 is the first command, -1 is the last command. 2 is the second command, -2 is the second to last command." Having it be inclusive make it nice and easy to explain, and I'm a big fan of clarity.

I'll make the change and get it pushed so you can review.

cmd2/history.py
```diff
@@ -60,108 +60,167 @@ def pr(self, script=False, expanded=False, verbose=False) -> str:
...
+    #    (?P<end>-?[1-9]{1}\d*)?       create a capture group named 'end' which matches an
+    #                                  optional minus sign, followed by exactly one non-zero
+    #                                  digit, and as many other digits as you want. This group is
+    #                                  optional so that we can match an input string like ':'
+    #                                  or '5:'. This regex will match 1, -1, 10, -10, but not
+    #                                  0 or -0.
+    #
+    #    \s*$                          match any whitespace at the end of the input. This is here so
+    #                                  you don't have to trim the input
+    #
```
tleonhardt: Things look good.  I’m not at my computer where I could manually test.  But as far as I can tell, this PR now looks great.

tleonhardt:
I did some manual testing and am now very happy with the way things work in this PR.  Nice job!

PR #648, opened 2019-03-15
Converted class attributes to instance attributes

### LINKED ISSUE #548: Consider converting cmd2.Cmd class attributes to instance attributes

tleonhardt:
This PR converts essentially all class attributes of ``cmd.Cmd`` to instance attributes. 

The following class attributes were added to ``cmd.Cmd`` for reasons of performance and/or convenience:
* `DEFAULT_SHORTCUTS`
* `DEFAULT_EDITOR`

Also:
* Cleaned up some PyCharm warnings
* Added ``allow_redirection``, ``terminators``, ``multiline_commands``, and ``shortcuts`` as optional arguments to ``cmd.Cmd.__init__()``
* Made the following attributes of ``StatementParser`` immutable:  ``terminators``, ``multiline_commands``, and ``shortcuts``
* Added ``allow_redirection`` and ``shortcuts`` read-only properties to ``cmd.Cmd`` for convenience in accessing these attributes from within ``self.statement_parser``
    * There are no longer certain attributes which need to be manually set before calling ``super.__init__()``
* Moved code for finding a sane default editor inside ``utils.find_editor()`` function
* Updated documentation and examples accordingly
* Deleted the following members of ``cmd2.Cmd`` that were now unnecessary:
    * ``reserved_words``
    * ``keywords``
    * ``_finalize_app_parameters()``

## TODO
* [x] Merge from master
* [x] Update CHANGELOG

This closes #548

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -3392,7 +3392,7 @@ class Cmd(cmd.Cmd):
 
         command = utils.quote_string_if_needed(os.path.expanduser(self.editor))
         if args.file_path:
-            command += " " + utils.quote_string_if_needed(os.path.expaclass nduser(args.file_path))
+            command += " " + utils.quote_string_if_needed(os.path.expanduser(args.file_path))
 
         self.do_shell(command)
 
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -297,37 +297,7 @@ class Cmd(cmd.Cmd):
     Line-oriented command interpreters are often useful for test harnesses, internal tools, and rapid prototypes.
     """
     DEFAULT_SHORTCUTS = {'?': 'help', '!': 'shell', '@': 'load', '@@': '_relative_load'}
-
-    # Attributes which ARE dynamically settable at runtime
-    colors = constants.COLORS_TERMINAL
-    continuation_prompt = '> '
-    debug = False
-    echo = False
-    editor = os.environ.get('EDITOR')
-    if not editor:
-        if sys.platform[:3] == 'win':
-            editor = 'notepad'
-        else:
-            # Favor command-line editors first so we don't leave the terminal to edit
-            for editor in ['vim', 'vi', 'emacs', 'nano', 'pico', 'gedit', 'kate', 'subl', 'geany', 'atom']:
-                if utils.which(editor):
-                    break
-    feedback_to_output = False  # Do not include nonessentials in >, | output by default (things like timing)
-    locals_in_py = False
-    quiet = False  # Do not suppress nonessential output
-    timing = False  # Prints elapsed time for each command
-
-    # To make an attribute settable with the "do_set" command, add it to this ...
-    settable = {'colors': 'Allow colorized output (valid values: Terminal, Always, Never)',
-                'continuation_prompt': 'On 2nd+ line of input',
-                'debug': 'Show full error stack on error',
-                'echo': 'Echo command issued into output',
-                'editor': 'Program used by ``edit``',
-                'feedback_to_output': 'Include nonessentials in `|`, `>` results',
-                'locals_in_py': 'Allow access to your application in py via self',
-                'prompt': 'The prompt issued to solicit input',
-                'quiet': "Don't print nonessential feedback",
-                'timing': 'Report execution times'}
+    DEFAULT_EDITOR = utils.find_editor()
 
     def __init__(self, completekey: str = 'tab', stdin=None, stdout=None, persistent_history_file: str = '',
                  persistent_history_length: int = 1000, startup_script: Optional[str] = None, use_ipython: bool = False,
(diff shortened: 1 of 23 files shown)


cmd2/cmd2.py
```diff
@@ -379,21 +373,20 @@ def __init__(self, completekey: str = 'tab', stdin=None, stdout=None, persistent
...
-            shortcuts=self.shortcuts,
-        )
+
+        if shortcuts is None:
+            shortcuts = self.DEFAULT_SHORTCUTS
+        shortcuts = sorted(shortcuts.items(), reverse=True)
+        self.statement_parser = StatementParser(allow_redirection=allow_redirection,
+                                                terminators=terminators,
+                                                multiline_commands=multiline_commands,
+                                                aliases=self.aliases,
```
kmvanbrunt: I think a method to retrieve aliases should be passed into `self.statement_parser`. Just in case someone ever recreates `self.aliases` after the creation of `self.statement_parser`.

cmd2/parsing.py
```diff
@@ -270,31 +270,42 @@ class StatementParser:
...
         else:
-            self.terminators = terminators
+            self.terminators = tuple(terminators)
         if multiline_commands is None:
-            self.multiline_commands = []
+            self.multiline_commands = ()
         else:
-            self.multiline_commands = multiline_commands
+            self.multiline_commands = tuple(multiline_commands)
         if aliases is None:
```
kmvanbrunt: Can aliases ever be none?

tests/test_completion.py
```diff
@@ -635,7 +635,7 @@ def test_tokens_for_completion_redirect(cmd2_app):
     endidx = len(line)
     begidx = endidx - len(text)
 
-    cmd2_app.allow_redirection = True
+    cmd2_app.statement_parser.allow_redirection = True
```
kmvanbrunt: Why not have these use `cmd2_app's` `allow_redirection` property instead?

cmd2/cmd2.py
tleonhardt: I was debating doing one of two things for this:
1. Pass in a callback function to the ``StatementParser`` that it uses to get the latest aliases
2. Create a read/write property in ``cmd2.Cmd`` that is used to access the master copy of aliases within ``StatementParser`` and not maintain its own copy

Do you think that one of this approaches has architectural advantages over the other?

cmd2/parsing.py
tleonhardt: No, but I did it this way for symmetry with the other arguments passed in to ``StatementParser``.  

If you think it is cleaner, I could remove the default values for the following arguments to ``StatementParser.__init__()`` since they should never be ``None`` when coming from ``cmd2.Cmd``:
* ``aliases``
* ``allow_redirection`` 
* ``shortcuts``

tests/test_completion.py
tleonhardt: Currently the ``allow_redireciton`` property is read-only - i.e. it provides the automatic getter, but not a corresponding setter.  I did debate making this a read-write property that provides both a getter and a setter.  Do you think that would be a good idea?

cmd2/cmd2.py
kmvanbrunt: Normally I'd say having the parser call back up to its owner isn't the best approach. But we do that in other places in `cmd2`. However, since the parser already owns the other settings, maybe just let it own aliases, too.

tleonhardt: Aliases are now stored solely in the ``StatementParser`` instance and I added a read-only property to ``cmd2.Cmd`` to get the aliases.

cmd2/parsing.py
tleonhardt: With the modifications I made, ``aliases`` can be (and are typically) ``None``

tests/test_completion.py
tleonhardt: I added a setter for the ``allow_redirection`` property and modified the unit test to use it.

cmd2/cmd2.py
```diff
@@ -553,10 +544,25 @@ def visible_prompt(self) -> str:
         """
         return utils.strip_ansi(self.prompt)
 
-    def _finalize_app_parameters(self) -> None:
-        """Finalize the shortcuts"""
-        # noinspection PyUnresolvedReferences
-        self.shortcuts = sorted(self.shortcuts.items(), reverse=True)
+    @property
+    def aliases(self) -> bool:
```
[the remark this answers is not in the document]
tleonhardt: That would be a cut-and-paste typo.  Thanks for catching it.

tleonhardt:
@kotfu If you get a chance to quickly look over this, I would appreciate your feedback.  I'll probably merge it in in another day or so.

tleonhardt:
I'm going to merge this now.  But if anyone has any recommendations, I am happy to address them in another PR.

PR #653, opened 2019-03-19
Pyscript fix

### LINKED ISSUE #613: Redirecting pyscript with app() commands closes cmd2

kmvanbrunt:
Fixes #613 

The root cause of #613 was `onecmd_plus_hooks()` calling `_restore_output()` if `self.redirecting` was **True**. While **pyscript** was redirecting its output to a file, the `PyscriptBridge` would call `onecmd_plus_hooks()` to handle `app('help')`. After the help command ended, `_restore_output()` would be called because  `self.redirecting` was **True** and therefore the file handle **pyscript** was using to redirect its output would be closed.

The new code keeps track of redirection for each command. Therefore `_restore_output()` only restores stuff if the current command set up redirection.

This also enhances our **pyscript** functionality quite a bit. We can now independently redirect each command running within the script. Therefore the outer **pyscript** call may be writing its output to a file while commands being run within the script can write anywhere they want, including piping to another process.

1. app("help > help.txt")
2. app("!cat file | grep string")

This PR also made a change to the piped process in redirection. We now set its **stdout** to be `self.stdout`.

Lastly, I no longer append a `\n` to commands in PyscriptBridge. This was originally done to assist in running multiline commands in a pyscript file. However, it had the unintended consequence of breaking redirection with non-multiline commands. From now on users are expected to include a terminator in multiline commands being run in a pyscript. This fits the expected syntax anyway since commands passed to `app()` should be the same as they would be typed on the command line.

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -1902,21 +1902,21 @@ class Cmd(cmd.Cmd):
             read_fd, write_fd = os.pipe()
 
             # Open each side of the pipe and set stdout accordingly
-            subproc_stdin = io.open(read_fd, 'r')
-            new_stdout = io.open(write_fd, 'w')
+            pipe_read = io.open(read_fd, 'r')
+            pipe_write = io.open(write_fd, 'w')
 
             # We want Popen to raise an exception if it fails to open the process.  Thus we don't set shell to True.
             try:
-                pipe_proc = subprocess.Popen(statement.pipe_to, stdin=subproc_stdin)
+                pipe_proc = subprocess.Popen(statement.pipe_to, stdin=pipe_read, stdout=self.stdout)
                 ret_val = RedirectionSavedState(self_stdout=self.stdout,
                                                 sys_stdout=None,
                                                 pipe_proc=self.pipe_proc)
-                self.stdout = new_stdout
+                self.stdout = pipe_write
                 self.pipe_proc = pipe_proc
             except Exception as ex:
                 self.perror('Not piping because - {}'.format(ex), traceback_war=False)
-                subproc_stdin.close()
-                new_stdout.close()
+                pipe_read.close()
+                pipe_write.close()
 
         elif statement.output:
             import tempfile
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -1848,29 +1848,9 @@ class Cmd(cmd.Cmd):
             # if we get here we must have:
             #   - a multiline command with no terminator
             #   - a multiline command with unclosed quotation marks
-            if not self.quit_on_sigint:
-                try:
-                    self.at_continuation_prompt = True
-                    newline = self.pseudo_raw_input(self.continuation_prompt)
-                    if newline == 'eof':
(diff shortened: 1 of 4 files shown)


cmd2/cmd2.py
```diff
@@ -3963,28 +3966,3 @@ def register_cmdfinalization_hook(self, func: Callable[[plugin.CommandFinalizati
         """Register a hook to be called after a command is completed, whether it completes successfully or not."""
         self._validate_cmdfinalization_callable(func)
         self._cmdfinalization_hooks.append(func)
-
-
-class Statekeeper(object):
```
tleonhardt: Nice job finally getting rid of this

cmd2/cmd2.py
```diff
@@ -2189,27 +2188,27 @@ def _cmdloop(self) -> bool:
                         self.poutput('{}{}'.format(self.prompt, line))
                 else:
                     # Otherwise, read a command from stdin
-                    if not self.quit_on_sigint:
-                        try:
-                            line = self.pseudo_raw_input(self.prompt)
-                        except KeyboardInterrupt:
+                    try:
+                        line = self.pseudo_raw_input(self.prompt)
+                    except KeyboardInterrupt as ex:
```
tleonhardt: I like how you cleaned up handling this case

tleonhardt:
I'm going to merge this in because I want to test it in combination with other changes.  We can always modify it later.  But so far it seems like a great improvement.

Issue #654, opened 2019-03-20
Investigate potential StdSim and/or ByteBuf improvements


tleonhardt:
The `StdSim` and `ByteBuf` classes were initially created as a convenient hack for use in unit and transcript testing with little to no thought given to factors like performance.  However, ``StdSim`` is now used by *pyscript_bridge.py* and hence by the **py**, **pyscript**, and **ipy** commands.

Issues that should be investigated include the following:
1. `ByteBuf` uses an immutable `bytes` object as a buffer to store text in the inner buffer.  If a lot of data gets sent here, constantly recreating new versions might be inefficient.  We should consider replacing it with some mutable type like a `bytearray`.
1. `StdSim` should probably behave like it's inner_stream. 
    * If the `inner_stream` is a `_io.TextIOWrapper` like `sys.stdout` typically is, then `StdSim` should provide the `buffer` interface.
    * If the `inner_stream` is a _`io.BufferedWriter` like a file opened in binary mode, then no `buffer` interface would be provided.
    *  In both cases, the `write` method should behave like the inner stream.
        * In the text case, `write(`) should accept strings, and in the binary case, bytes.

These changes will support wrapping `stdout` no matter what type of stream it is.

kmvanbrunt:
After further thought, I think adding support for binary streams won't be necessary since **cmd2** assumes `stdout` is a text stream in all of its code.  That's why we call `self.stdout.write()` with a string. If we support binary streams, then these calls would need to be changed to encode the string into bytes before calling write. In a terminal application, I can't think of a good use case for that.

I did make some other enhancements to the `StdSim` class that I will create a PR for and close this ticket. After these are merged in, I believe `StdSim` to be in a good state.

PR #655, opened 2019-03-21
Capture popen

### LINKED ISSUE #657: Stop supporting cmd2 outside of main thread
Much of **cmd2** relies on catching `KeyboardInterrupts`, which means it has to be run in the main thread.  In issue #584, we added a check in `cmdloop()` to prevent installing a signal handler if it is not running in the main thread.

Much of the new additions, especially multiple levels of redirection/piping rely on the user being able to hit Ctrl-C to raise a `KeyboardInterrupt`. I propose we reverse the fix made in #584 since so much of the code can't work outside of the main thread.

kmvanbrunt:
Added ability for pyscripts to capture stdout and stderr streams of shell commands and processes we are piping to.

Also fixes #657

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -694,22 +694,21 @@ class Cmd(cmd.Cmd):
                     pager = self.pager
                     if chop:
                         pager = self.pager_chop
-                    self.pipe_proc = subprocess.Popen(pager, shell=True, stdin=subprocess.PIPE)
+                    pipe_proc = subprocess.Popen(pager, shell=True, stdin=subprocess.PIPE)
                     try:
-                        self.pipe_proc.stdin.write(msg_str.encode('utf-8', 'replace'))
-                        self.pipe_proc.stdin.close()
+                        pipe_proc.stdin.write(msg_str.encode('utf-8', 'replace'))
+                        pipe_proc.stdin.close()
                     except (OSError, KeyboardInterrupt):
                         pass
 
                     # Less doesn't respect ^C, but catches it for its own UI purposes (aborting search etc. inside less)
                     while True:
                         try:
-                            self.pipe_proc.wait()
+                            pipe_proc.wait()
                         except KeyboardInterrupt:
                             pass
                         else:
                             break
-                    self.pipe_proc = None
                 else:
                     self.decolorized_write(self.stdout, msg_str)
             except BrokenPipeError:
@@ -1655,7 +1654,7 @@ class Cmd(cmd.Cmd):
         :param frame
         """
 
-        # Save copy of pipe_proc since it could theoretically change while this is running
+        # Save copy of pipe_proc_reader since it could theoretically change while this is running
         pipe_proc_reader = self.pipe_proc_reader
 
         if pipe_proc_reader is not None:
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -2941,10 +2941,14 @@ class Cmd(cmd.Cmd):
 
         expanded_command = ' '.join(tokens)
 
-        # Use pipes for stdout and stderr so they can be captured by our process
-        proc = subprocess.Popen(expanded_command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True)
-        proc_reader = utils.ProcReader(proc, self.stdout, sys.stderr)
-        proc_reader.wait()
+        # If stdout is a StdSim, then we will use pipes so we can save stdout and stderr output
+        if isinstance(self.stdout, utils.StdSim):
+            proc = subprocess.Popen(expanded_command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True)
+            proc_reader = utils.ProcReader(proc, self.stdout, sys.stderr)
+            proc_reader.wait()
+        else:
+            proc = subprocess.Popen(expanded_command, stdout=self.stdout, shell=True)
+            proc.communicate()
 
     @staticmethod
     def _reset_py_display() -> None:
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -434,7 +434,7 @@ class Cmd(cmd.Cmd):
         self._script_dir = []
 
         # A flag used to protect critical sections in the main thread from stopping due to a KeyboardInterrupt
-        self.sigint_protection = utils.ContextFlag(False)
+        self.sigint_protection = utils.ContextFlag()
 
         # When this is not None, then it holds a ProcReader for the pipe process created by the current command
         self.cur_pipe_proc_reader = None
@@ -711,9 +711,7 @@ class Cmd(cmd.Cmd):
                     # still receive the SIGINT since it is in the same process group as us.
                     with self.sigint_protection:
                         pipe_proc = subprocess.Popen(pager, shell=True, stdin=subprocess.PIPE)
-                        pipe_proc.stdin.write(msg_str.encode('utf-8', 'replace'))
-                        pipe_proc.stdin.close()
-                        pipe_proc.communicate()
+                        pipe_proc.communicate(msg_str.encode('utf-8', 'replace'))
                 else:
                     self.decolorized_write(self.stdout, msg_str)
             except BrokenPipeError:
@@ -1728,13 +1726,10 @@ class Cmd(cmd.Cmd):
             saved_state = None
 
             try:
+                # Get sigint protection while we set up redirection
                 with self.sigint_protection:
-                    # Set up our redirection state variables
                     redir_error, saved_state = self._redirect_output(statement)
                     self.cur_pipe_proc_reader = saved_state.pipe_proc_reader
-
-                    if self._in_py:
-                        self._last_result = None
 
                 # Do not continue if an error occurred while trying to redirect
                 if not redir_error:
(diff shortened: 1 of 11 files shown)


tleonhardt:
Does this close #654?

tleonhardt:
Just curious, what wasn't ready about this code change?

kmvanbrunt:
Some little edge cases. For instance, waiting for a subprocess to end could get interrupted and the process would become a zombie until cmd2 exited.

I'm getting close.

tleonhardt:
Yeah, all of the subprocess stuff took a lot of work to get reliable and small changes can make it brittle.

kmvanbrunt:
Try a big change like the multiple levels of redirection caused by allowing pyscript to redirect. I am learning a lot though, so it's been fun.

cmd2/utils.py
```diff
@@ -317,6 +344,11 @@ def clear(self) -> None:
         """Clear the internal contents"""
         self.buffer.byte_buf = b''
```
tleonhardt: If you are changing internal representation to a bytearray, then this needs to be changed too

tests/test_argparse.py
```diff
@@ -112,62 +112,60 @@ def do_talk(self, args, extra):
...
-    run_cmd(argparse_app, 'speak "')
-    out, err = capsys.readouterr()
-    assert err == "Invalid syntax: No closing quotation\n"
+def test_invalid_syntax(argparse_app):
+    out, err = run_cmd(argparse_app, 'speak "')
+    assert err[0] == "Invalid syntax: No closing quotation"
 
 def test_argparse_basic_command(argparse_app):
-    out = run_cmd(argparse_app, 'say hello')
+    out, err = run_cmd(argparse_app, 'say hello')
```
tleonhardt: Capturing both stdout and stderr is a nice improvement for testability

cmd2/utils.py
```diff
@@ -317,6 +344,11 @@ def clear(self) -> None:
...
 
+    @staticmethod
+    def isatty() -> bool:
+        """StdSim will never be considered an interactive stream"""
+        return False
```
kmvanbrunt: I should probably change this to:
```Python
@staticmethod
def isatty() -> bool:
    if self.echo:
        return self.inner_stream.isatty()
    else:
        return False
```
@tleonhardt Thoughts?

tleonhardt: Yeah that probably makes sense, but if you do that make sure to update the comment 😀

tleonhardt:
Does this PR close any other Issues other than the one called out?

cmd2/cmd2.py
```diff
@@ -302,11 +302,20 @@ class EmptyStatement(Exception):
...
 DisabledCommand = namedtuple('DisabledCommand', ['command_function', 'help_function'])
 
-# Used to restore state after redirection ends
-# redirecting and piping are used to know what needs to be restored
-RedirectionSavedState = utils.namedtuple_with_defaults('RedirectionSavedState',
-                                                       ['redirecting', 'self_stdout', 'sys_stdout',
-                                                        'piping', 'pipe_proc'])
+
+class RedirectionSavedState(object):
+    # Created by each command to store information about their redirection
```
tleonhardt: Recommend converting this to a docstring comment

cmd2/cmd2.py
```diff
@@ -302,11 +302,20 @@ class EmptyStatement(Exception):
...
+    def __init__(self):
```
tleonhardt: Recommend taking `self_stdout`, `sys_stdout`, and `pipe_proc_reader` as required arguments and then using their values to assign the three ``self.saved_*`` attributes.

cmd2/cmd2.py
```diff
@@ -424,8 +433,13 @@ def __init__(self, completekey: str = 'tab', stdin=None, stdout=None, persistent
         # Used load command to store the current script dir as a LIFO queue to support _relative_load command
         self._script_dir = []
 
-        # Used when piping command output to a shell command
-        self.pipe_proc = None
+        # A flag used to protect critical sections in the main thread from stopping due to a KeyboardInterrupt
```
tleonhardt: Interesting/weird.  It is a context manager which is also treated as a boolean flag within the sigint handler.  I would recommend rewording the comments associated with this.  Maybe something like:
> A context manager used to protect critical sections in the main thread from stopping due to a KeyboardInterrupt

cmd2/cmd2.py
```diff
@@ -424,8 +433,13 @@ def __init__(self, completekey: str = 'tab', stdin=None, stdout=None, persistent
...
+        self.sigint_protection = utils.ContextFlag()
+
+        # If the current command created a process to pipe to, then this will be a ProcReader object.
+        # Otherwise the value will be None. This member is used to know when a pipe process can be killed
```
tleonhardt: Recommend trying to shorten this comment up a little bit.  Maybe something like:
```sh
# If the current command created a process to pipe to, then this will be a ProcReader object.
# Otherwise it will be None. Its used to know when a pipe process can be killed and/or waited upon.
```

cmd2/utils.py
```diff
@@ -342,27 +374,119 @@ def write(self, b: bytes) -> None:
...
-    if not editor:
-        if sys.platform[:3] == 'win':
-            editor = 'notepad'
-        else:
-            # Favor command-line editors first so we don't leave the terminal to edit
-            for editor in ['vim', 'vi', 'emacs', 'nano', 'pico', 'gedit', 'kate', 'subl', 'geany', 'atom']:
-                if which(editor):
-                    break
-    return editor
+class ContextFlag(object):
```
tleonhardt: I would recommend renaming and/or changing the docstring to make it clear that this is first and foremost a context manager which happens to be used as a boolean flag within the sigint handler.

tests/conftest.py
```diff
@@ -13,6 +14,13 @@
 import cmd2
 from cmd2.utils import StdSim
 
+# Python 3.4 require contextlib2 for temporarily redirecting stderr and stdout
```
tleonhardt: On a vaguely related note, I've been meaning to create an Issue to discuss when we should end support for Python 3.4.  Support for further bug fixes to Python 3.4 ended on March 15, 2019.  We should plan to continue to support it for awhile yet, but should probably end official support for it sometime in 2019 simply to ease the burden on our own maintenance.

tests/conftest.py
```diff
@@ -126,19 +134,33 @@ def normalize(block):
...
+        with redirect_stdout(copy_cmd_stdout):
+            with redirect_stderr(copy_stderr):
+                app.onecmd_plus_hooks(cmd)
+    finally:
+        app.stdout = copy_cmd_stdout.inner_stream
+        sys.stdout = saved_sysout
+
+    out = copy_cmd_stdout.getvalue()
+    err = copy_stderr.getvalue()
+    return normalize(out), normalize(err)
```
tleonhardt: I like this nice improvement to `run_cmd()` where it now returns the contents of both **stdout** and **stderr** - it largely eliminates the need to use the `copses` fixture.

cmd2/utils.py
```diff
@@ -342,27 +374,119 @@ def write(self, b: bytes) -> None:
...
+        self._err_thread = threading.Thread(name='out_thread', target=self._reader_thread_func,
+                                            kwargs={'read_stdout': False})
+
+        # Start the reader threads for pipes only
+        if self._proc.stdout is not None:
+            self._out_thread.start()
+        if self._proc.stderr is not None:
+            self._err_thread.start()
+
+    def send_sigint(self) -> None:
```
tleonhardt: Strongly recommend adding more complete unit tests for this new class including for the following methods:
* `send_sigint`
* `terminate`
* `wait` (case err thread is alive and there are an out and err to communicate)
* `_write_bytes` (else case where stream.write() is called)

cmd2/utils.py
```diff
@@ -342,27 +374,119 @@ def write(self, b: bytes) -> None:
...
+    Its main use is as a flag to prevent the SIGINT handler in cmd2 from raising a KeyboardInterrupt
+    while a critical code section has set the flag to True. Because signal handling is always done on the
+    main thread, this class is not thread-safe since there is no need.
+    """
+    def __init__(self):
+        # When this flag has a positive value, it is considered set.
+        # When it is 0, it is not set. It should never go below 0.
+        self.__count = 0
+
+    def __bool__(self):
```
tleonhardt: Strongly recommend adding more complete unit tests for this new class including for the following methods:
* `__bool__`
* `__exit__` (case where it raises a ValueError)

tests/conftest.py
tleonhardt: I created Issue #661 for this

cmd2/utils.py
```diff
@@ -342,27 +375,140 @@ def write(self, b: bytes) -> None:
...
+        # Start the reader threads for pipes only
+        if self._proc.stdout is not None:
+            self._out_thread.start()
+        if self._proc.stderr is not None:
+            self._err_thread.start()
+
+    def send_sigint(self) -> None:
+        """Send a SIGINT to the process similar to if <Ctrl>+C were pressed."""
+        import signal
+        if sys.platform.startswith('win'):
```
kmvanbrunt: What was the reason for changing SIGINT to CTRL_C_EVENT on Windows?

PR #656, opened 2019-03-22
Multiline prompt

kmvanbrunt:


cmd2/cmd2.py
```diff
@@ -2111,7 +2111,7 @@ def pseudo_raw_input(self, prompt: str) -> str:
                 else:
                     line = input()
                     if self.echo:
-                        sys.stdout.write('{}{}\n'.format(self.prompt, line))
+                        sys.stdout.write('{}{}\n'.format(prompt, line))
```
tleonhardt: These changes look fine.  I had never considered prompts with newlines, but I guess that is a valid scenario.

PR #663, opened 2019-04-10
Pyscript now saves command output during the same period that redirection does

### LINKED ISSUE #659: Consider what output the pyscript is capturing in its StdSim objects
**Question 1**
When we redirect a command, any output in **postparsing** and **cmdfinalization** hooks is not included in the redirected output. However, any output in **precmd** and **postcmd** hooks is redirected.

To be consistent, should an `app()` call in a **pyscript** capture this same output and exclude the **postparsing** and **cmdfinalization** hooks? Right now `app()` captures it all for `CommandResult.stdout`.
 
**Question 2**
Consider the command: `app('help > out.txt')`

What is the expected result?
1. `CommandResult.stdout` contains the output of the **help** command 
2. `CommandResult.stdout` is empty and `CommandResult.stderr` to contain text only if an error occurred redirecting output

Assume both cases do write the output of **help** to **out.txt** if no error occurred.

kmvanbrunt:
Fixes #659

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -1675,10 +1675,13 @@ class Cmd(cmd.Cmd):
         statement = self.statement_parser.parse_command_only(line)
         return statement.command, statement.args, statement.command_and_args
 
-    def onecmd_plus_hooks(self, line: str) -> bool:
+    def onecmd_plus_hooks(self, line: str, pyscript_bridge_call: bool = False) -> bool:
         """Top-level function called by cmdloop() to handle parsing a line and running the command and all of its hooks.
 
         :param line: line of text read from input
+        :param pyscript_bridge_call: This should only ever be set to True by PyscriptBridge to signify the beginning
+                                     of an app() call in a pyscript. It is used to enable/disable the storage of the
+                                     command's stdout.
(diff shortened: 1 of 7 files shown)


cmd2/cmd2.py
```diff
@@ -3026,7 +3036,6 @@ def do_py(self, args: argparse.Namespace) -> bool:
         if self._in_py:
             err = "Recursively entering interactive Python consoles is not allowed."
             self.perror(err, traceback_war=False)
-            self._last_result = CommandResult('', err)
```
tleonhardt: CommandResult is now an unused import which needs to be removed so flake8 checks will pass

tleonhardt:
I recommend adding something to the **CHANGELOG** in relation to this PR

kmvanbrunt:
Closing until I can make some final changes.

tests/pyscript/stdout_capture.py
```diff
@@ -0,0 +1,26 @@
+# flake8: noqa F821
+# This script demonstrates when output of a command finalization hook is captured by a pyscript app() call
```
tleonhardt: Thanks for adding a test

PR #664, opened 2019-04-11
Fixed issue where pointer to bytearray was being returned instead of copied bytes

kmvanbrunt:


cmd2/utils.py
```diff
@@ -325,7 +325,7 @@ def getvalue(self) -> str:
 
     def getbytes(self) -> bytes:
         """Get the internal contents as bytes"""
-        return self.buffer.byte_buf
+        return bytes(self.buffer.byte_buf)
```
tleonhardt: Good catch.  The change to `bytearray` for internal implementation of `ByteBuf` is since the last release, correct? (meaning there is no need to update the CHANGELOG)

kmvanbrunt: Correct

PR #666, opened 2019-04-25
Stdsim fixes

### LINKED ISSUE #654: Investigate potential StdSim and/or ByteBuf improvements

kmvanbrunt:
Closes #654 
Removed support for wrapping binary streams in `ProcReader`.

These changes also improve the flushing behavior of when `pyscript` echos the output of a shell command. Since we obtain that output over binary pipes, we echo it using `StdSim.inner_stream.buffer.write()` which does not follow line buffering behavior by default. The new code will call `flush()` if the bytes being written contain a newline or carriage return and the inner stream has line buffering turned on.

This makes echoed shell command output print at the same frequency it would when run at the terminal. Commands like `ping` are a good test case for this. ex: `app('ping localhost')`

### DIFF
--- a/cmd2/utils.py
+++ b/cmd2/utils.py
@@ -8,7 +8,7 @@ import subprocess
 import sys
 import threading
 import unicodedata
-from typing import Any, BinaryIO, Iterable, List, Optional, TextIO, Union
+from typing import Any, Iterable, List, Optional, TextIO, Union
 
 from wcwidth import wcswidth
 
@@ -297,7 +297,7 @@ class StdSim(object):
                  encoding: str = 'utf-8', errors: str = 'replace') -> None:
         """
         Initializer
-        :param inner_stream: the emulated stream
+        :param inner_stream: the wrapped stream. Should be a TextIO or StdSim instance.
         :param echo: if True, then all input will be echoed to inner_stream
         :param encoding: codec for encoding/decoding strings (defaults to utf-8)
         :param errors: how to handle encoding/decoding errors (defaults to replace)
@@ -350,6 +350,17 @@ class StdSim(object):
         else:
             return False
 
+    @property
+    def line_buffering(self) -> bool:
+        """
+        Handle when the inner stream doesn't have a line_buffering attribute which is the case
+        when running unit tests because pytest sets stdout to a pytest EncodedFile object.
+        """
+        try:
+            return self.inner_stream.line_buffering
+        except AttributeError:
+            return False
+
     def __getattr__(self, item: str):
         if item in self.__dict__:
             return self.__dict__[item]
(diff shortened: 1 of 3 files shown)


cmd2/utils.py
```diff
@@ -350,6 +350,17 @@ def isatty(self) -> bool:
         else:
             return False
 
+    @property
+    def line_buffering(self) -> bool:
+        """
```
tleonhardt: Consider adding a unit test or two to make sure pith positive and negative cases of this method are cover d.

cmd2/utils.py
```diff
@@ -374,14 +388,22 @@ def write(self, b: bytes) -> None:
         if self.std_sim_instance.echo:
             self.std_sim_instance.inner_stream.buffer.write(b)
 
+            # Since StdSim wraps TextIO streams, we will flush the stream if line buffering is on
+            # and the bytes being written contain a new line character. This is helpful when StdSim
+            # is being used to capture output of a shell command because it causes the output to print
+            # to the screen more often than if we waited for the stream to flush its buffer.
+            if self.std_sim_instance.line_buffering:
+                if any(newline in b for newline in ByteBuf.NEWLINES):
+                    self.std_sim_instance.flush()
```
tleonhardt: This flushing is a nice improvement

tests/test_utils.py
```diff
@@ -218,6 +207,26 @@ def test_stdsim_pause_storage(stdout_sim):
     stdout_sim.buffer.write(b_str)
     assert stdout_sim.getbytes() == b''
 
+def test_stdsim_line_buffering(base_app):
```
tleonhardt: Thanks for adding the unit test

PR #670, opened 2019-05-06
Macro refactor

kmvanbrunt:
* Fixed issue where aliases and macros could not contain terminator characters in their values
* History now shows what was typed for macros and not the resolved value by default. This is consistent with the behavior of aliases. Use the `expanded` or `verbose` arguments to `history` to see the resolved value for the macro.
* Added `_input_line_to_statement()` which wraps  `_complete_statement()` and resolves macros.
* Refactored alias resolution in parsing code

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@
       since the output will print at the same frequency as when the command is run in a terminal.
     * **ACArgumentParser** no longer prints complete help text when a parsing error occurs since long help messages
      scroll the actual error message off the screen.
-     * Exceptions occurring in tab completion functions are now printed to stderr before returning control back to
+    * Exceptions occurring in tab completion functions are now printed to stderr before returning control back to
     readline. This makes debugging a lot easier since readline suppresses these exceptions.
 * Potentially breaking changes
     * Replaced `unquote_redirection_tokens()` with `unquote_specific_tokens()`. This was to support the fix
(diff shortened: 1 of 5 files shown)


kmvanbrunt:
@kotfu Do you have an opinion on me adding a wrapper function around `_complete_statement()` called `_input_line_to_statement()`? It loops over  `_complete_statement()` and breaks when the returned statement does not contain a macro as its command. For each macro it sees, it resolves it to a new input line and sends that line through `_complete_statement()`.

The main change is that no one calls  `_complete_statement()`  directly anymore. They call the wrapper instead. If you think other users may have relied on  `_complete_statement()`, I can move all the logic into that function and remove the wrapper.

tleonhardt:
@kmvanbrunt My view is that any method in `cmd2` which begins with an underscore is a method we do not intend end users to call and for which we do not make any promises in regards to API stability.

CHANGELOG.md
```diff
@@ -1,16 +1,23 @@
 ## 0.9.13 (TBD, 2019)
 * Bug Fixes
     * Fixed issue where the wrong terminator was being appended by `Statement.expanded_command_line()`
+    * Fixed issue where aliases and macros could not contain terminator characters in their values
```
tleonhardt: Changes look good

tleonhardt:
Should any documentation updates be made in regards to this PR?

PR #671, opened 2019-05-06
Improved tab completion error feedback

### LINKED ISSUE #668: Give informative error message for invalid completer instead of empty completion results.
The following code suppresses an error when a completion function with an invalid signature is provided https://github.com/python-cmd2/cmd2/blob/de701086ff832bad0f0d97ffb10c2159d56ede7d/cmd2/argparse_completer.py#L691-L693

Instead of returning an empty list, I think that an informative error message explaining that the function signature is incorrect would be much more helpful. When I was implementing completion I was confused about why my completer wasn't returning any results until I traced it to this `return []`.

kmvanbrunt:
Fixes #668 
Exceptions occurring in tab completion functions are now printed to stderr before returning control back to readline.

No changes have been made to the argparse completion code yet. We need to figure out the best approach. Do we just remove the try/except statements? Or do we keep them and just call raise?

### DIFF
--- a/cmd2/argparse_completer.py
+++ b/cmd2/argparse_completer.py
@@ -667,7 +667,7 @@ class AutoCompleter(object):
 
                 if callable(arg_choices[0]):
                     completer = arg_choices[0]
-                elif isinstance(arg_choices[0], str) and callable(getattr(self._cmd2_app, arg_choices[0])):
+                else:
                     completer = getattr(self._cmd2_app, arg_choices[0])
 
                 # extract the positional and keyword arguments from the tuple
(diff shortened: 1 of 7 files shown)


cmd2/cmd2.py
```diff
@@ -1581,6 +1578,24 @@ def complete(self, text: str, state: int) -> Optional[str]:
...
+        control to readline.
+
+        :param text: the current word that user is typing
+        :param state: non-negative integer
+        """
+        # noinspection PyBroadException
+        try:
+            return self._complete_worker(text, state)
+        except Exception as e:
+            self.perror(e)
```
tleonhardt: This should be a big help to people for developers that are trying to debug tab-completion issues.

cmd2/argparse_completer.py
```diff
@@ -704,32 +701,17 @@ def _resolve_choices_for_arg(self, action: argparse.Action, used_values=()) -> L
...
-                    try:
-                        args = args(self._cmd2_app)
-                    except TypeError:
-                        args = args()
+                    args = args(self._cmd2_app)
                 except TypeError:
-                    return []
-
-            try:
-                iter(args)
```
tleonhardt: You dropped the `iter(args)` line which is what might have generated an exception prior to the next line.  I'm not sure what the effect would be, but are you sure this is the correct modification?

tleonhardt: Nevermind.  The list comprehension will throw an exception if args isn't iterable, so we don't need an explicit check.

PR #672, opened 2019-05-06
ACArgumentParser no longer prints complete help text when a parsing error occurs

### LINKED ISSUE #667: Don't print complete help text on argparse error
**ACArgumentParser** overrides **argparse.ArgumentParser's** `error()` function. One of its behaviors is to print all of the help text when there is a parsing error. Help text can be pretty long and has the tendency to scroll the actual error message off the screen.

I suggest we use the normal behavior of only printing the error and usage statement when a parsing error occurs.

kmvanbrunt:
Fixes #667

### DIFF
--- a/cmd2/argparse_completer.py
+++ b/cmd2/argparse_completer.py
@@ -908,7 +908,7 @@ class ACHelpFormatter(argparse.RawTextHelpFormatter):
                 # join lines into usage
                 usage = '\n'.join(lines)
 
-        # prefix with 'usage:'
+        # prefix with 'Usage:'
         return '%s%s\n\n' % (prefix, usage)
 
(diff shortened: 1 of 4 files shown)


cmd2/argparse_completer.py
```diff
@@ -1019,9 +1016,9 @@ def error(self, message: str) -> None:
                 formatted_message += '\n       ' + line
             linum += 1
 
-        sys.stderr.write(Fore.LIGHTRED_EX + '{}\n\n'.format(formatted_message) + Fore.RESET)
-        # sys.stderr.write('{}\n\n'.format(formatted_message))
-        self.print_help()
+        self.print_usage(sys.stderr)
```
tleonhardt: Seems like a good change

PR #676, opened 2019-05-13
Pipe chaining

kmvanbrunt:
- Added capability to redirect pipe commands and chain them together. This involved changing `Statement.pipe_to` from a list to a string.
- Fixed parsing issue in case where output redirection (e.g. > file) appears before a pipe. In that case, the pipe was given precedence even though it appeared later in the command.
- Preserving originally typed quotes of `Statement.output_to` for use in `Statement.post_command()`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,9 +15,11 @@
      scroll the actual error message off the screen.
     * Exceptions occurring in tab completion functions are now printed to stderr before returning control back to
     readline. This makes debugging a lot easier since readline suppresses these exceptions.
+    * Added capability to chain pipe commands and redirect their output (e.g. !ls -l | grep user | wc -l > out.txt)
 * Potentially breaking changes
     * Replaced `unquote_redirection_tokens()` with `unquote_specific_tokens()`. This was to support the fix
       that allows terminators in alias and macro values.
+    * Changed `Statement.pipe_to` to a string instead of a list 
 * **Python 3.4 EOL notice**
     * Python 3.4 reached its [end of life](https://www.python.org/dev/peps/pep-0429/) on March 18, 2019
     * This is the last release of `cmd2` which will support Python 3.4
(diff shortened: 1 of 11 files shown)


cmd2/argparse_completer.py
```diff
@@ -1051,7 +1049,7 @@ def format_help(self) -> str:
         formatter.add_text(self.epilog)
 
         # determine help from format above
-        return formatter.format_help()
+        return formatter.format_help() + '\n'
```
tleonhardt: This is fine, but could make use of `LINE_FEED` constant in [constants.py](https://github.com/python-cmd2/cmd2/blob/master/cmd2/constants.py#L21).

Alternatively, if we don't mind it being `\r\n` on Windows, there is the built-in `os.linesep` constant.

cmd2/cmd2.py
```diff
@@ -1519,18 +1519,19 @@ def _complete_worker(self, text: str, state: int) -> Optional[str]:
                             # Check if any portion of the display matches appears in the tab completion
                             display_prefix = os.path.commonprefix(self.display_matches)
 
-                            # For delimited matches, we check what appears before the display
-                            # matches (common_prefix) as well as the display matches themselves.
-                            if (' ' in common_prefix) or (display_prefix and ' ' in ''.join(self.display_matches)):
+                            # For delimited matches, we check for a space in what appears before the display
+                            # matches (common_prefix) as well as in the display matches themselves.
+                            if ' ' in common_prefix or (display_prefix
+                                                        and any(' ' in match for match in self.display_matches)):
```
tleonhardt: Just curious - why did you change the code this way?  It seems equivalent.  Is it just for readability?

CHANGELOG.md
```diff
@@ -5,7 +5,11 @@
     * History now shows what was typed for macros and not the resolved value by default. This is consistent with
     the behavior of aliases. Use the `expanded` or `verbose` arguments to `history` to see the resolved value for
     the macro.
+    * Fixed parsing issue in case where output redirection appears before a pipe. In that case, the pipe was given
```
tleonhardt: This PR contains a really nice set of enhancements and bug fixes that make `cmd2` applications behave significantly closer to how a "normal" shell behaves.  Nice work!

tleonhardt:
@kmvanbrunt @kotfu Do either of you understand the AppVeyor Windows CI failures?  From what I can see they are not actually related to this PR, but are some systematic problem with AppVeyor at the moment and/or how we have it configured.  However our [AppVeyor configuration](https://github.com/python-cmd2/cmd2/blob/master/.appveyor.yml) hasn't changed recently in 7 months.

tleonhardt:
I pushed a trivial update to to README.md on the *master* branch and verified that the AppVeyor issue is completely unrelated to anything on this branch.

cmd2/cmd2.py
kmvanbrunt: Seems like more efficient code since I'm no longer joining every match into one string before searching for a space.
The new code just iterates the list and stops the moment a space is found.

cmd2/argparse_completer.py
kmvanbrunt: We use `\n` and not `\r\n` everywhere else in **cmd2**, so I'm OK not using `os.linesep`.

PR #681, opened 2019-05-15
Ns provider

### LINKED ISSUE #596: Pass a custom Namespace to args
Kind of a follow up on https://github.com/python-cmd2/cmd2/issues/498

I want to pass a custom Namespace when parsing arguments while keeping the autocompletion:

```
    @with_argparser(parser)
    def do_load_pcap(self, cmdline):
          myNamespace = Namespace()
          myNamespace.toto = self.data
          args = parser.parse_args(shlex.split(cmdline), myNamespace)
          
```
I modified 
https://github.com/python-cmd2/cmd2/blob/b6c39365e0331ab64aad074565fe73fb074a3a7f/cmd2/cmd2.py#L270-L271
so that it doesn't parse stuff.
```
def with_argparser_test(argparser: argparse.ArgumentParser,
                   preserve_quotes: bool=False) -> Callable[[argparse.Namespace], Optional[bool]]:
    """A decorator to alter a cmd2 method to populate its ``args`` argument by parsing arguments
    with the given instance of argparse.ArgumentParser.

    :param argparser: unique instance of ArgumentParser
    :param preserve_quotes: if True, then arguments passed to argparse maintain their quotes
    :return: function that gets passed the argparse-parsed args
    """
    import functools

    def arg_decorator(func: Callable[[Any], Optional[bool]]):
        @functools.wraps(func)
        def cmd_wrapper(instance, cmdline):
            # lexed_arglist = parse_quoted_string(cmdline, preserve_quotes)
            # try:
            #     args = argparser.parse_args(lexed_arglist)
            # except SystemExit:
            #     return
            # else:
            return func(instance, cmdline)

        # argparser defaults the program name to sys.argv[0]
        # we want it to be the name of our command
        # argparser.prog = func.__name__[len(COMMAND_FUNC_PREFIX):]

        # If the description has not been set, then use the method docstring if one exists
        if argparser.description is None and func.__doc__:
            argparser.description = func.__doc__

        # Set the command's help text as argparser.description (which can be None)
        # cmd_wrapper.__doc__ = argparser.description

        # Mark this function as having an argparse ArgumentParser
        setattr(cmd_wrapper, 'argparser', argparser)

        return cmd_wrapper

    return arg_decorator

```
which can be used like
```
    @with_argparser_test(parser)
    def do_load_pcap(self, args):
        """
        Load the file as the current one
        """
        args = shlex.split(args)
        parser = self.do_load_pcap.argparser
        args = parser.parse_args(args)
        
        self.poutput("Loading %s" % args.input_file)

```

I wonder if cmd2 could propose some better interface. This is just an example, that adds a parameter "dont_parse" to the with_argparser decorator. If it's true, cmd2 could pass the parser along to the member function (instead of `parser = self.do_load_pcap.argparser` which requires some knowledge about cmd2 implementation)
```
@with_argparser(dont_quote, dont_parse=True)
def func(self, parser, args):
      parser.parse_args(args)
```

kmvanbrunt:
Added support for custom Namespaces in the argparse decorators
Fixes #596

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,10 +19,13 @@
      scroll the actual error message off the screen.
     * Exceptions occurring in tab completion functions are now printed to stderr before returning control back to
     readline. This makes debugging a lot easier since readline suppresses these exceptions.
+    * Added support for custom Namespaces in the argparse decorators. See description of `ns_provider` argument
+    for more information.
 * Potentially breaking changes
     * Replaced `unquote_redirection_tokens()` with `unquote_specific_tokens()`. This was to support the fix
       that allows terminators in alias and macro values.
-    * Changed `Statement.pipe_to` to a string instead of a list 
+    * Changed `Statement.pipe_to` to a string instead of a list
+    * `preserve_quotes` is now a keyword-only argument in the argparse decorators
 * **Python 3.4 EOL notice**
     * Python 3.4 reached its [end of life](https://www.python.org/dev/peps/pep-0429/) on March 18, 2019
     * This is the last release of `cmd2` which will support Python 3.4
--- a/docs/argument_processing.rst
+++ b/docs/argument_processing.rst
@@ -247,7 +247,7 @@ argument list instead of a string::
             pass
 
 
-Using the argument parser decorator and also receiving a a list of unknown positional arguments
+Using the argument parser decorator and also receiving a list of unknown positional arguments
 ===============================================================================================
 If you want all unknown arguments to be passed to your command as a list of strings, then
 decorate the command method with the ``@with_argparser_and_unknown_args`` decorator.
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -19,10 +19,13 @@
      scroll the actual error message off the screen.
     * Exceptions occurring in tab completion functions are now printed to stderr before returning control back to
     readline. This makes debugging a lot easier since readline suppresses these exceptions.
+    * Added support for custom Namespaces in the argparse decorators. See description of `ns_provider` argument
```
tleonhardt: This new feature would benefit from a mention in the Sphinx docs section about the argparse decorators and an example.

tleonhardt:
@teto Can you try out the change in the PR since it was done at your request?

teto:
sure, if you don't mind waiting a few days though (busy).

CHANGELOG.md
kmvanbrunt: Done

teto:
I may have asked this before but is there anyway to force the version in setuptools_scm ? 
Otherwise I hade to modify my installation to include git, fetch everything from github including the git folder etc: https://github.com/pypa/setuptools_scm/issues/278
I can do it (have done it before) but having a quick hack to say "ok this is cmd2 1.0" would be cool

tleonhardt:
@teto I'm not aware of any way to force the version in `setuptools_scm`.  We haven't given any thought to developers attempting to do `cmd2` work without `Git`.  Nor do we intend to give any thought to that.

docs/argument_processing.rst
```diff
@@ -275,6 +275,31 @@ Here's what it looks like::
...
+        """Populate an argparse Namespace with current settings"""
+        ns = argparse.Namespace()
+        ns.app_settings = self.settings
+        return ns
+
+To use this function with the argparse decorators, do the following::
+
+    @with_argparser(my_parser, ns_provider=settings_ns_provider)
+
+The Namespace is passed by the decorators to the ``argparse`` parsing functions which gives your custom code access
```
tleonhardt: Thanks for updating the docs

teto:
I was going to advise to use setuptools-git when I checked setuptools_scm to motivate the decision and finally found out about the   `SETUPTOOLS_SCM_PRETEND_VERSION="0.9.13";` variable that can skip the whole setuptools process.
I finally tried the PR and it works for me thanks !

PR #689, opened 2019-05-25
Persistent history for #669

### LINKED ISSUE #669: Consider loading persistent_history_file into cmd2 history in addition to readline history
@tleonhardt previously mentioned this following in https://github.com/python-cmd2/cmd2/issues/265 (when `readline` history was introduced)

> We would need to decide if this is done for `readline` history, `cmd2` history command history, or both. Perhaps there would be two separate options for making each of these persistent?

In the implementation of this feature (https://github.com/python-cmd2/cmd2/pull/270) only `readline` history support was implemented.

I think it would in fact make sense for this file to also affect the `cmd2` history command, as this is the behavior that other shells exhibit with their history files. It seems more consistent and expected to have both histories linked up.

kotfu:
Fixes #669.

We have a readline history and a native in-memory history. Previous to this PR, only readline history was persistent between invocations of a `cmd2` based app. Now both readline and native history are persistent. Plaintext history files are ignored, and do not cause read errors. All new history files written after this PR sill be in binary pickle format.

This includes several breaking changes noted in the changelog, including:

- Change persistent history format to pickle instead of plain text
- refactoring HistoryItem to not subclass string (it wouldn't unpickle)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,4 +28,4 @@ dmypy.json
 dmypy.sock
 
 # cmd2 history file used in hello_cmd2.py
-cmd2_history.txt
+cmd2_history.dat
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,9 +33,7 @@
     * Changed `Statement.pipe_to` to a string instead of a list
     * `preserve_quotes` is now a keyword-only argument in the argparse decorators
     * Refactored so that `cmd2.Cmd.cmdloop()` returns the `exit_code` instead of a call to `sys.exit()`
-        * It is now applicaiton developer's responsibility to treat the return value from `cmdloop()` accordingly
-      , and is in a binary format,
-      not a text format.
+      It is now application developer's responsibility to treat the return value from `cmdloop()` accordingly
     * Only valid commands are persistent in history between invocations of `cmd2` based apps. Previously
       all user input was persistent in history. If readline is installed, the history available with the up and
       down arrow keys (readline history) may not match that shown in the `history` command, because `history`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,13 +23,25 @@
     * Added support for custom Namespaces in the argparse decorators. See description of `ns_provider` argument
     for more information.
     * Transcript testing now sets the `exit_code` returned from `cmdloop` based on Success/Failure
-* Potentially breaking changes
+    * The history of entered commands previously was saved using the readline persistence mechanism,
+      and only persisted if you had readline installed. Now history is persisted independent of readline; user
+      input from previous invocations of `cmd2` based apps now shows in the `history` command.
+
+* Breaking changes
     * Replaced `unquote_redirection_tokens()` with `unquote_specific_tokens()`. This was to support the fix
       that allows terminators in alias and macro values.
     * Changed `Statement.pipe_to` to a string instead of a list
     * `preserve_quotes` is now a keyword-only argument in the argparse decorators
     * Refactored so that `cmd2.Cmd.cmdloop()` returns the `exit_code` instead of a call to `sys.exit()`
         * It is now applicaiton developer's responsibility to treat the return value from `cmdloop()` accordingly
+      , and is in a binary format,
+      not a text format.
+    * Only valid commands are persistent in history between invocations of `cmd2` based apps. Previously
+      all user input was persistent in history. If readline is installed, the history available with the up and
+      down arrow keys (readline history) may not match that shown in the `history` command, because `history`
+      only tracks valid input, while readline history captures all input.
+    * History is now persisted in a binary format, not plain text format. Previous history files are destroyed
+      on first launch of a `cmd2` based app of version 0.9.13 or higher.
 * **Python 3.4 EOL notice**
     * Python 3.4 reached its [end of life](https://www.python.org/dev/peps/pep-0429/) on March 18, 2019
     * This is the last release of `cmd2` which will support Python 3.4
@@ -38,7 +50,7 @@
 * Bug Fixes
     * Fixed a bug in how redirection and piping worked inside ``py`` or ``pyscript`` commands
     * Fixed bug in `async_alert` where it didn't account for prompts that contained newline characters
-    * Fixed path completion case when CWD is just a slash. Relative path matches were incorrectly prepended with a slash. 
+    * Fixed path completion case when CWD is just a slash. Relative path matches were incorrectly prepended with a slash.
 * Enhancements
     * Added ability to include command name placeholders in the message printed when trying to run a disabled command.
         * See docstring for ``disable_command()`` or ``disable_category()`` for more details.
@@ -60,7 +72,7 @@
         * ``_report_disabled_command_usage()`` - in all cases since this is called when a disabled command is run
     * Removed *** from beginning of error messages printed by `do_help()` and `default()`
     * Significantly refactored ``cmd.Cmd`` class so that all class attributes got converted to instance attributes, also:
-        * Added ``allow_redirection``, ``terminators``, ``multiline_commands``, and ``shortcuts`` as optional arguments 
+        * Added ``allow_redirection``, ``terminators``, ``multiline_commands``, and ``shortcuts`` as optional arguments
         to ``cmd.Cmd.__init__()`
         * A few instance attributes were moved inside ``StatementParser`` and properties were created for accessing them
     * ``self.pipe_proc`` is now called ``self.cur_pipe_proc_reader`` and is a ``ProcReader`` class.
@@ -98,7 +110,7 @@
     ``cmd2`` convention of setting ``self.matches_sorted`` to True before returning the results if you have already
     sorted the ``CompletionItem`` list. Otherwise it will be sorted using ``self.matches_sort_key``.
     * Removed support for bash completion since this feature had slow performance. Also it relied on
-    ``AutoCompleter`` which has since developed a dependency on ``cmd2`` methods. 
+    ``AutoCompleter`` which has since developed a dependency on ``cmd2`` methods.
     * Removed ability to call commands in ``pyscript`` as if they were functions (e.g. ``app.help()``) in favor
     of only supporting one ``pyscript`` interface. This simplifies future maintenance.
     * No longer supporting C-style comments. Hash (#) is the only valid comment marker.
@@ -118,7 +130,7 @@
     * Fixed bug where the ``set`` command was not tab completing from the current ``settable`` dictionary.
 * Enhancements
     * Changed edit command to use do_shell() instead of calling os.system()
-    
+
 ## 0.9.8 (February 06, 2019)
 * Bug Fixes
     * Fixed issue with echoing strings in StdSim. Because they were being sent to a binary buffer, line buffering
@@ -139,9 +151,9 @@
 * Deletions (potentially breaking changes)
     * Deleted ``Cmd.colorize()`` and ``Cmd._colorcodes`` which were deprecated in 0.9.5
     * Replaced ``dir_exe_only`` and  ``dir_only`` flags in ``path_complete`` with optional ``path_filter`` function
-    that is used to filter paths out of completion results. 
+    that is used to filter paths out of completion results.
     * ``perror()`` no longer prepends "ERROR: " to the error message being printed
-    
+
 ## 0.9.6 (October 13, 2018)
 * Bug Fixes
     * Fixed bug introduced in 0.9.5 caused by backing up and restoring `self.prompt` in `pseudo_raw_input`.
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -23,13 +23,27 @@
...
+      input from previous invocations of `cmd2` based apps now shows in the `history` command.
+
+* Breaking changes
     * Replaced `unquote_redirection_tokens()` with `unquote_specific_tokens()`. This was to support the fix
       that allows terminators in alias and macro values.
     * Changed `Statement.pipe_to` to a string instead of a list
     * `preserve_quotes` is now a keyword-only argument in the argparse decorators
     * Refactored so that `cmd2.Cmd.cmdloop()` returns the `exit_code` instead of a call to `sys.exit()`
         * It is now applicaiton developer's responsibility to treat the return value from `cmdloop()` accordingly
+      , and is in a binary format,
```
tleonhardt: These two lines appears to be a spurious cut-and-paste unrelated to the adjacent content.  Also, it looks like application is misspelled in the line above.

cmd2/history.py
```diff
@@ -11,21 +11,25 @@
...
+class HistoryItem():
     """Class used to represent one command in the History list"""
-    listformat = ' {:>4}  {}\n'
-    ex_listformat = ' {:>4}x {}\n'
+    _listformat = ' {:>4}  {}\n'
+    _ex_listformat = ' {:>4}x {}\n'
 
-    def __new__(cls, statement: Statement):
-        """Create a new instance of HistoryItem
+    # def __new__(cls, statement: Statement):
```
tleonhardt: Recommend deleting commented-out dead code

cmd2/history.py
```diff
@@ -222,5 +228,19 @@ def regex_search(self, regex: str) -> List[HistoryItem]:
...
+        :param max_length: the maximum length of the history, if negative, all history
+                           items will be deleted
+        :return: nothing
+        """
+        if max_length <= 0:
+            # remove all history
+            del self[:]
+        elif len(self) > max_length:
+            last_element = len(self) - max_length
+            del self[0:last_element]
```
tleonhardt: Would it be useful to add `__str__` and/or `__repr__` dunder methods to the `HistoryItem` class?

tests/test_history.py
```diff
@@ -383,62 +458,64 @@ def hist_file():
...
-def test_existing_history_file(hist_file, capsys):
-    import atexit
-    import readline
-
-    # Create the history file before making cmd2 app
-    with open(hist_file, 'w'):
-        pass
+def test_bad_history_file_path(capsys, request):
+    # Use a directory path as the history file
+    test_dir = os.path.dirname(request.module.__file__)
```
tleonhardt: This won't work on all of our CI systems due to permission issues - specifically it is failing on AppVeyor.  I recommend changing it to use a tempdir/tempfile since that should be writeable on all of our CI systems.

CHANGELOG.md
kotfu: Yup, big oopsie on my part. Fixed.

cmd2/history.py
kotfu: Deleted. Good catch.

cmd2/history.py
kotfu: Very useful. Done.

tests/test_history.py
kotfu: Fixed. Now using `tempfile.TemporaryDirectory()` as a context manager. around the test code.

cmd2/history.py
```diff
@@ -7,29 +7,31 @@
 
 from typing import List, Union
 
+import attr
+
 from . import utils
 from .parsing import Statement
 
-
+@attr.s(frozen=True)
```
tleonhardt: Good idea to make these immutable.

tests/test_history.py
```diff
@@ -383,62 +458,61 @@ def hist_file():
...
-    os.remove(hist_file)
-
 def test_bad_history_file_path(capsys, request):
-    # Use a directory path as the history file
-    test_dir = os.path.dirname(request.module.__file__)
+    with tempfile.TemporaryDirectory() as test_dir:
+        # Create a new cmd2 app
+        cmd2.Cmd(persistent_history_file=test_dir)
+        _, err = capsys.readouterr()
+        assert 'is a directory' in err
```
tleonhardt: This test is still failing on Windows/AppVeyor.  It looks like for that OS, it is not getting an `IsADirectoryError`, but an `OSError` instead.

One thing I've been thinking recently is that since we are now Python 3.4+, we should consider using [Pathlib](https://docs.python.org/3/library/pathlib.html) for paths and to do things like testing to make sure the path is a file.

kotfu: Yeah, I saw that it's still broken too. I'm already catching the `IsADirectoryError` exception, which is what shows the error message that appveyor is barfing on. This test was written to make sure that we generated an error if someone passed a directory as the persistent history file. For some reason, AppVeyor is throwing a permission error, not the `IsADirectoryError`. I'm working on it....

kotfu: OK, I'm stumped. `tempfile.TemporaryDirectory()` is supposed to create a directory that's readable and writable by the current user. It sure seems like it did, because I can create a new directory `somedir` inside of it. But when we go to `open(hist_file, 'rb')` either one of those directories, we get a permission error, not an `IsADirectoryError`.

tleonhardt: I'm spinning up my Windows VM to try to investigate.  This is weird.

tleonhardt: So on Windows when you try to open a directory in Python 3, you get an OSError with a message about permission denied.

kotfu: Got it. So instead of relying on the `IsADirectoryError` exception to tell me we are trying to open a directory, I need to manually check whether it's a directory using `os.is_dir()`. I'll try that.

tleonhardt: Yup you got it.  But `pathlib` has some advantages over `os`:  https://treyhunner.com/2018/12/why-you-should-be-using-pathlib/

We might want to experiment with using that because we can do things like:
```Python3
import pathlib
hist_file = pathlib.Path(hist_file).expanduser().resolve()
if hist_file.is_dir():
    # Handle case where the specified history file is actually a directory

if not hist_file.parent.is_dir()
    # Handle the problem that the directory the history file should be in isn't a directory
```

For Python 3.6 and newer you can use a `Path` directly in `open()`.  Before that you cannot, but you can cast to a `str` or use the `read_text` or `read_bytes` methods.

kotfu: It worked with the manual check if it's a directory, but I'm liking the `pathlib` approach, I didn't even know that thing existed. I'll switch it over from `os.path` to `path lib`.

cmd2/cmd2.py
```diff
@@ -3469,15 +3469,18 @@ def _initialize_history(self, hist_file):
 
         # first we try and unpickle the history file
         history = History()
+        # on Windows, trying to open a directory throws a permission
+        # error, not a `IsADirectoryError`. So we'll check it ourselves.
+        if os.path.isdir(hist_file):
```
tleonhardt: This seems like a reasonable solution.

tleonhardt:
Changes look good.  I haven't had a chance to manually test yet, but plan to do so soon.

If we could get test coverage up of new code that would be a good thing.

cmd2/cmd2.py
kotfu: Pathlib on python 3.4 doesn't have `expanduser()`. Switching back to `os.path`

tleonhardt: Ah good call.  I thought there was a reason I was waiting until we deprecated Python 3.4 before I started adding `pathlib` usage, but I couldn't recall what it was.  Thats it!

cmd2/history.py
```diff
@@ -7,25 +7,32 @@
...
+    idx = attr.ib(default=None, validator=attr.validators.instance_of(int))
 
-        We must override __new__ because we are subclassing `str` which is
-        immutable and takes a different number of arguments as Statement.
-        """
-        hi = super().__new__(cls, statement.raw)
-        hi.statement = statement
-        hi.idx = None
-        return hi
+    def __str__(self):
```
tleonhardt: Recommend adding two simple unit tests for this method, one for the `if` and another for the `else` case.

cmd2/history.py
```diff
@@ -7,25 +7,29 @@
...
+        """A convenient human readable representation of the history item"""
+        return self.statement.raw
```
tleonhardt: @kotfu I simplified the implementation for this since the attrs validator assures we have a Statement.  Please make sure this and the unit test I added seems OK to you.

kotfu: Looks good. I'm working to get `history.py` and `do_history()` to 100% test coverage before we merge this.

tleonhardt: This should put `history.py` at 100% coverage.  But there are a number of little edge cases in `do_history` remaining.  Good luck!

cmd2/cmd2.py
```diff
@@ -402,13 +403,15 @@ def __init__(self, completekey: str = 'tab', stdin=None, stdout=None, persistent
         self.hidden_commands = ['eof', 'eos', '_relative_load']
 
         # Commands to exclude from the history command
+        # initialize history
+        self.persistent_history_length = persistent_history_length
```
tleonhardt: Documentation for `persistent_history_file` and `persistent_history_length` should be updated in the `cmd2.Cmd.__init__` docstring.

Also, documentation on "Searchable command history" should be mildly updated in the Sphinx docs.

tleonhardt:
An alternative to the idea of the boolean flag in `cmd2.Cmd.__init__()` discussed above would be to have the **history** command only show history from the current session by default but to add a flag to the **history** command, perhaps `-a`/`--all`, which then shows the complete history including that persisted from previous sessions.  I believe this ideas was previously discussed in #669

tleonhardt:
@kotfu and/or @kmvanbrunt 
Please review the changes I made for designing and implementing the new `-a`/`--all` flag for the **history** command which enables display of all commands including those persisted from previous sessions (which are not displayed by default).

The `History` class now keeps track of the starting index for the current session and the `span()`, `str_search()`, and `regex_search()` methods all take an optional `include_persisted` boolean argument which specifies whether or history persisted from previous sessions should be included.

If a starting index is specified in the argument to the `span()` method, then the value of `include_persisted` is ignored and the user gets what they ask for since they are being specific.

.gitignore
```diff
@@ -28,4 +28,4 @@ dmypy.json
 dmypy.sock
 
 # cmd2 history file used in hello_cmd2.py
-cmd2_history.txt
+cmd2_history.dat
```
tleonhardt: Maybe ignore both the old and new name just to be safe?

cmd2/cmd2.py
```diff
@@ -3397,10 +3371,11 @@ def do_history(self, args: argparse.Namespace) -> None:
             # Clear command and readline history
             self.history.clear()
 
+            if self.persistent_history_file:
```
tleonhardt: Recent changes look fine

teto:
May I inquire why the change to binary fromat for something that is supposed to be human readable ? It's great to be able to just open an editor and look at the history or even to prefill an history. 
sounds like a regression to me.

teto:
why do we need 2 histories ? `readline history and a native in-memory history.`.
Readline is a widely available dependency and this seems to add lots of complexity for both the user and the cmd2 developers ? I guess I am missing something xD

tleonhardt:
@teto The change to the format of the persistent history file is because we had a user request, #669, to persist the `cmd2` **history** command history in addition to the `readline` history.

For this, the core `cmd2` developers had a lively debate in regards to the if we should implement this and if so how we could do it in a way which was both convenient for end users and maintainable for core developers.  In the end, we chose to do it using a single binary Pickle file from which we restore both the `cmd2` history and `readline` history.   I think most end users want persistent history and they don't care what format it gets persisted in and whether or not that is human readable.

In regards to why we want two histories, there are several reasons.  First off `cmd2` is a cross-platform framework which works the same way on Windows, macOS, and Linux.   As a primarily Mac and Linux user I am quite familiar with and love `readline` history along with the <Ctrl>+r and arrow-key capabilities built into it.  However, few Windows users are familiar with `readline` and using it isn't very discoverable for them.  A built-in **history** command on the other hand is very discoverable for all users.  Secondly, the built-in `cmd2` ** history** command has capabilities that go significantly beyond `readline` history, in part because it uses knowledge of internal `cmd2` representations.

If your `cmd2` development is negatively impacted by this change, can you please explain in detail precisely how and perhaps we can do something to help.

teto:
I don't really mind the 2 histories. as a user, it mostly confuses me but that's fine. Hadn't thought of windows users I admit :/ I was just a bit surprised by the `+563 −238` diff which makes me worried in terms of maintenance etc, I don't want cmd2 to become so complex that no new maintainer want to look at the code, it might the only program with 2 histories I know.

> I think most end users want persistent history and they don't care what format it gets persisted in and whether or not that is human readable.

if people don't care what's the motivation for changing (btw I do care :p ) ? Any advantage of binary serialisation doesn't make sense for this usecase. Neither bash or zsh serialise the history. I like to have a look at it. What if I want to sync it on git ? I am not familiar with pickling but I fear it's not portable.  What if the computer unexpectedly shuts down ? with zsh I can just open the history file and remove the last line to fix it but I don't think that will be possible with pickling.
My cmd2 commands are rather long and I rely on history a lot.

teto:
Looking at the changelog, seems like the history items has grown to some big object, which would explain why one wouId want to use pickle. I will regret the txt based format but not enough to nag the nice maintainers :p thanks for your work.

kotfu:
A couple of thoughts about concerns raised by @teto.

- History has always been a big (for this application) object; with this change we save the object instead of it disappearing when the app exits
- picking works where python works, but is not portable to other tools
- We have the same risk of data loss on unexpected shutdown that we did before we implemented this change. We have always accumulated the history in memory and only wrote it out on a clean exit.
- There is a robust text-based interface to history via the built-in **history** command. You can save historical commands to a text file, search them, etc. It may meet many of your needs, but won't let you sync history with git.
- If you want to persist history to a plain text file, you can use the python `atexit` module to register your own function which rips through the history array and writes it to the text file of your choosing

Issue #692, opened 2019-06-04
`None` prompt exits cmdloop


deargle:
This is a silly edge-case, but if the prompt is None, the cmdloop silently exits. A prompt of `''` does not exit.

I don't want to have a `None` prompt, but I accidentally had one while debugging something else and was scratching my head.
``` python
from cmd2 import Cmd
class MyApp(Cmd, object):

    def __init__(self):
        self.prompt = None
        Cmd.__init__(self)

app = MyApp()
app.cmdloop()
```

Python 2.7, cmd2 v0.8.9

kmvanbrunt:
**cmd2** expects `prompt` to always be a string. The current version of **cmd2** does a bit more processing of `prompt` than v0.8.9, and therefore altering its type will certainly break things. This is to be expected.

If a user actually wants a blank prompt, they can do the following:

```Python
self.prompt = ''
```

tleonhardt:
I did some manual testing and at least with my setup encountered what I consider a legitimate bug that we need to deal with.

I tested on macOS Mojave 10.14.5 using Python 3.7.3 and latest `cmd2` *master* branch from GitHub with Bash shell and iTerm2.  

I tested two cases:
1. prompt set to empty string
    * Caused an infinite loop that got stuck on startup
1. prompt set to None
    * Caused a crash that caused `cmd2` to exit

I think we should add a check somewhere along the lines of the following pseudocode:
```Python3
if not self.prompt or not isinstance(self.prompt, str):
    self.prompt = DEFAULT_PROMPT
```

tleonhardt:
Uh, I was an idiot.  There was never an infinite loop when I set the prompt to an empty string.  I just tested it with an example without an intro banner so it looked like nothing happened.  But then I tried issuing a command like **help** and saw that `cmd2` was running.

deargle:
Yeah similar to popping a non-tty shell through metasploit or whatever.

kmvanbrunt:
The reason setting prompt to `None` fails silently is because `_cmdloop()` is returning within its `finally`. This suppresses the exception caused by the bad prompt. My **script_refactor** branch already fixed the issue in `_cmdloop()` and therefore shows a stack trace. I will close this ticket when that branch merges.

kmvanbrunt:
Closed by #696

Issue #688, opened 2019-05-23
flag to exit cmd2/stop the loop on command failure


teto:
Follow up of https://github.com/python-cmd2/cmd2/issues/684.

This is some - late (sorry) - insight on https://github.com/python-cmd2/cmd2/pull/685. I've been trying to leverage cmd2 to help me with testing so here is my process: I have 2 scripts.
generate_transcripts et run_tests. run_tests fails as soon as a test fail, perfect ! 
Similarly, I would like `generate_transcripts` to exit when it fails to generate a transcript.

```
# Make it fail if any call fails
set -o errexit
function gen_transcript()
{
	suffix="$1"
	mptcpanalyzer "load tests/script_${suffix}.txt -t tests/trans_${suffix}.txt" "quit"
}

gen_transcript "tcp"
gen_transcript "mptcp"
...
```
E.g., I would like to have my program to stop with a non 0 exit code when "load script.txt -t text.txt" fails.
While the PR https://github.com/python-cmd2/cmd2/pull/685 targeted one command, I feel like there is the possibility to abstract the mechanism to apply it more generally. Like a flag `--exit-on-error`.
I don't know about the API, if it should be added per-command:
```
    @with_argparser(parser)
    @exit_on_error
    def do_mycommand
```
or as a global flag:
`my_cmd2_program --exit-on-error "load test.txt -t test.txt"`
What do you think ?

tleonhardt:
@kmvanbrunt I think this request relates to some of the changes you were thinking of making.

kmvanbrunt:
This should be fixed by #687. Neither `_generate_transcript()` or `_test_transcript()` stop when `onecmd_plus_hooks()` says to.

teto:
@kmvanbrunt how does it translate as a cmd2 API user ? does generating a transcript always exit cmd2 ? or is it an option ? should I override onecmd_plus_hooks ?

kmvanbrunt:
Generating a transcript will only exit the command loop if one of the commands it runs returns **True** for its **Stop** value. For instance, running **quit** while generating a transcript will stop both transcript generation and **cmd2's** command loop.

kmvanbrunt:
Generating a transcript will stop early if any of the commands it is running returns **True** for **stop**. So as long as the command that fails sets **exit_code** and returns **True**, then you should get what you need when the app exits. 

This line of code you use to run tests should work with no changes:
```Python
[...]

PR #696, opened 2019-06-10
Script refactor

### LINKED ISSUE #687: Handle return code of onecmd_plus_hooks
There are a few cases in **cmd2** where `onecmd_plus_hooks()` returns **True**, which means stop the command loop, but stop is not passed up. Evaluate these cases and determine if stop should be handled.

kmvanbrunt:
- Text scripts now run immediately instead of adding their commands to `cmdqueue`. This allows easy capture of the entire script's output.
- Transcript generation/runs now stops if any command it runs returns True for stop.
- Added member to `CommandResult` called `stop` which is the return value of `onecmd_plus_hooks` after it runs the given command line.
- Removed internally used `eos` command that was used to keep track of when a text script's commands ended
- Removed `cmd2` member called `_STOP_AND_EXIT` since it was just a boolean value that should always be True
- Removed `cmd2` member called `_should_quit` since `PyscriptBridge` now handles this logic
- Fixed issue where `cmdloop()` suppressed exceptions by returning within its `finally` code

Fixes #687, #688, #692


cmd2/cmd2.py
```diff
@@ -400,13 +400,13 @@ def __init__(self, completekey: str = 'tab', stdin=None, stdout=None, persistent
                          'timing': 'Report execution times'}
 
         # Commands to exclude from the help menu and tab completion
-        self.hidden_commands = ['eof', 'eos', '_relative_load']
+        self.hidden_commands = ['eof', '_relative_load']
```
tleonhardt: Nice, I always felt that the **eos** command was an ugly but functional hack that I created.

cmd2/cmd2.py
```diff
@@ -1810,47 +1804,25 @@ def _run_cmdfinalization_hooks(self, stop: bool, statement: Optional[Statement])
...
-        NOTE: You may need this method even if you only have one command. If
-        that command is a load, then you will need this command to fully process
-        all the subsequent commands that are loaded from the script file. This
-        is an improvement over onecmd_plus_hooks, which expects to be used
-        inside of a command loop which does the processing of loaded commands.
+            if self.echo:
+                self.poutput('{}{}'.format(self.prompt, line))
 
-        Example: cmd_obj.runcmds_plus_hooks(['load myscript.txt'])
+            if self.onecmd_plus_hooks(line):
```
tleonhardt: Ok, so now we just run all of the commands in a loop instead of appending them to the cmdqueue.  That seems reasonable.  I'm fearful that there could be some really obscure edge case where this would behave very differently, but I can't think of any right now.

cmd2/cmd2.py
```diff
@@ -2271,15 +2241,13 @@ def _cmdloop(self) -> bool:
...
             while not stop:
                 if self.cmdqueue:
-                    # Run command out of cmdqueue if nonempty (populated by load command or commands at invocation)
-                    line = self.cmdqueue.pop(0)
-
-                    if self.echo and line != 'eos':
-                        self.poutput('{}{}'.format(self.prompt, line))
+                    # Run commands out of cmdqueue if nonempty (populated by commands at invocation)
+                    stop = self.runcmds_plus_hooks(self.cmdqueue)
+                    self.cmdqueue.clear()
```
tleonhardt: I'm not sure I like this implementation relative to the old one where we just pop an element off the queue. 

What if the act of running commands in the queue actually adds commands to the queue?  Or what if there is a background thread running which adds commands to the queue?  This would then immediately delete those commands.  

I strongly recommend you switch this back to an implementation that pops a single command off the queue.

cmd2/cmd2.py
```diff
@@ -3592,12 +3569,14 @@ def _generate_transcript(self, history: List[Union[HistoryItem, str]], transcrip
...
-            if len(history) > 1:
+            if commands_run > 1:
                 plural = 'commands and their outputs'
             else:
                 plural = 'command and its output'
             msg = '{} {} saved to transcript file {!r}'
-            self.pfeedback(msg.format(len(history), plural, transcript_file))
+            self.pfeedback(msg.format(commands_run, plural, transcript_file))
+
+        return stop
```
tleonhardt: I'm not sure if I'd want generating a transcript to quit the overall cmd2 application.  But I'm also not sure I wouldn't.  I guess its a judgement call.

cmd2/cmd2.py
kotfu: Agree with @tleonhardt, we should go back to the implementation that pops a single command off the queue when we need a new command. As coded in this PR, the behavior of running the script could be different that what happened when it first happened. For example, if the third command in the script adds a command to `cmdqueue`, that command won't get executed when we run the script.

cmd2/cmd2.py
kotfu: I feel pretty strongly that generating a transcript shouldn't stop the application. I think it's the wrong behavior, and it's a change from current behavior.

teto: Just as a quick input: when run in interactive mode, I don't want it to stop the application. Yet I have a bunch of scripts that only generates transcript and if they fail, I would like to know about it (having cmd2 return an error).

kmvanbrunt:
Uh oh, I'm getting some push back here on 2 issues. Here is my thinking on both.

**Command Queue Issue**
I agree the code to run commands in `cmdqueue` as written in this PR is wrong. My intention was for **cmd2** to only use `cmdqueue` for startup commands. I should have actually moved the code outside of the `while` loop.

The problem is that **cmd2** has at least 6 places that could be considered a command loop.
1. _cmdloop()
2. history -r
3. load
4. pyscript
5. transcript running
6. transcript generation

Only `_cmdloop()` checks the contents of `cmdqueue` before running the next command. For this reason, running commands from the command line will behave differently in cases where a command appends to the command queue.

For instance, suppose I have a command called **save** that adds a command called **print_results** to the queue.

Then at the command line, the user runs these commands:
**save**
**help**

My history would be:
1 **save**
2 **print_results**
3 **help**

If I then do, `history -r`, commands will run this way:
**save**
**print_results**
**help**
**print_results**

`cmdqueue` is a remnant of **cmd**, which only runs commands in one place. We run commands from the main loop, history, scripts, and transcripts. Therefore I propose eliminating support of `cmdqueue` and instead create a member list called `_startup_commands`. These commands will run before the `while` loop in `_cmdloop()`.

**Transcript Generation Issue**
This is a tricky one. The behavior before this PR was wrong in that the value of `stop` returned by `onecmd_plus_hooks` was never read and therefore transcript generation kept going.

My PR added code to stop transcript generation when any command returned **True**. That value gets passed up to the command loop, which then stops. Technically the whole application doesn't stop. Just the command loop does.

Since transcript generation is running actual commands, then behavior should be identical to running these same commands from the CLI or scripts. I agree that it seems weird to terminate the command loop after transcript generation, but it's also weird to put quit in a transcript.

In applications that I've written, commands that return **True** for **stop** do so after transitioning the app to a shut down state (e.g. Network and database connections closed). In that case, I don't want the command loop continuing. I told the app to quit, so it should quit.

I'm not sure I've altered existing behavior since we never handled **stop** in the first place. What I'm trying to do is have consistent behavior across all ways of running commands. Basically, if a command tells the command loop to stop, then it should in all cases, regardless of how it was run.

I wrote this in issue #688. Perhaps I didn't explain well enough.

> Generating a transcript will only exit the command loop if one of the commands it runs returns **True** for its **Stop** value. For instance, running **quit** while generating a transcript will stop both transcript generation and **cmd2's** command loop..

kmvanbrunt:
@kotfu @tleonhardt 
Since adding `stop` handling to transcript generation is new behavior, I took my best guess. After reading my reasons, if you still disagree with stopping the whole command loop, let me know what the desired behavior should be.

1. Should I just terminate the transcript generation loop?
2. Should I also print a message to the user? Right now it already prints a message using `pfeedback()` saying how many command ran. If it stops early, I could let them know which command caused a stop.

tleonhardt:
The current unit test failures are pretty weird and I don't understand them, but I have been able to replicate them locally.

If I just run `py.test` without any arguments, all tests pass.

However, if I run `py.test` with any arguments, such as `py.test -v`, then the same 6 tests fail and they do so because the arguments to `py.test` get passed as commands at invocation during these tests.

tleonhardt:
I made a partial reversion of the last commit to get the unit tests passing again until such as time as we can figure out WTF was going wrong.

cmd2/cmd2.py
```diff
@@ -466,11 +462,20 @@ def __init__(self, completekey: str = 'tab', stdin=None, stdout=None, persistent
         # If this string is non-empty, then this warning message will print if a broken pipe error occurs while printing
         self.broken_pipe_warning = ''
 
+        # Commands that will run at the beginning of the command loop
+        self._startup_commands = []
```
tleonhardt: Changing to the use of the `self._startup_commands` list instead of using `self.cmdqueue` inherited from `cmd` solves the "Command queue issue" alluded to in the comments.

Overall, I think we should do a better job in the Sphinx docs of calling out the elements in `cmd` which are not used or supported in `cmd2`.  We have a "Note" for this in our [Overview](https://cmd2.readthedocs.io/en/latest/overview.html).  But I think it is long overdue that we expand upon that.

cmd2/cmd2.py
tleonhardt: I do not think generating a transcript should ever stop the application, a stop should just stop the transcript generation loop.  If it stops early, it would probably also be a good idea to print a warning message to the user letting them know which command caused it to stop early.

kmvanbrunt: Should this warning message be printed with `perror` or `pfeedback`?

cmd2/cmd2.py
```diff
@@ -466,11 +462,33 @@ def __init__(self, completekey: str = 'tab', stdin=None, stdout=None, persistent
...
+
         # If a startup script is provided, then add it in the queue to load
         if startup_script is not None:
             startup_script = os.path.abspath(os.path.expanduser(startup_script))
             if os.path.exists(startup_script) and os.path.getsize(startup_script) > 0:
-                self.cmdqueue.append("load '{}'".format(startup_script))
+                self._startup_commands.append("load '{}'".format(startup_script))
+
+        # Transcript files to run instead of interactive command loop
+        self._transcript_files = None
```
tleonhardt: This should be set to `transcript_files` and the `elif` below should be deleted.  The intention is that transcript files can be provided in the call to `__init__` regardless of whether or not `allow_cli_args` is true, but if `allow_cli_args` is true **and** the `-t / --test` flag was provided, then that would override whatever was passed to `__init__`

cmd2/cmd2.py
kmvanbrunt: Transcript generation now just stops and prints a message if all commands have not run

cmd2/constants.py
```diff
@@ -20,7 +21,13 @@
...
 LINE_FEED = '\n'
 
-# values for colors setting
+# Values for colors setting
 COLORS_NEVER = 'Never'
 COLORS_TERMINAL = 'Terminal'
 COLORS_ALWAYS = 'Always'
+
+# Text colors
+SUCCESS_COLOR = Fore.LIGHTGREEN_EX
```
tleonhardt: Good call creating constants for logical colors

tleonhardt:
@kotfu If you get a chance to look at this, there are a lot of changes.  I think @kmvanbrunt addressed all of the concerns we both had, but I'd appreciate any further feedback you might have.

The major fundamental change is that `self.cmdqueue` is now completely unused in `cmd2` and its use has been replaced by `self._startup_commands`.  But `self._startup_commands` is is only used for supporting startup scripts and commands at invocation and is completely unused during the main REPL.  A side affect is that `load` no longer queues commands, but runs them directly.

I can't think of any negative unintended consequences.  But this is a major change in implementation, so I'd appreciate another brain thinking about it.

tleonhardt:
@teto Can you verify that this PR is getting you most of what you want in terms of the transcript testing and generation changes you were looking for?

PR #703, opened 2019-06-15
Minimize public API of cmd2.Cmd class

### LINKED ISSUE #677: Minimize public API
In preparation for the 1.0.0 release and to align with [Semantic Versioning](https://semver.org/), I think it would be a good idea if we look at the overall `cmd2` public API presented to end users and either explicitly make *internal* functions/methods such by beginning their name with an underscore and/or to do a better job clarifying what we consider the public API to be. 

One way or the other, I think we want to clearly convey the parts of `cmd2` we intend to remain stable within major releases and the parts we don't make any promises about.

### LINKED ISSUE #683: `shortcuts` should not be a publicly accessible attribute
We have some inconsistencies in our api and our documentation. In 0.9.12 we added ``shortcuts`` as an optional argument to `__init__()`, and changed the initialization code associated with these shortcuts. This created the following issues:

- the changelog said nothing about disabling the `.shortcuts` attribute. Prior to 0.9.12 you could modify that attribute to change the shortcuts. After 0.9.12, the only way to change the shortcuts is to pass them to `__init__()`
- What's the point of the `shortcuts` attribute on a `cmd` object if the caller has to pass it in, and it's read only? When `cmd2` code needs it, can get it from the `statement_parser`, and code that instantiates `cmd2.Cmd()` or a subclass thereof already know it because they passed it in.
- the docstring for the __init__ method says you should pass shortcuts as a dictionary, and then the initialization code processes those and stores a tuple in the `shortcuts` attribute (which has a property to access it). It seems inconsistent to pass a dictionary and access a tuple.
- The DEFAULT_SHORTCUTS variable should really be in `constants.py`, not `cmd2.py`

### LINKED ISSUE #691: Consider refactoring unit tests which are too dependent upon specific implementation
We have a small set of unit tests which are brittle because they break every time we add, delete, or rename any of the built-in commands (i.e. `do_*` methods).  These aren't difficult to update, but it still takes time and requires developer effort which could be better spent elsewhere.

We should consider intelligently refactoring these tests so they focus more on testing that the application *behaves* in the correct way without testing the specific implementation.

If we do this correctly, then hard-coded constants like `BASE_HELP` and `BASE_HELP_VERBOSE` in [conftest.py](https://github.com/python-cmd2/cmd2/blob/master/tests/conftest.py) which explicitly depend on matching the exact format of output should go away and be replaced with code that is more dynamic, flexible, and robust to change.

### LINKED ISSUE #695: Consider refactoring transcript testing code to provide better info to user
The information displayed to the user by the output of running transcript test(s) leaves a lot to be desired, especially compared to a good testing framework.  

First, it would probably be a good idea to display the version of Python and cmd2 this is being run with.

Next, it should show how many test files it was invoked with.

Then, it should show which transcript file is being executed along with showing pass/fail status on a per-file basis and should fail fast (i.e. quit as soon as there is a failure).

Finally it should show an overall pass/fail status and perhaps an overall time to execute.

tleonhardt:
Numerous changes have been made in order to minimize the public API of the `cmd2.Cmd` class in preparation for moving to [Semantic Versioning](https://semver.org) starting with the 1.0 release of `cmd2`.

I also refactored some of the unit tests to make them less dependent on specific implementation and more resilient to adding, deleting, or renaming commands.

Additionally, I started to refactor how information is displayed to the user when running in transcript testing mode.

Closes #677 
Closes #683 
Closes #691 
Closes #695 

**TODO:**
* [x] @kmvanbrunt review tab-completion related member variables around cmd2.py lines 482 - 517
* [x] @kmvanbrunt review tab-completion related methods around cmd2.py lines 703 - 1573
* [ ] @kotfu make sure the changes broadly match your expectations in terms of the 3 categories you outlined in #677

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 0.9.14 (TBD, 2019)
 * Enhancements
     * Added support for and testing with Python 3.8, starting with 3.8 beta
+    * Improved information displayed during transcript testing
 * Breaking Changes
     * Python 3.4 reached its [end of life](https://www.python.org/dev/peps/pep-0429/) on March 18, 2019 and is no longer supported by `cmd2`
         * If you need to use Python 3.4, you should pin your requirements to use `cmd2` 0.9.13
@@ -9,9 +10,9 @@
         * We make no API stability guarantees about these internal functions
 * **Renamed Commands Notice**
     * The following commands have been renamed. The old names will be supported until the next release. 
-        * load --> run_script
-        * _relative_load --> _relative_run_script
-        * pyscript --> run_pyscript
+        * `load` --> `run_script`
+        * `_relative_load` --> `_relative_run_script`
+        * `pyscript` --> `run_pyscript`
 
 ## 0.9.13 (June 14, 2019)
 * Bug Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,12 @@
-## 0.9.14 (TBD, 2019
+## 0.9.14 (TBD, 2019)
 * Enhancements
     * Added support for and testing with Python 3.8, starting with 3.8 beta
 * Breaking Changes
     * Python 3.4 reached its [end of life](https://www.python.org/dev/peps/pep-0429/) on March 18, 2019 and is no longer supported by `cmd2`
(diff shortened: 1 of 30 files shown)


tleonhardt:
I'm going to go ahead and merge this.  If anyone has further feedback, we can create another PR.

Issue #698, opened 2019-06-14
Revisit color support for poutput, perror, etc.


tleonhardt:
Currently the signature for `poutput` and `perror` are as follows:
```Python3
def poutput(self, msg: Any, end: str = '\n', color: str = '') -> None:
    pass

def perror(self, err: Union[str, Exception], traceback_war: bool = True, err_color: str = Fore.LIGHTRED_EX,
               war_color: str = Fore.LIGHTYELLOW_EX) -> None:
    pass
```

Where `color`, `err_color` and `war_color` are optional ANSI escape codes to style/format the output with and if present, we reset the foreground color after the output.  However, we never considered that someone may set the background color or other effects and we do not reset those.

We should probably refactor these methods to take separate foreground and background styles such as:
```Python3
def poutput(self, msg: Any, end: str = '\n', fg: str = '', bg: str = '') -> None:
    pass
```
And the foreground and background should be reset respectively at the end depending on what is provided.

We may also want to consider making these more user friendly where the user can pass in an English name for a color such as `'green'` and we have two dictionaries maintained inside the `cmd2.Cmd` instance called something like `self.fg_color` and `self.bg_color` which map a string to an ANSI escape sequence.  These dictionaries could be pre-populated with default styles for common colors as well as logical situations like `warning`, `error`, `success`, etc.  But `cmd2` application developers would be able to override these default styles for any given color.

xNinjaKittyx:
I just quickly hacked up something:
```
    def poutput(self, msg: Any, end: str = '\n', fg: str = '', bg: str = '') -> None:
        if msg is not None and msg != '':
            try:
                values = []
                try:
                    if fg:
                        values.append(getattr(Fore, fg.upper()))
                    if bg:
                        values.append(getattr(Back, bg.upper()))
                except AttributeError:
                     raise ValueError('Color does not exist.')
                values.append(msg)
                values.append(Style.RESET_ALL)

                if end:
                    values.append(end)

                self.decolorized_write(self.stdout, ''.join(values))
            except BrokenPipeError:
                ...
```

Test code:
```
import cmd2


class MyApp(cmd2.Cmd):
    def do_foo(self, args):
        self.poutput('something', fg='blue', bg='red')


MyApp().cmdloop()
```

Just one of many ways possible... `getattr` feels kind of "hacky". However, having 2 dictionaries that track Foreground + Background seems really awkward...

[...]

PR #704, opened 2019-06-25
#698 - Revisiting Color Support

xNinjaKittyx:
Addresses #698 

Changes:
- `poutput` and `perror` no longer use `color` kwarg, which has been split into `fg` and `bg`.
- `fg` and `bg` use human names of color like `blue` or `red`.
- `perror` is split into 2 functions `perror` and `pexcept`. `pexcept` will behave identical to the previous `perror`, but `perror` will only print to stderr with no stacktraces or Exception messages.


tleonhardt:
Recommend updating test_utils.py to add new unit tests for style_message() so that new utility function has 100% code coverage.

tleonhardt:
Thanks for the PR!  I'll take a look at it tonight.

kmvanbrunt:
I've made a few changes and doc updates. Here are some remaining questions I have:
1. Should `style_message` have an `end` argument? That's not related to style and adding a newline by default when a developer wanted to color a string seems unexpected.
2. I'm uncomfortable going to a string approach for specifying colors. Strings lead to typos that static analyzers won't catch.

tleonhardt:
@xNinjaKittyx By the way, your PR is excellent, so please don't take any offense at other developers collaborating to make it even better.  We might have a moderately peculiar culture here on this project, where it is extremely common for developers to collaborate and work to improve each other's PRs.

tleonhardt:
I like this PR as-is, but I also like the direction @kmvanbrunt is moving in to address the suggestion by @kotfu to decouple the functions for outputting text from the one for styling it and think that is probably a more flexible and maintainable improvement albeit at the cost of a small amount of convenience. 

Some things I think we will need to address once the changes stabilize and before merging into *master* include:

# TODO
* [ ] Update the CHANGELOG
* [ ] Make sure Sphinx docs are still accurate, particularly for section on [output functions](https://cmd2.readthedocs.io/en/latest/unfreefeatures.html?highlight=poutput#poutput-pfeedback-perror-ppaged)
* [ ] Move styling function and associated color dictionaries to a new file called something like **color.py** and add two-stage functions for retrieving the color text which first look in the dictionaries, but then attempt to fallback to getting the color directly from `colorama` using the `getattr` "hack" suggested by @xNinjaKittyx

kmvanbrunt:
@xNinjaKittyx 
Sorry for basically stealing this PR. Issue #698 was based on a conversation I had with @tleonhardt, so I already had some thoughts on the direction we wanted to go.

Then I looked at the [`style()`](https://github.com/pallets/click/blob/a936b99212c69a36a2cc2828252b2c4280c0c88c/click/termui.py#L372) function you posted from **click** and I got very inspired. Since a lot of code needed to be changed, I just went ahead and did it to have it done quickly. Time was an important factor because @tleonhardt and I are doing a **cmd2** talk at PyOhio in a month. I wanted to stabilize the code and give us enough time to make the presentation.

Thanks for your help and don't be discouraged from contributing in the future. I don't usually steal :)

xNinjaKittyx:
@kmvanbrunt It's all good. I'm glad that inspired you :D.
I've never heard about PyOhio, but if they post the content online, I'll make sure to give it a watch.

I use this library quite a bit at work, so I thought I'd contribute when I see something I can do. I'll be on the lookout.

tleonhardt:
@kotfu If you get a chance to look at this, your feedback would be appreciated. 

@xNinjaKittyx Can you please review that changes that @kmvanbrunt and I have made to your PR to make sure they all seem reasonable to you?

xNinjaKittyx:
I think there's a few more breaking changes that might need to be documented.

- the signature of `self.poutput()` and `self.perror()` is now declaring that `end` is a mandatory kwarg now.
- `self.perror()` and the new `pexcept()` no longer have traceback_war, err_color, war_color (I know color is mentioned, but might be helpful to also mention the specific variable names?)

PR #710, opened 2019-07-06
Add Pipfile to support Pipenv for development

tleonhardt:
This PR adds a **Pipfile** for using [Pipenv](https://docs.pipenv.org/en/latest/) for development.  

`Pipenv` essentially combines the features of `pip` and `virtualenv` into a single tool.  Adding this Pipfile makes it extremely easy to setup a `cmd2` development environment using `pipenv`.  

Also:
- Added `Pipfile.lock` and `.venv` to .gitignore since `cmd2` is a library, not an application and as such we do not intend to use `pipenv` for deployment purposes, only development
- In **setup.py**, replaced `pylint` with `flake8` in the *dev* section since we are using `flake8` for linting as part of our CI process
- Updated **CONTRIBUTING.md** within info about our support for `Pipenv`
- Updated **invoke** task for `flake8` to exclude unnecessary directories so that `flake8` doesn't take forever when using a virtual environment embedded in the `cmd2` directory

To create a virtual environment and install everything needed for `cmd2` development, do the following from a GitHub checkout:
```sh
pipenv install --dev
```

To create a new virtualenv, using a specific version of Python you have installed (and on your PATH), use the --python VERSION flag, like so:
```sh
pipenv install --dev --python 3.7
```

Then you can enter that virtual environment with:
```sh
pipenv shell
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,6 @@ cmd2_history.dat
 
 # Pipenv Lock file
 Pipfile.lock
+
+# Virtualenv directory
+.venv
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,6 @@ dmypy.sock
 
 # cmd2 history file used in hello_cmd2.py
 cmd2_history.dat
+
+# Pipenv Lock file
+Pipfile.lock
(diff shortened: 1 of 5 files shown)


tleonhardt:
I added a section to CONTRIBUTING.md with info on our support for `Pipenv`

PR #715, opened 2019-07-08
Started filling in some missing stuff in the new documentation structure

tleonhardt:
Changes include:
* Removed a couple TODO-style lines from the index which looked like they had been done
* Added info to the "why migrate to cmd2" page
* Added missing info to the "incompatibilities" page

### DIFF
--- a/docs/migrating/incompatibilities.rst
+++ b/docs/migrating/incompatibilities.rst
@@ -43,6 +43,13 @@ contains A list of queued input lines. The cmdqueue list is checked in
 processed in order, as if entered at the prompt.
 
 Since version 0.9.13 ``cmd2`` has removed support for ``Cmd.cmdqueue``.
-Reasoning about application behavior is much easier without this queue present.
+Because ``cmd2`` supports running commands via the main ``cmdloop()``, text
+scripts, Python scripts, transcripts, and history replays, the only way to
+preserve consistent behavior across these methods was to eliminate the command
+queue. Additionally, reasoning about application behavior is much easier
+without this queue present.
+
 If developers need this sort of thing, they can add it in their application.
+However, if they are not extremely careful there would likely be unintended
+consequences.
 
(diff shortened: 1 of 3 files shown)


docs/migrating/incompatibilities.rst
```diff
@@ -7,16 +7,42 @@ Incompatibilities
...
+------------
+In cmd_, the `Cmd.cmdqueue
+<https://docs.python.org/3/library/cmd.html#cmd.Cmd.cmdqueue>`_ attribute
+contains A list of queued input lines. The cmdqueue list is checked in
+``cmdloop()`` when new input is needed; if it is nonempty, its elements will be
+processed in order, as if entered at the prompt.
+
+Since version 0.9.13 ``cmd2`` has removed support for ``Cmd.cmdqueue``.
+Reasoning about application behavior is much easier without this queue present.
+If developers need this sort of thing, they can add it in their application.
```
kmvanbrunt: I would add that because **cmd2** supports running commands via the main loop, text scripts, Python scripts, transcripts, and history replays, the only way to preserve consistent behavior across these methods was to eliminate the command queue.

docs/migrating/why.rst
```diff
@@ -3,23 +3,40 @@ Why Migrate to cmd2
...
+feature-rich interactive command-line applications which delight customers.
 
-Describe why you would want to migrate, and the benefits of doing so
-
-Unicode
-
-features
-
-active community
+``cmd2`` can be used as a drop-in replacement for cmd_.  Simply importing
```
kmvanbrunt: Refer the user to the incompatibilities page

docs/migrating/incompatibilities.rst
tleonhardt: Done

docs/migrating/why.rst
tleonhardt: Done

PR #718, opened 2019-07-11
Auto completer refactor

kmvanbrunt:
This PR is large and breaks stuff with the goal of simplifying our API.

Completely overhauled how argparse argument tab completion is defined.

Instead of:
```Python
setattr(parser.add_argument('my_arg', help='help stuff'),
            ACTION_ARG_CHOICES, ('path_complete',))
```

You do:
```Python
parser.add_argument('my_arg', help='help stuff', completer_method=path_complete)
```

Also patched argparse functions to enable nargs ranges on all parsers and not just the cmd2 parser.

The changes are documented in the header of argparse_custom.py.


cmd2/argparse_completer.py
```diff
@@ -428,159 +217,177 @@ def process_action_nargs(action: argparse.Action, arg_state: AutoCompleter._Argu
...
+                        flag_arg_state = AutoCompleter._ArgumentState(action)
+
+                        # It's possible we already have consumed values for this flag if it was used earlier
+                        # in the command line. Reset them now for this use of the flag.
+                        consumed_arg_values[flag_arg_state.action.dest] = []
+
+                        # Keep track of what flags have already been used
+                        # Flags with action set to append, append_const, and count can be reused
+                        if not is_last_token and \
+                                not isinstance(flag_arg_state.action, argparse._AppendAction) and \
```
xNinjaKittyx: Just a thought, but this check could be line reduced to
`not isinstance(flag_arg_state.action, (argparse._AppendAction, argparse._AppendConstAction, argparse._CountAction))`

since it's equivalent to `not (a or b or c)` -> `not a and not b and not c`

cmd2/argparse_completer.py
```diff
@@ -615,139 +421,115 @@ def _format_completions(self, action, completions: List[Union[str, CompletionIte
...
-                except TypeError:
-                    args = args()
+            # Since arg_choices can be any iterable type, convert to a list
+            arg_choices = list(arg_choices)
 
-            # filter out arguments we already used
-            return [arg for arg in args if arg not in used_values]
+            # Since choices can be various types like int, we must convert them to strings
+            for index, choice in enumerate(arg_choices):
+                if not isinstance(choice, str,):
```
xNinjaKittyx: Lingering comma?

kmvanbrunt: Yep, thanks

cmd2/argparse_completer.py
kmvanbrunt: Good suggestion. I didn't realize isinstance() could take a tuple of types.

cmd2/argparse_custom.py
```diff
@@ -0,0 +1,666 @@
...
+        return format
+
+    # noinspection PyProtectedMember
+    def _format_args(self, action, default_metavar) -> str:
+        get_metavar = self._metavar_formatter(action, default_metavar)
+        # Begin cmd2 customization (less verbose)
+        nargs_range = getattr(action, ATTR_NARGS_RANGE, None)
+
+        if nargs_range is not None:
+            result = '{}{{{}..{}}}'.format('%s' % get_metavar(1), nargs_range[0], nargs_range[1])
```
xNinjaKittyx: Just out of curiosity, is there a difference between doing `'%s' % get_metavar(1)` vs `str(get_metavar(1))` and just plain `get_metavar(1)` in this scenario?

cmd2/argparse_custom.py
```diff
@@ -0,0 +1,666 @@
...
+    def _format_usage(self, usage, actions, groups, prefix) -> str:
+        if prefix is None:
+            prefix = _('Usage: ')
+
+        # if usage is specified, use that
+        if usage is not None:
+            usage %= dict(prog=self._prog)
+
+        # if no optionals or positionals are available, usage is just prog
+        elif usage is None and not actions:
```
tleonhardt: If this is just a copy from `argparse` then fine to leave *as-is*, but this check is redundant, since the above `if` handles the case when `usage` is **NOT** `None`, when we get to this `elif`, we know that `usage` must be `None`.  So this could just be:
```Python
elif not actions:
```

And the below `elif` could just be an `else`.

cmd2/cmd2.py
```diff
@@ -46,7 +46,7 @@
 from . import constants
 from . import plugin
 from . import utils
-from .argparse_completer import AutoCompleter, ACArgumentParser, ACTION_ARG_CHOICES
+from .argparse_custom import Cmd2ArgParser, CompletionItem
```
tleonhardt: Consider renaming `Cmd2ArgParser` to just `ArgParser` - in this context it is clear it is coming from `cmd2`, so the **Cmd2** prefix in the name is perhaps redundant.

cmd2/cmd2.py
```diff
@@ -467,6 +467,11 @@ def __init__(self, completekey: str = 'tab', stdin=None, stdout=None, *,
         # Otherwise it can be set to any custom key to meet your needs.
         self.matches_sort_key = ALPHABETICAL_SORT_KEY
 
+        # The maximum number of CompletionItems to display during tab completion. If the number of completion
+        # suggestions exceeds this number, they will be displayed in the typical columnized format and will
+        # not include the description value of the CompletionItems.
+        self.max_completion_items = 50
```
tleonhardt: Would it be useful for this to be in `settable`?

cmd2/cmd2.py
```diff
@@ -835,23 +841,8 @@ def tokens_for_completion(self, line: str, begidx: int, endidx: int) -> Tuple[Li
 
         return tokens, raw_tokens
 
-    # noinspection PyUnusedLocal
-    @staticmethod
-    def basic_complete(text: str, line: str, begidx: int, endidx: int, match_against: Iterable) -> List[str]:
```
tleonhardt: Should we add more to the CHANGELOG in regards to some of the more minor breaking changes such as moving `basic_complete` from a method in `cmd.Cmd` to a function in `utils`?

Same goes for making all arguments passed the standard four to completer functions keyword only?

tests/test_argparse_completer.py
```diff
@@ -0,0 +1,678 @@
+# coding=utf-8
+# flake8: noqa E302
+"""
+Unit/functional testing for argparse completer in cmd2
```
tleonhardt: Great job increasing code coverage with new and improved unit tests!

kmvanbrunt:
Temporarily closing PR. I am adding one more feature.

cmd2/argparse_completer.py
```diff
@@ -428,159 +217,176 @@ def process_action_nargs(action: argparse.Action, arg_state: AutoCompleter._Argu
...
+
         # we're not at a positional argument, see if we're in a flag argument
-        elif not current_is_positional:
-            if flag_action is not None:
-                consumed = consumed_arg_values[flag_action.dest]\
-                    if flag_action.dest in consumed_arg_values else []
-                # current_items.extend(self._resolve_choices_for_arg(flag_action, consumed))
-                completion_results = self._complete_for_arg(flag_action, text, line, begidx, endidx, consumed)
+        if not current_is_positional:
+            if flag_arg_state is not None:
```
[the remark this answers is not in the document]
kmvanbrunt: This no longer applies due to extensive refactoring last night

cmd2/argparse_custom.py
kmvanbrunt: This is a copy from argparse, so I'll leave it.

cmd2/argparse_custom.py
kmvanbrunt: Yes, since get_metavar(1) is a 1-item tuple the behavior is different.
Assume it's value is `('{2, 3}',)`

`'%s' % get_metavar(1)` is the same as `str(get_metavar(1)[0]`
So they would print: `{2, 3}`

whereas

`str(get_metavar)` will print the entire tuple: `('{2, 3}',)`

kmvanbrunt: Instead of switching to `get_metavar(1)[0]`, I'll keep the `'%s' % get_metavar(1)` approach to match the code style from the argparse function we are overriding.

kmvanbrunt:
PR is open again

kotfu:
This is much nicer than what we have now. I like it a lot. A couple of thoughts:

- Should we name our custom `ArgParser` class `ArgumentParser`? Then you would either use `cmd2.ArgumentParser` or `argparse.ArgumentParser`.
- The custom `ArgParser` class over-rides several private methods of `argparse.ArgumentParser`. This makes me nervous. It could lead to a situation where a minor python revision requires us to do a **x.0.0** version release. 
- These changes to ArgumentParser are potentially interesting enough to split out into a separate library. Have you thought about it?

cmd2/cmd2.py

kmvanbrunt:
@kotfu Thanks for the feedback. Here are my thoughts.

1. I'm OK renaming the parser to `cmd2.ArgumentParser`. @tleonhardt What are your thoughts?
2. I too am a little uncomfortable overriding private methods of `ArgumentParser`. Hopefully we won't get into much trouble since most of the functions are merely wrappers that handle our custom data before calling down to the original.
3. I hadn't considered splitting any of the `ArgumentParser` changes into another library. What are you envisioning?

CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
 ## 0.9.15 (July TBD, 2019)
+*# Enhancements
+    * Greatly simplified using argparse-based tab completion. The new interface is a complete overhaul that breaks
+    the previous way of specifying completion and choices functions. See header of argparse_completer.py for more
+    information.
```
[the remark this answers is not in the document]
kmvanbrunt: Linking to argparse_custom.py would be better since I moved all the documentation there. Where should I put the link?

tleonhardt:
I'm fine renaming our custom parser to `ArgumentParser` and I'm also fine leaving it as `ArgParser` - I see advantages and disadvantages either way.  The advantage of naming it slightly differently is it emphasizes the differences from `argparse.ArgumentParser`.  The advantage of naming it the same is it emphasizes the similarities.

CHANGELOG.md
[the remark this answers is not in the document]
tleonhardt: I would modify the CHANGELOG to say something like:
* Greatly simplified using argparse-based tab completion. The new interface is a complete overhaul that breaks the previous way of specifying completion and choices functions. See header of [argparse_custom.py](https://github.com/python-cmd2/cmd2/blob/master/cmd2/argparse_custom.py) for more information.

The link won't work until after the PR merges to master, but at least it will give anyone wanting to know more an easy link to more info.

kmvanbrunt:
I'll merge this for now and give some thought to the ArgParser/ArgumentParser name choice.

Issue #572, opened 2018-10-09
Goals for 1.0 release of cmd2


tleonhardt:
What are and/or should be our goals for a 1.0.0 release of ``cmd2``?  This issue is for discussing that topic.

Tentatively, my goals for a 1.0 release are as follows:
1. Stabilize the API so we don’t have many deletions or deprecations in the future
1. Make any remaining big changes we want before this
1. Improve documentation
1. Release 1.0 when it is ready
1. Once 1.0 has been released, do a better job of making its existence known in the community

The following Milestones were created to work towards whatever goals we mutually agree to:
* 0.9.14 - Get any remaining “big changes” into this release if possible.  Any remaining deprecations prior to 1.0.0 should go in this release.  We should also strive to fix as many known bugs as possible in this release.
    * July 2019
* 0.9.15 - Last release prior to 1.0.0. No further deprecations, but deletion of previously deprecated OK.  We should strive hard to have a stable API at this point.
    * August 2019
* 1.0.0 - Stable 1.0.0 release ready for wide usage.  Bug fix and polish release. Focus on documentation improvements, bug fixes, etc. but not major new features. No deletions or deprecations.
    * September 2019

marklalor:
For 0.9.9, can there be discussion about some high-level documentation ideas, like the order that documentation is presented in. As an example, I think tab completion should be higher up in the "Features requiring no modifications" section. I'm sure there are plenty of things of that nature that I haven't thought about.

tleonhardt:
I like he idea of having a discussion for the 0.9.9 release about how to best improve the existing documentation.  A significant reorganization would be beneficial - we have added a lot over time, but haven’t made the effort to go back and reorder things.

I think that documentation improvements are **the** most important improvements we can make for a 1.0.0 release.

tleonhardt:
I'm going to close this since I think the goals are clear and the only thing really remaining is to improve the documentation a bit further, and there is another ticket for that open.

PR #720, opened 2019-07-15
Write documentation for migrating from cmd

### LINKED ISSUE #719: Finish migration documentation
Finish the documentation showing how to migrate from cmd.

kotfu:
I've got another section of the documentation filled out: how to migrate from cmd to cmd2. Closes #719.


docs/migrating/incompatibilities.rst
```diff
@@ -14,24 +14,33 @@ The `Cmd.emptyline()
...
-whitespace.  This has the added benefit of full unicode support, unlike cmd_ or
-versions of ``cmd2`` prior to 0.9.0.
+contains the string of characters accepted for command names. cmd_ uses those
+characters to split the first "word" of the input, without requiring the user
+to type a space. For example, if ``identchars`` contained a string of all
+alphabetic characters, the user could enter a command like ``L20`` and it would
+be interpreted as the command ``L`` with the first argument of ``20``.
+
+Since version 0.9.0, ``cmd2`` has ignored ``identchars``; the parsing logic in
+``cmd2`` splits the command and arguments on whitespace. While cmd_ technically
```
kmvanbrunt: > While cmd_ technically supports unicode, but using non-ascii unicode characters in command names while
simultaneously using ``identchars`` functionality is non-trivial.

This sentence is a bit awkward.

kmvanbrunt: > While cmd_ technically supports unicode, but using non-ascii unicode characters in command names while
simultaneously using ``identchars`` functionality is non-trivial.

This sentence is a bit awkward.

kotfu: More than a bit awkward, it's terrible. Fixed.

cmd2/cmd2.py
```diff
@@ -2977,7 +2977,7 @@ def do_set(self, args: argparse.Namespace) -> None:
         if orig_value != new_value:
             onchange_hook = getattr(self, '_onchange_{}'.format(param), None)
             if onchange_hook is not None:
-                onchange_hook(old=orig_value, new=new_value)
+                onchange_hook(old=orig_value, new=new_value)  # pylint: disable=not-callable
```
tleonhardt: Would it be better to do something like the following:
```Python
if onchange_hook is not None and callable(onchange_hook):
    onchange_hook(old=orig_value, new=new_value)
```

docs/features/clipboard.rst
```diff
@@ -1,5 +1,10 @@
 Clipboard Integration
 =====================
 
+  - sent to the operating system paste buffer, by ending with a bare ``>``, as
```
tleonhardt: Should this begin with something like **Command output can be sent ...***?

cmd2/cmd2.py
kotfu: I dunno. If there is an attribute that a developer created called `_onchange_mysetting` that isn't a callable, then the code as written would generate an exception. Same thing if it was callable, but took 1 (or 3) parameters. Is it better to silently ignore the malformed or uncallable onchange hook, or to throw an exception?

By way of reference, in the new implementation of hooks, it checks both callable and method signature when you register a hook, and won't register the hook if the registered callable doesn't have the right number of parameters that are annotated with the expected types for each parameter. The registration hook method throws TypeErrors if anything is amiss.

I was just trying to get rid of the pylint error. Pylint still generates the error even if we check if the hook is callable.

I've pushed a new commit with the proposed change, because even it could still throw errors if the method has the wrong number of parameters, at least we won't try something that isn't callable.

docs/features/clipboard.rst
kotfu: Yes, this was just a stupid cut and paste from the old documentation. New commit pushed with a real clipboard integration page.

cmd2/cmd2.py
```diff
@@ -2962,8 +2962,8 @@ def do_set(self, args: argparse.Namespace) -> None:
         # See if we need to call a change hook for this settable
         if orig_value != new_value:
             onchange_hook = getattr(self, '_onchange_{}'.format(param), None)
-            if onchange_hook is not None:
-                onchange_hook(old=orig_value, new=new_value)
+            if onchange_hook is not None and callable(onchange_hook):
```
tleonhardt: Your comments made me change my mind - if a developer uses this incorrectly it would be better to throw an exception than fail silently.  Sorry 😐

kotfu: No worries. I've reverted the change, and am going to merge this PR.

kotfu:
This PR helps with #572, improving documentation before 1.0 release.

PR #730, opened 2019-07-18
Added sigint protection while entering/leaving cmd2 and Python interactive shells

### LINKED ISSUE #686: Add more sigint protection
Add some more sigint protection around blocks of code that alter or restore critical state data of cmd2. This includes blocks that edit or restore tab completion settings like in `_cmdloop()` and `do_py()`.

kmvanbrunt:
Fixes #686
This should prevent getting into mixed environment states by allowing environment changes to occur atomically.

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -2297,12 +2297,12 @@ class Cmd(cmd.Cmd):
 
         This serves the same role as cmd.cmdloop().
         """
-        readline_settings = None
+        saved_readline_settings = None
 
         try:
             # Get sigint protection while we set up readline for cmd2
             with self.sigint_protection:
-                readline_settings = self._set_up_cmd2_readline()
+                saved_readline_settings = self._set_up_cmd2_readline()
 
             # Run startup commands
             stop = self.runcmds_plus_hooks(self._startup_commands)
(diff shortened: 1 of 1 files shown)


cmd2/cmd2.py
```diff
@@ -280,6 +281,24 @@ def cmd_wrapper(cmd2_instance, statement: Union[Statement, str]):
     return arg_decorator
 
 
+class _SavedReadlineSettings:
```
tleonhardt: Does this code belong in rl_utils.py?

cmd2/cmd2.py
```diff
@@ -280,6 +281,24 @@ def cmd_wrapper(cmd2_instance, statement: Union[Statement, str]):
...
+    """readline settings that are backed up when switching between readline environments"""
+    def __init__(self):
+        self.completer = None
+        self.delims = ''
+        self.basic_quotes = None
+
+
+class _SavedCmd2Env:
```
tleonhardt: Should this be in either utils.py or in pyscript_bridge.py?

cmd2/cmd2.py
```diff
@@ -38,6 +38,7 @@
 import re
 import sys
 import threading
+from code import InteractiveConsole
```
tleonhardt: Lots of new code is added here, but no new unit tests.  Should a few unit tests be added?

cmd2/cmd2.py
kmvanbrunt: Since this class is only used in cmd2.py, I'm OK leaving it.

cmd2/cmd2.py
kmvanbrunt: Since this class is only used in cmd2.py, I'm OK leaving it.

cmd2/cmd2.py
kmvanbrunt: I really just moved existing code into functions. No unit test or coverage changes.

kmvanbrunt:
@tleonhardt No need for CHANGELOG updates on this

PR #732, opened 2019-07-19
Fix broken documentation links

tleonhardt:
Fixed broken documentation links in `README.md`, `CONTRIBUTING.md`, and for recent releases in `CHANGELOG.md`.

Also:
- Moved **Examples** before **API Reference** in overall documentation index
- Added an example project in `README.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ Instructions for implementing each feature follow.
         self.stdout.write('{}\n'.format(' '.join(words)))
     ```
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -109,15 +109,15 @@ Instructions for implementing each feature follow.
         - Do `help history` in any `cmd2` application for more information
 
 - Simple scripting using ASCII text files with one command + arguments per line
```
kmvanbrunt: This should probably just say "text files" since utf-8 also works.

tleonhardt: Good catch

PR #738, opened 2019-07-22
Presentation changes

kmvanbrunt:


cmd2/cmd2.py
```diff
@@ -2391,7 +2396,7 @@ def _alias_delete(self, args: argparse.Namespace) -> None:
             self.aliases.clear()
             self.poutput("All aliases deleted")
         elif not args.name:
-            self.do_help('alias delete')
+            self.perror("Either --all or alias name(s) must be specified")
```
tleonhardt: Is there a way to enforce things to be correct at the argparse level so that this check is unnecessary?

kmvanbrunt: None that I know of

cmd2/argparse_custom.py
```diff
@@ -81,15 +81,19 @@ def my_completer_function(text, line, begidx, endidx):
             parser.add_argument('-o', '--options', completer_method=cmd2.Cmd.path_complete)
 
 
-    In all cases in which function/methods are passed you can use functools.partial() to prepopulate
-    values of the underlying function.
+    You can use functools.partial() to prepopulate values of the underlying choices and completer functions/methods.
```
tleonhardt: Nice comments added

PR #747, opened 2019-08-02
Fixed inconsistent treatment of allow_redirection

### LINKED ISSUE #746: Purpose of allow_redirection
What is `allow_redirection` supposed to do? Is it there only to prevent redirection of stdout or is it also intended to alter parsing behavior?

When set to **False,** it slightly, but not completely, alters parsing.
- `command > out.txt` results in `command `
- `command >out.txt` results in `command  >out.txt`

If  `allow_redirection` is False, should we treat redirectors as literals or preserve normal parsing behavior and just consider `allow_redirection` to be a security setting?

See #15 and #45 for some context.

kmvanbrunt:
Fixed inconsistent parsing/tab completion behavior based on the value of `allow_redirection`. This flag is only meant to be a security setting that prevents redirection of stdout and should not alter parsing logic.

Fixed #746

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
 ## 0.9.16 (TBD, 2019)
+* Bug Fixes
+    * Fixed inconsistent parsing/tab completion behavior based on the value of `allow_redirection`. This flag is
+    only meant to be a security setting that prevents redirection of stdout and should not alter parsing logic.
 * Enhancements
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -1,4 +1,7 @@
 ## 0.9.16 (TBD, 2019)
+* Bug Fixes
```
tleonhardt: Looks good

PR #749, opened 2019-08-02
Added terminators to completion delimiters

kmvanbrunt:
Tab completion should have always been using terminators as word delimiters.
This change also resulted in eliminating duplicated code since cmd2's `tokens_for_completion()` can just call  `StatementParser.split_on_punctuation()` now.

@kotfu @tleonhardt What are your thoughts on me changing `StatementParser.split_on_punctuation()` to public?

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -824,56 +824,8 @@ class Cmd(cmd.Cmd):
                     # Return empty lists since this means the line is malformed.
                     return [], []
 
-        # We need to treat redirection characters (|, <, >) as word breaks when they are in unquoted strings.
-        # Go through each token and further split them on these characters. Each run of redirect characters
-        # is treated as a single token.
-        raw_tokens = []
-
-        for cur_initial_token in initial_tokens:
-
-            # Save tokens up to 1 character in length or quoted tokens. No need to parse these.
-            if len(cur_initial_token) <= 1 or cur_initial_token[0] in constants.QUOTES:
-                raw_tokens.append(cur_initial_token)
-                continue
-
-            # Iterate over each character in this token
-            cur_index = 0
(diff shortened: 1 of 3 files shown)


cmd2/cmd2.py
```diff
@@ -824,56 +824,8 @@ def tokens_for_completion(self, line: str, begidx: int, endidx: int) -> Tuple[Li
...
-
-                # Save the current token
-                raw_tokens.append(cur_raw_token)
-                cur_raw_token = ''
-
-                # Check if we've viewed all characters
-                if cur_index >= len(cur_initial_token):
-                    break
+        # Further split tokens on punctuation characters
+        raw_tokens = self.statement_parser.split_on_punctuation(initial_tokens)
```
tleonhardt: Nice simplification and code reuse

cmd2/parsing.py
```diff
@@ -675,7 +675,7 @@ def _command_and_args(tokens: List[str]) -> Tuple[str, str]:
 
         return command, args
 
-    def _split_on_punctuation(self, tokens: List[str]) -> List[str]:
+    def split_on_punctuation(self, tokens: List[str]) -> List[str]:
```
tleonhardt: I'm fine making this public, especially since we are using it in cmd2.py

tests/test_completion.py
```diff
@@ -689,26 +689,27 @@ def test_tokens_for_completion_unclosed_quote(cmd2_app):
     assert expected_tokens == tokens
     assert expected_raw_tokens == raw_tokens
 
-def test_tokens_for_completion_redirect(cmd2_app):
-    text = '>>file'
-    line = 'command | < {}'.format(text)
+def test_tokens_for_completion_punctuation(cmd2_app):
```
tleonhardt: Thanks for updating the tests

PR #752, opened 2019-08-06
Removed ability for aliases and macros to share names with commands

### LINKED ISSUE #751: Alias and Macro names
It has been brought to my attention by another developer that allowing aliases and macros to share names with commands could cause issues with users being unwise. Specifically he said:

> being able to create an alias that overrides/hides a command is bad for usability

Currently we have the following restrictions:

1. Alias cannot share its name with a macro
2. Macro cannot share its name with an alias or a multiline command

When I first wrote aliases, I did prevent them from sharing their name with a command. Then I decided to remove this restriction to be like bash. However, in bash you can still call the original command if you know its path. (e.g. /bin/ls). **cmd2** does not work in the same way since our commands are not executable files somewhere on disk.

I'm OK adding the restriction back, but I would like some consensus on the issue. And I really don't want the answer to be, "add a flag".

@kotfu @tleonhardt @anselor What are your thoughts?

kmvanbrunt:
Closes #751 
This will help avoid confusion for both developers and users.

Developers won't have to worry that calling the onecmd() family of functions will run a different command than expected because it was aliased. Each of these functions was simplified by removing their **expand** boolean argument. That results in less maintenance.

User will always know what command they are running if a command name is used since these names can no longer be used for aliases or macros.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
     * Added `set_choices_function()`, `set_choices_method()`, `set_completer_function()`, and `set_completer_method()`
     to support cases where this functionality needs to be added to an argparse action outside of the normal
     `parser.add_argument()` call.
+* Breaking Changes
+    * Aliases and macros can no longer have the same name as a command
 
 ## 0.9.15 (July 24, 2019)
 * Bug Fixes
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -1672,13 +1672,10 @@ class Cmd(cmd.Cmd):
         statement = self.statement_parser.parse_command_only(line)
         return statement.command, statement.args, statement.command_and_args
 
-    def onecmd_plus_hooks(self, line: str, *, expand: bool = True, add_to_history: bool = True,
-                          py_bridge_call: bool = False) -> bool:
+    def onecmd_plus_hooks(self, line: str, *, add_to_history: bool = True, py_bridge_call: bool = False) -> bool:
         """Top-level function called by cmdloop() to handle parsing a line and running the command and all of its hooks.
 
         :param line: command line to run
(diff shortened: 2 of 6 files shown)


cmd2/cmd2.py
```diff
@@ -1672,13 +1672,10 @@ def parseline(self, line: str) -> Tuple[str, str, str]:
         statement = self.statement_parser.parse_command_only(line)
         return statement.command, statement.args, statement.command_and_args
 
-    def onecmd_plus_hooks(self, line: str, *, expand: bool = True, add_to_history: bool = True,
-                          py_bridge_call: bool = False) -> bool:
+    def onecmd_plus_hooks(self, line: str, *, add_to_history: bool = True, py_bridge_call: bool = False) -> bool:
```
tleonhardt: This is a nice simplification that ripples through a number of places and can help avoid some potentially confusing scenarios for users.

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
     * Added `set_choices_function()`, `set_choices_method()`, `set_completer_function()`, and `set_completer_method()`
     to support cases where this functionality needs to be added to an argparse action outside of the normal
     `parser.add_argument()` call.
+* Breaking Changes
+    * Aliases and macros can no longer have the same name as a command
```
tleonhardt: Recommend to also update the appropriate Sphinx documentation: https://cmd2.readthedocs.io/en/latest/features/shortcuts_aliases_macros.html

docs/features/shortcuts_aliases_macros.rst
```diff
@@ -65,6 +65,7 @@ Use ``alias delete`` to remove aliases
 
 For more details run: ``help alias delete``
 
+Note: Aliases cannot have the same name as a command or macro
```
tleonhardt: Technically "or shortcut", but this is probably fine

PR #754, opened 2019-08-07
Disabled tab completion during a select call

kmvanbrunt:
Disabled tab completion during a select call.
Added graceful handling of Ctrl-D during select. It now reprompts the user.

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -684,7 +684,7 @@ class Cmd(cmd.Cmd):
             final_msg = ansi.style_error(final_msg)
 
         if not self.debug:
-            warning = "\nTo enable full traceback, run the following command:  'set debug true'"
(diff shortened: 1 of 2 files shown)


cmd2/cmd2.py
```diff
@@ -2904,15 +2926,28 @@ def select(self, opts: Union[str, List[str], List[Tuple[Any, Optional[str]]]],
...
         for (idx, (_, text)) in enumerate(fulloptions):
             self.poutput('  %2d. %s' % (idx + 1, text))
+
         while True:
             safe_prompt = rl_make_safe_prompt(prompt)
-            response = input(safe_prompt)
+
+            try:
+                with self.sigint_protection:
+                    disable_completion()
```
tleonhardt: Yeah, this seems like a good idea.

PR #755, opened 2019-08-07
Verifying command names in __init__

kmvanbrunt:
Preventing things like commands that begin with a shortcut

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -448,6 +448,12 @@ class Cmd(cmd.Cmd):
                                                 multiline_commands=multiline_commands,
                                                 shortcuts=shortcuts)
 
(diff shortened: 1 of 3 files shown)


cmd2/cmd2.py
```diff
@@ -448,6 +448,12 @@ def __init__(self, completekey: str = 'tab', stdin=None, stdout=None, *,
                                                 multiline_commands=multiline_commands,
                                                 shortcuts=shortcuts)
 
+        # Verify commands don't have invalid names (like starting with a shortcut)
+        for cur_cmd in self.get_all_commands():
```
tleonhardt: Yeah this is prudent and I can't think of a better way to validate it.

PR #756, opened 2019-08-08
Do not print traceback warning text if debug is not a settable parameter

kmvanbrunt:


cmd2/cmd2.py
```diff
@@ -689,7 +689,7 @@ def pexcept(self, msg: Any, *, end: str = '\n', apply_style: bool = True) -> Non
         if apply_style:
             final_msg = ansi.style_error(final_msg)
 
-        if not self.debug:
+        if not self.debug and 'debug' in self.settable:
```
tleonhardt: Nice little ruggedization

PR #758, opened 2019-08-09
No longer treating empty text scripts as an error condition

kmvanbrunt:


cmd2/cmd2.py
```diff
@@ -3790,9 +3790,8 @@ def do_run_script(self, args: argparse.Namespace) -> Optional[bool]:
             self.perror("'{}' is not a file".format(expanded_path))
             return
 
-        # Make sure the file is not empty
+        # An empty file is not an error, so just return
```
tleonhardt: Is it worth adding something in the CHANGELOG in regards to this change?

PR #774, opened 2019-09-13
Completion state

### LINKED ISSUE #762: pass the namespace to completion function
I would like to set a `choices_function` that depends on the previous arguments, i.e., I believe cmd2 should pass the current parser namespace when calling the `choices_function`
Like I have this first argument that loads a csv file, and I would like the 2nd argument to choose between availables ids in the previously loaded csv file.
Right now the` choices_function` looks a bit useless and can only return static results.

kmvanbrunt:
Choices/Completer functions can now be passed a dictionary that maps command-line tokens to their argparse argument. This is helpful when one argument determines what is tab completed for another argument. If these functions have an argument called `arg_tokens`, then AutoCompleter will automatically pass this dictionary to them.

Closes #762

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,11 @@
     * Fixed a bug when running a cmd2 application on Linux without Gtk libraries installed
 * Enhancements
     * No longer treating empty text scripts as an error condition
+    * Allow dynamically extending a `cmd2.Cmd` object instance with a `do_xxx` method at runtime
     * Choices/Completer functions can now be passed a dictionary that maps command-line tokens to their
     argparse argument. This is helpful when one argument determines what is tab completed for another argument.
     If these functions have an argument called `arg_tokens`, then AutoCompleter will automatically pass this
-    Namespace to them.
+    dictionary to them.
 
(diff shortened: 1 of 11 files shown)


tests/test_argparse_completer.py
```diff
@@ -30,6 +30,9 @@
 completions_from_function = ['completions', 'function', 'fairly', 'complete']
 completions_from_method = ['completions', 'method', 'missed', 'spot']
 
+set_one_choices = ['this', 'is', 'set', 'one']
```
tleonhardt: Thanks for adding tests.

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@
 * Enhancements
     * No longer treating empty text scripts as an error condition
     * Allow dynamically extending a `cmd2.Cmd` object instance with a `do_xxx` method at runtime
+    * Choices/Completer functions can now be passed a dictionary that maps command-line tokens to their
```
tleonhardt: I like the change to a dictionary instead of an argparse.Namespace

PR #783, opened 2019-09-24
Help fix

kmvanbrunt:
Fixed bug introduced by get_names() looking at self instead of self.__class__ where help functions for hidden and disabled commands were not being filtered out as help topics.

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -1625,19 +1625,8 @@ class Cmd(cmd.Cmd):
 
     def get_visible_commands(self) -> List[str]:
         """Return a list of commands that have not been hidden or disabled"""
-        commands = self.get_all_commands()
-
-        # Remove the hidden commands
-        for name in self.hidden_commands:
-            if name in commands:
-                commands.remove(name)
-
-        # Remove the disabled commands
-        for name in self.disabled_commands:
-            if name in commands:
-                commands.remove(name)
-
-        return commands
+        return [command for command in self.get_all_commands()
+                if command not in self.hidden_commands and command not in self.disabled_commands]
 
     def _get_alias_completion_items(self) -> List[CompletionItem]:
         """Return list of current alias names and values as CompletionItems"""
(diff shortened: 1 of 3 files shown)


cmd2/cmd2.py
```diff
@@ -1625,19 +1625,8 @@ def get_all_commands(self) -> List[str]:
...
-            if name in commands:
-                commands.remove(name)
-
-        # Remove the disabled commands
-        for name in self.disabled_commands:
-            if name in commands:
-                commands.remove(name)
-
-        return commands
+        return [command for command in self.get_all_commands()
```
tleonhardt: This is much cleaner but equivalent to the previous code

cmd2/cmd2.py
```diff
@@ -1659,9 +1648,13 @@ def _get_commands_aliases_and_macros_for_completion(self) -> List[str]:
...
     def get_help_topics(self) -> List[str]:
-        """ Returns a list of help topics """
-        return [name[len(HELP_FUNC_PREFIX):] for name in self.get_names()
-                if name.startswith(HELP_FUNC_PREFIX) and callable(getattr(self, name))]
+        """Return a list of help topics"""
+        all_topics = [name[len(HELP_FUNC_PREFIX):] for name in self.get_names()
+                      if name.startswith(HELP_FUNC_PREFIX) and callable(getattr(self, name))]
+
+        # Filter out hidden and disabled commands
+        return [topic for topic in all_topics
```
tleonhardt: Would it be better to combine the `if`'s in this list comprehension with the previous list comprehension?  Or would that make the code difficult to read?

tests/test_cmd2.py
```diff
@@ -1981,6 +1981,23 @@ def test_get_help_topics(base_app):
     custom_help = base_app.get_help_topics()
     assert len(custom_help) == 0
 
+def test_get_help_topics_hidden():
```
tleonhardt: Thanks for adding relevant tests

cmd2/cmd2.py
kmvanbrunt: It would look like this

```Python
"""Return a list of help topics"""
all_topics = [name[len(HELP_FUNC_PREFIX):] for name in self.get_names()
              if name.startswith(HELP_FUNC_PREFIX) and callable(getattr(self, name)) and
              [name[len(HELP_FUNC_PREFIX):] not in self.hidden_commands and name[len(HELP_FUNC_PREFIX):] not in self.disabled_commands]
```

PR #789, opened 2019-10-18
Script check

kmvanbrunt:
Send all startup script paths to run_script. Previously we didn't do this if the file was empty, but that showed no record of the run_script command in history.

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -498,10 +498,10 @@ class Cmd(cmd.Cmd):
         # Commands that will run at the beginning of the command loop
         self._startup_commands = []
 
-        # If a startup script is provided, then execute it in the startup commands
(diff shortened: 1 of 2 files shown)


cmd2/cmd2.py
```diff
@@ -498,10 +498,10 @@ def __init__(self, completekey: str = 'tab', stdin=None, stdout=None, *,
         # Commands that will run at the beginning of the command loop
         self._startup_commands = []
 
-        # If a startup script is provided, then execute it in the startup commands
+        # If a startup script is provided and exists, then execute it in the startup commands
         if startup_script:
             startup_script = os.path.abspath(os.path.expanduser(startup_script))
-            if os.path.exists(startup_script) and os.path.getsize(startup_script) > 0:
+            if os.path.exists(startup_script):
```
tleonhardt: This change seems reasonable

PR #794, opened 2019-10-30
ipy fix

### LINKED ISSUE #792: use_ipython = False should not delete do_ipy from the class
When `use_ipython` is set to False, only that `cmd2.Cmd` instance should have `do_ipy` removed. Currently the method being deleted from the class.

kmvanbrunt:
Fixed bug where setting `use_ipython` to False removed ipy command from the entire `cmd2.Cmd` class instead of just the instance being created
Fixes #792

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -368,10 +368,10 @@ class Cmd(cmd.Cmd):
         :param shortcuts: dictionary containing shortcuts for commands. If not supplied, then defaults to
                           constants.DEFAULT_SHORTCUTS.
         """
-        # If use_ipython is False, make sure the do_ipy() method doesn't exit
+        # If use_ipython is False, make sure the ipy command isn't available in this instance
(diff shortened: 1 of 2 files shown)


cmd2/cmd2.py
```diff
@@ -368,10 +368,10 @@ def __init__(self, completekey: str = 'tab', stdin=None, stdout=None, *,
         :param shortcuts: dictionary containing shortcuts for commands. If not supplied, then defaults to
                           constants.DEFAULT_SHORTCUTS.
         """
-        # If use_ipython is False, make sure the do_ipy() method doesn't exit
+        # If use_ipython is False, make sure the ipy command isn't available in this instance
         if not use_ipython:
             try:
-                del Cmd.do_ipy
+                self.do_ipy = None
```
tleonhardt: This seems like a better way to deal with it, good call

PR #795, opened 2019-10-30
do_history no longer calls do_edit

### LINKED ISSUE #793: History should not call do_edit
The `do_history` method calls `do_edit` and this causes problems when a **cmd2** app has overridden the **edit** command. Change `do_history` to call non-command function for its edit functionality.

kmvanbrunt:
do_history no longer calls do_edit since edit is a command that is commonly overridden
Fixes #793

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -433,7 +433,7 @@ class Cmd(cmd.Cmd):
         self._initialize_history(persistent_history_file)
 
         # Commands to exclude from the history command
-        self.exclude_from_history = '''history edit eof'''.split()
+        self.exclude_from_history = ['eof', 'history']
 
         # Dictionary of macro names and their values
         self.macros = dict()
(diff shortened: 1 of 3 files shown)


cmd2/cmd2.py
```diff
@@ -3520,14 +3520,9 @@ def do_history(self, args: argparse.Namespace) -> Optional[bool]:
                     else:
                         fobj.write('{}\n'.format(command.raw))
             try:
-                # Handle potential edge case where the temp file needs to be quoted on the command line
-                quoted_fname = utils.quote_string(fname)
-
-                # noinspection PyTypeChecker
-                self.do_edit(quoted_fname)
-
+                self._run_editor(fname)
```
tleonhardt: Not directly calling the **edit** command see within the **history** command is a good thing as it removes the possibility of a pathological infinite loop which then allows us to put **edit** commands in the history.

PR #796, opened 2019-10-30
Recursively set parser.prog

kmvanbrunt:
It is no longer necessary to set the `prog` attribute of an argparser with subcommands. cmd2 now automatically sets the prog value of it and all its subparsers so that all usage statements contain the top level command name and not sys.argv[0].

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -177,14 +177,37 @@ def with_argument_list(*args: List[Callable], preserve_quotes: bool = False) -> 
         return arg_decorator
 
 
-def with_argparser_and_unknown_args(argparser: argparse.ArgumentParser, *,
+def set_parser_prog(parser: argparse.ArgumentParser, prog: str):
+    """
+    Recursively set prog attribute of a parser and all of its subparsers so that the root command
+    is a command name and not sys.argv[0].
+    :param parser: the parser being edited
+    :param prog: value for the current parsers prog attribute
+    """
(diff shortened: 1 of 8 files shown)


cmd2/cmd2.py
```diff
@@ -177,14 +177,38 @@ def cmd_wrapper(cmd2_app, statement: Union[Statement, str]):
         return arg_decorator
 
 
-def with_argparser_and_unknown_args(argparser: argparse.ArgumentParser, *,
+# noinspection PyProtectedMember
+def set_parser_prog(parser: argparse.ArgumentParser, prog: str):
```
tleonhardt: Would this utility function be more appropriate in **argparse_custom.py** or here?

Actually, it begs the question would all of the decorators and associated utility functions belong better in a separate file?

cmd2/cmd2.py
```diff
@@ -2396,7 +2418,7 @@ def _alias_list(self, args: argparse.Namespace) -> None:
                          "An alias is a command that enables replacement of a word by another string.")
     alias_epilog = ("See also:\n"
                     "  macro")
-    alias_parser = Cmd2ArgumentParser(description=alias_description, epilog=alias_epilog, prog='alias')
+    alias_parser = Cmd2ArgumentParser(description=alias_description, epilog=alias_epilog)
```
tleonhardt: This is a very nice little improvement that provides a better user experience.  Thanks for the PR!

PR #798, opened 2019-10-31
Refactoring

kmvanbrunt:
Moving a bit of code around to shrink cmd2.py

- Moved cmd2 decorators into a new file called decorators.py
- Moved some constants from cmd2.py to constants.py
- Made some cmd2 module constants cmd2.Cmd class constants

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -89,14 +89,6 @@ try:
 except ImportError:  # pragma: no cover
     ipython_available = False
 
-INTERNAL_COMMAND_EPILOG = ("Notes:\n"
-                           "  This command is for internal use and is not intended to be called from the\n"
-                           "  command line.")
-
-# Sorting keys for strings
-ALPHABETICAL_SORT_KEY = utils.norm_fold
-NATURAL_SORT_KEY = utils.natural_keys
-
 
 class _SavedReadlineSettings:
     """readline settings that are backed up when switching between readline environments"""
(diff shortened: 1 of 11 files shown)


cmd2/cmd2.py
```diff
@@ -89,235 +89,6 @@
...
-        setattr(cmd_wrapper, CMD_ATTR_ARGPARSER, parser)
-        setattr(cmd_wrapper, CMD_ATTR_PRESERVE_QUOTES, preserve_quotes)
-
-        return cmd_wrapper
-
-    # noinspection PyTypeChecker
-    return arg_decorator
-
 
 class _SavedReadlineSettings:
```
tleonhardt: If we are moving things out of **cmd2.py**, should we consider moving `_SavedReadlineSettings` to **rl_utils.py** and `_SavedCmd2Env` to **utils.py**?

docs/features/argument_processing.rst
```diff
@@ -40,11 +40,11 @@ Decorators provided by cmd2 for argument processing
 ``cmd2`` provides the following decorators for assisting with parsing arguments
 passed to commands:
 
-.. automethod:: cmd2.cmd2.with_argument_list
+.. automethod:: cmd2.decorators.with_argument_list
```
tleonhardt: I wonder if it should be `autofunction` instead of `automethod` here?  Or should it be `automethod` in the **api/decorators.rst** file?  Or does it just not matter?

Issue #765, opened 2019-08-24
1.0.0 documentation additions and updates


kotfu:
This issue will track all the documentation updates and changes to be done in advance of releasing 1.0.0.

tleonhardt:
FYI, I merged *master* into the *doc_additions* branch (primarily to make sure we didn't have any conflicts)

tleonhardt:
@kotfu Would it be OK if we create a PR to merge in your changes on the **doc_additions** branch?  I'd like to get them in sooner rather than later.  I know we still have more doc improvements after that, but I'd rather get what we have in.

tleonhardt:
# Summary of documentation improvements needed
This comment is essentially a **TODO** list of documentation improvements which should be done.

## Missing Sections
These are sections of documentation which we haven't created a place for yet, but probably should:
* Something focused on piping and redirecting and the powers it provides
  * Make sure it shows complicated but useful examples like piping and redirecting in the same command and multiple pipes
* Something about recommended best practices when using `cmd2` with an overall application plugin framework and how to deal with things like sub-command arguments processed by a plugin
* More examples in the Examples section of the Sphinx docs

## Empty Sections
These are sections which exist, but don't contain any content:
* [Features/Initialization](https://cmd2.readthedocs.io/en/latest/features/initialization.html)
* [Features/Plugins](https://cmd2.readthedocs.io/en/latest/features/plugins.html)
* [Features/Startup Commands](https://cmd2.readthedocs.io/en/latest/features/startup_commands.html])

## Woefully Inadequate/Incomplete Sections
These are sections which exist and contain some minimal content, but the content they contain doesn't do `cmd2`'s feature set justice:
* [Features/Completion](https://cmd2.readthedocs.io/en/latest/features/completion.html#)
  * `cmd2`'s tab-completion capabilities are fantastic and extensive - what we have here doesn't even begin to scratch the surface
* [Features/Scripts/Python Scripts](https://cmd2.readthedocs.io/en/latest/features/scripting.html#python-scripts)

tleonhardt:
@kotfu I merged **master** into the **generating_output_docs** branch.  There were a few conflicts due to some ansi/style stuff getting renamed.  But I think it is in a good state.  It looks like you made a bunch of nice documentation updates there that we should get merged in sometime soon.

PR #799, opened 2019-11-02
Doc updates

tleonhardt:
This PR fills in a few sections of the Sphinx docs that were previously empty after the refactor.  It also corrects a couple pieces of documentation elsewhere which were out of date.

This partially addresses #765


docs/features/startup_commands.rst
```diff
@@ -1,2 +1,66 @@
...
+application runs like a non-interactive command-line utility (CLU).  This
+means that it can then be scripted from an external application and easily used
+in automation.
+
+.. note::
+
+   If you wish to disable cmd2's consumption of command-line arguments, you can
+   do so by setting the  ``allow_cli_args`` argument of your ``cmd2.Cmd`` class
+   instance to ``False``.  This would be useful, for example, if you wish to
+   use something like Argparse_ to parse the overall command line arguments for
```
kmvanbrunt: What's causing the underline and extra space?

docs/features/completion.rst
```diff
@@ -31,3 +31,50 @@ similar to the following to your class which inherits from ``cmd2.Cmd``::
 
     # Make sure you have an "import functools" somewhere at the top
     complete_bar = functools.partialmethod(cmd2.Cmd.path_complete, path_filter=os.path.isdir)
+
+
+Tab Completion Using Argparse Decorators
```
kmvanbrunt: This would be a good place to include most, if not all, of the header to [argparse_custom.py](https://github.com/python-cmd2/cmd2/blob/master/cmd2/argparse_custom.py).

The documentation there is pretty thorough. What do you think?

docs/features/startup_commands.rst
tleonhardt: That is reStructuredText formatting.  The underscore at the end means it is inserting a link defined elsewhere.  The extra indentation is for grouping what is in the Note.

docs/features/completion.rst
tleonhardt: I think we should include that either via an automodule or autoclass Sphinx Autodoc

tleonhardt: Unfortunately Sphinx autodoc is [very particular](http://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html) about formatting of docstrings for it to work.  

For this to work, the docstrings must be written in correct reStructuredText. You can then use all of the usual Sphinx markup in the docstrings, and it will end up correctly in the Sphinx documentation.

So to use ``automodule`` to import this documentation would require quite a bit of reformatting of that top-of-module docstring.

tleonhardt: @kmvanbrunt If you want to either fix the docstring so it works with **automodule** or manually put the equivalent documentation in the reStructuredText, feel free.  I don't have time to do this right now.  Alternatively, if you like we could create a new issue to do it so we don't forget about it.

tleonhardt: I created #800 to track this work

PR #804, opened 2019-11-11
Fix bug where cmd2 ran 'stty sane' command when stdin was not a terminal

### LINKED ISSUE #803: Getting "stty: stdin isn't a terminal" errors when piping input to my app
I have a ``cmd2`` based app which can accept commands piped via stdin, like so:
```
$ echo some_command | my_app
```

When used in this way, I get the following error:

"stty: stdin isn't a terminal"

I believe this error is caused by a bug in the ``_run_cmdfinalization_hooks()`` method of ``cmd2.Cmd``. If stdout is a tty, and we are not on windows, it runs ``stty sane``, which generates this error message.

kmvanbrunt:
Fixes #803

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 * Bug Fixes
     * Fixed bug where setting `use_ipython` to False removed ipy command from the entire `cmd2.Cmd` class instead of
     just the instance being created
+    * Fix bug where cmd2 ran 'stty sane' command when stdin was not a terminal
 * Enhancements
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 * Bug Fixes
     * Fixed bug where setting `use_ipython` to False removed ipy command from the entire `cmd2.Cmd` class instead of
     just the instance being created
+    * Fix bug where cmd2 ran 'stty sane' command when stdin was not a terminal
```
tleonhardt: Thanks for the quick fix and for updating changelog

Issue #802, opened 2019-11-10
Allow readline.__file__ to be missing


jayvdb:
See indygreg/PyOxidizer#69 for more info about why, but the tl;dr version is `__file__` is an optional attribute and should not be relied upon.

Currently cmd2 fails under PyOxidizer with:

```py
  File "cmd2", line 15, in <module>
  File "cmd2.cmd2", line 56, in <module>
  File "cmd2.rl_utils", line 120, in <module>
AttributeError: module 'readline' has no attribute '__file__'
```

If `sys.oxidized` is set, and `readline` is imported (and isnt libedit), then any symbols needed are in the binary, if that is helpful.

tleonhardt:
Getting `cmd2` to interoperate properly with [PyOxidizer](https://github.com/indygreg/PyOxidizer) would likely not be a trivial thing to do due to some of the *interesting* design choices made by `PyOxidizer`.  It seems like it would be possible, but honestly I don't see a way to do it reliably without moving to Python 3.8.  

Changes which would appear to be required include the following:
* Replace usage of `pkg_resources` in [__init__.py](https://github.com/python-cmd2/cmd2/blob/master/cmd2/__init__.py) since this module doesn't appear to be available within `PyOxidizer` applications (found by my testing)
    * We use this in combination with [setuptools_scm](https://github.com/pypa/setuptools_scm) for managing the `cmd2` package version
    * Python 3.8 appears to have an [importlib.metadata](https://docs.python.org/3/library/importlib.metadata.html) module which could probably fill the need, but that doesn't  exist prior to 3.8
    * Another option would be to re-design how `cmd2` versions are done, but this would be a significant refactoring which I don't believe we would like to undertake
* Replace usage of `__file__` in [rl_utils.py](https://github.com/python-cmd2/cmd2/blob/master/cmd2/rl_utils.py) as you mention
    * Here we just need to get the path to a C code dynamic library so we can load it with [ctypes](https://docs.python.org/3/library/ctypes.html), we do **NOT** want to open the file, nor do we want to create it
    * It looks like [importlib.resources](https://docs.python.org/3/library/importlib.html) is the recommended way to load "resource" files when using `PyOxidizer`, but I don't see any suitable APIs there for this
    * It does look like [importlib.util.find_spec()](https://docs.python.org/3/library/importlib.html#importlib.util.find_spec) would do what we want for finding the underlying dynamic library, but it is unclear to me if this would be properly supported within `PyOxidizer`

[...]

PR #807, opened 2019-11-15
Gracefully degrade when loading libreadline fails

jayvdb:
libreadline may not be loadable for many reasons, especially
in environments where a packager is using zipapp or PyOxidizer.
Avoid critical failure; degrade gracefully to operating without
libreadline and notify the user this has happened.

Related to https://github.com/python-cmd2/cmd2/issues/802

### DIFF
--- a/cmd2/rl_utils.py
+++ b/cmd2/rl_utils.py
@@ -28,6 +28,7 @@ class RlType(Enum):
 
 # Check what implementation of readline we are using
 rl_type = RlType.NONE
+readline_lib = None
 
 # Tells if the terminal we are running in supports vt100 control characters
 vt100_support = False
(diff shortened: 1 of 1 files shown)


cmd2/rl_utils.py
```diff
@@ -113,11 +114,19 @@ def pyreadline_remove_history_item(pos: int) -> None:
 elif 'gnureadline' in sys.modules or 'readline' in sys.modules:
     # We don't support libedit
     if 'libedit' not in readline.__doc__:
-        rl_type = RlType.GNU
-
         # Load the readline lib so we can access members of it
-        import ctypes
-        readline_lib = ctypes.CDLL(readline.__file__)
+        try:
+            import ctypes
```
kmvanbrunt: Why would an import of ctypes fail?

jayvdb: ctypes is a module.  It doesnt need to exist, and is one of the first to be disabled in embedded Python, especially on new arch/platforms where the entire gcc toolchain isnt available/stable yet.  In a completely static build of Python with all needed modules built in, there is no need for ctypes, and it adds bloat.

kmvanbrunt:
We are comfortable with cmd2 working in typical desktop environments running standard Python distributions as well as in containers like docker. If we choose to add exception handling to this code later, one of our authors will do it.

PR #810, opened 2019-11-17
cmd2-specific input() function

### LINKED ISSUE #778: Create cmd2-specific input() function
Provide developers a function for reading input with the following signature:

`def read_input(self, prompt: str, enable_completion: bool = False) -> str`

The name isn't set in stone.

Difference between this function and the existing `_pseudo_read_input`

1. Raises `EOFError` exceptions instead returning the string `eof`
2. Turns off tab completion by default

`_pseudo_read_input` will be replaced by something similar to the following:

```Python
def _read_command_line(self, prompt: str) -> str:
    try:
        # Wrap in try since terminal_lock may not be locked
        try:
            # Command line is about to be drawn. Allow asynchronous changes to the terminal.
            self.terminal_lock.release()
        except RuntimeError:
            pass
        return self.read_input(prompt, enable_completion=True)
    except EOFError:
        return 'eof'
    finally:
        # Command line is gone. Do not allow asynchronous changes to the terminal.
        self.terminal_lock.acquire()
```

kmvanbrunt:
Closes #778

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -2031,14 +2031,14 @@ class Cmd(cmd.Cmd):
                     # Deal with the vagaries of readline and ANSI escape codes
                     safe_prompt = rl_make_safe_prompt(prompt)
 
-                    # Check if tab completion should be disabled
                     with self.sigint_protection:
+                        # Check if tab completion should be disabled
                         if not allow_completion:
                             disable_completion()
                     line = input(safe_prompt)
                 finally:
-                    # Check if we need to reenable tab completion
                     with self.sigint_protection:
+                        # Check if we need to re-enable tab completion
                         if not allow_completion:
                             enable_completion()
             else:
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -416,6 +416,10 @@ class Cmd(cmd.Cmd):
             self.perror('Invalid value: {} (valid values: {}, {}, {})'.format(new_val, ansi.ANSI_TERMINAL,
                                                                               ansi.ANSI_ALWAYS, ansi.ANSI_NEVER))
 
+    def _completion_supported(self) -> bool:
+        """Return whether tab completion is supported"""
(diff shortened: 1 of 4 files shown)


cmd2/utils.py
```diff
@@ -520,10 +520,15 @@ def send_sigint(self) -> None:
...
         if sys.platform.startswith('win'):
-            signal_to_send = signal.CTRL_C_EVENT
+            self._proc.send_signal(signal.CTRL_C_EVENT)
         else:
-            signal_to_send = signal.SIGINT
-        self._proc.send_signal(signal_to_send)
+            # Since cmd2 uses shell=True in its Popen calls, we need to send the SIGINT to
+            # the whole process group to make sure it propagates further than the shell
+            try:
+                group_id = os.getpgid(self._proc.pid)
```
tleonhardt: Is `self._proc` and `self._proc.pid` guaranteed to exist?  Would it be better as the following?
```Python
group_id = os.getpgid(os.getpid())
```

I think the current process pid is guaranteed to exist, so you might not need to wrap it in the `try / except ProcessLookupError` if you do it the above way?

kmvanbrunt: `ProcReader._proc` is a `subprocess.Popen` object and is passed into `ProcReader` at initialization. This code is attempting to send a SIGINT to its process group, and not the current process' group. The `ProcessLookupError` is there because `os.getpgid()` and `os.killpg()` can fail if the subprocess has already ended.

cmd2/cmd2.py
```diff
@@ -1322,7 +1326,7 @@ def complete(self, text: str, state: int) -> Optional[str]:
         """
         # noinspection PyBroadException
         try:
-            if state == 0 and rl_type != RlType.NONE:
+            if state == 0:
```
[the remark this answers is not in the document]
kmvanbrunt: The check doesn't hurt, but it's unnecessary since only readline calls this function which means `rl_type` can't be `RlType.NONE`.

Issue #815, opened 2019-11-23
pwarning() should have an apply_style argument


kotfu:
`perror()` and `pexcept()` both have an `apply_style` argument which defaults to `True`. If that argument is `True`, then the method applies an ansi style to the message before sending it to the output stream. `pwarning()`` always applies the style. It should have an `apply_style` argument like it's sister methods.

kmvanbrunt:
`pwarning()` is a convenience wrapper around `perror()` for adding the warning style. Does it make sense to add the `apply_style` argument? If a user wants to print to stderr with no style, they can just call `perror()` with `apply_style` set to False.

tleonhardt:
I can see the advantage of adding the `apply_style` argument to `pwarning()` for consistency and in case the developer is programmatically setting `apply_style` to these calls.  So overall I think it might be a good idea.  I do see the point of view that it is just a convenience wrapper.  However it is a convenience wrapper with contextual meaning based on color of the output which is distinct from the default style/color of `perror`.

Regardless I don't feel particularly strongly one way or the other.

kmvanbrunt:
I can see a case where a user overrides `pwarning()` to prepend "WARNING" or something. In that case they may want an apply_style.

PR #816, opened 2019-11-23
Updated arguments to pfeedback and ppaged for consistency with other …

### LINKED ISSUE #814: pfeedback() needs similar arguments as poutput()
the `pfeedback()` method only accepts one argument, the string to output. All of our other output methods like `poutput()`, `perror()`, etc. take an `end` argument which defaults to `\n`. We should add this argument to `pfeedback()` too.

kmvanbrunt:
Closes #814 
Updated arguments to pfeedback and ppaged for consistency with other cmd2 print functions

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -530,12 +530,15 @@ class Cmd(cmd.Cmd):
 
         WARNING: On Windows, the text always wraps regardless of what the chop argument is set to
         """
-        import subprocess
-        if not msg:
+        # msg can be any type, so convert to string before checking if it's blank
+        msg_str = str(msg)
+
+        # Consider None to be no data to print
+        if msg is None or msg_str == '':
             return
 
         try:
-            msg_str = '{}{}'.format(msg, end)
+            import subprocess
 
             # Attempt to detect if we are not running within a fully functional terminal.
             # Don't try to use the pager when being run by a continuous integration system like Jenkins + pexpect.
@@ -550,6 +553,7 @@ class Cmd(cmd.Cmd):
             if functional_terminal and not self._redirecting and not self.in_pyscript() and not self.in_script():
                 if ansi.allow_ansi.lower() == ansi.ANSI_NEVER.lower():
                     msg_str = ansi.strip_ansi(msg_str)
+                msg_str += end
 
                 pager = self.pager
                 if chop:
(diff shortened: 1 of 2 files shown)


cmd2/cmd2.py
```diff
@@ -528,43 +531,43 @@ def ppaged(self, msg: str, end: str = '\n', chop: bool = False) -> None:
...
+                if chop:
+                    pager = self.pager_chop
+
+                # Prevent KeyboardInterrupts while in the pager. The pager application will
+                # still receive the SIGINT since it is in the same process group as us.
+                with self.sigint_protection:
+                    pipe_proc = subprocess.Popen(pager, shell=True, stdin=subprocess.PIPE)
+                    pipe_proc.communicate(msg_str.encode('utf-8', 'replace'))
+            else:
+                self.poutput(msg_str, end='')
```
tleonhardt: Should this be `end=end`?

cmd2/cmd2.py
```diff
@@ -502,23 +502,26 @@ def pexcept(self, msg: Any, *, end: str = '\n', apply_style: bool = True) -> Non
         # Set apply_style to False since style has already been applied
```
tleonhardt: Should `pwarning()` have an `apply_style` optional argument added to it to be consistent with `perror()` and `pexcept()`?

cmd2/cmd2.py
kmvanbrunt: Good catch. I wonder how long it's been like that.

cmd2/cmd2.py
kmvanbrunt: See discussion on #815 for this question.

cmd2/cmd2.py
```diff
@@ -527,44 +530,48 @@ def ppaged(self, msg: str, end: str = '\n', chop: bool = False) -> None:
...
-                else:
-                    self.poutput(msg_str, end='')
-            except BrokenPipeError:
-                # This occurs if a command's output is being piped to another process and that process closes before the
-                # command is finished. If you would like your application to print a warning message, then set the
-                # broken_pipe_warning attribute to the message you want printed.`
-                if self.broken_pipe_warning:
-                    sys.stderr.write(self.broken_pipe_warning)
+        # msg can be any type, so convert to string before checking if it's blank
+        msg_str = str(msg)
```
tleonhardt: I think this is fine, but would there ever be a reasonable case in which `msg` wouldn't be convertible to a `str` and this would throw an Exception?

PR #817, opened 2019-11-23
Added apply_style to pwarning()

### LINKED ISSUE #815: pwarning() should have an apply_style argument

kmvanbrunt:
Closes #815

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -467,13 +467,17 @@ class Cmd(cmd.Cmd):
             final_msg = "{}".format(msg)
         ansi.ansi_aware_write(sys.stderr, final_msg + end)
 
-    def pwarning(self, msg: Any, *, end: str = '\n') -> None:
-        """Apply the warning style to a message and print it to sys.stderr
+    def pwarning(self, msg: Any, *, end: str = '\n', apply_style: bool = True) -> None:
+        """Like perror, but applies ansi.style_warning by default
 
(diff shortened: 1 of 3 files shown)


cmd2/cmd2.py
```diff
@@ -467,13 +467,17 @@ def perror(msg: Any, *, end: str = '\n', apply_style: bool = True) -> None:
...
 
         :param msg: message to print (anything convertible to a str with '{}'.format() is OK)
         :param end: string appended after the end of the message, default a newline
+        :param apply_style: If True, then ansi.style_warning will be applied to the message text. Set to False in cases
+                            where the message text already has the desired style. Defaults to True.
         """
-        self.perror(ansi.style_warning(msg), end=end, apply_style=False)
+        if apply_style:
+            msg = ansi.style_warning(msg)
+        self.perror(msg, end=end, apply_style=False)
```
tleonhardt: This looks fine, but it makes me wonder if it would be better to refactor `perror` to accept an additional optional keyword-only argument called something like `style` and then have `pwarning` call `perror` and pass on `apply_style` and set `style = ansi.syle_warning(msg)`?

PR #826, opened 2019-12-02
Printer changes

### LINKED ISSUE #824: Why is perror() a static method?
All of the other output methods (`pfeedback()`, `pwarning()`), are regular methods on `cmd.Cmd2`. Why is `perror()` a static method?

kmvanbrunt:
Closes #824 

- Added `self` to `perror()` for consistency and ease of overriding
- Gave the `msg` argument of printing functions (`poutput`, `perror`...) a default value. This allows calls with no args (e.g. `poutput()`)
- `ppaged()` will now produce output even if `msg` is blank or `None`


a
```diff
@@ -0,0 +1,2 @@
+
```
tleonhardt: Is this some extra file which should not be committed?

cmd2/cmd2.py
```diff
@@ -530,13 +530,6 @@ def ppaged(self, msg: Any, *, end: str = '\n', chop: bool = False) -> None:
 
         WARNING: On Windows, the text always wraps regardless of what the chop argument is set to
         """
-        # msg can be any type, so convert to string before checking if it's blank
-        msg_str = str(msg)
```
tleonhardt: Does this significantly alter behavior for any realistic use case?

cmd2/cmd2.py
```diff
@@ -428,7 +428,7 @@ def visible_prompt(self) -> str:
         """
         return ansi.strip_ansi(self.prompt)
 
-    def poutput(self, msg: Any, *, end: str = '\n') -> None:
+    def poutput(self, msg: Any = '', *, end: str = '\n') -> None:
```
tleonhardt: Should we add something to change log for this?

PR #829, opened 2019-12-06
Enabled line buffering when redirecting output to a file

kmvanbrunt:
This makes tailing a file that's being outputted to more responsive

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -1881,7 +1881,6 @@ class Cmd(cmd.Cmd):
                     mode = 'a'
                 else:
                     mode = 'w'
-                mode += 't'
 
                 try:
                     # Use line buffering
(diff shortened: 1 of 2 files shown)


cmd2/cmd2.py
```diff
@@ -1874,22 +1874,26 @@ def _redirect_output(self, statement: Statement) -> Tuple[bool, utils.Redirectio
...
-                # going to a file
-                mode = 'w'
-                # statement.output can only contain
-                # REDIRECTION_APPEND or REDIRECTION_OUTPUT
+                # statement.output can only contain REDIRECTION_APPEND or REDIRECTION_OUTPUT
                 if statement.output == constants.REDIRECTION_APPEND:
                     mode = 'a'
+                else:
+                    mode = 'w'
+                mode += 't'
```
tleonhardt: I thought text mode is the default and thus that this is redundant but explicit?

kmvanbrunt: You are correct, it is redundant.  I removed it.

PR #832, opened 2019-12-10
Took out rstrip() calls that are no longer neeeded

kmvanbrunt:
Removed some calls to `rstrip()` that I don't think are needed.
@kotfu Is this change correct

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -3358,7 +3358,7 @@ class Cmd(cmd.Cmd):
             with os.fdopen(fd, 'w') as fobj:
                 for command in history:
                     if command.statement.multiline_command:
-                        fobj.write('{}\n'.format(command.expanded.rstrip()))
+                        fobj.write('{}\n'.format(command.expanded))
                     else:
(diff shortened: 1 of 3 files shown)


cmd2/cmd2.py
```diff
@@ -3358,7 +3358,7 @@ def do_history(self, args: argparse.Namespace) -> Optional[bool]:
             with os.fdopen(fd, 'w') as fobj:
                 for command in history:
                     if command.statement.multiline_command:
-                        fobj.write('{}\n'.format(command.expanded.rstrip()))
+                        fobj.write('{}\n'.format(command.expanded))
```
tleonhardt: Are these no longer needed because expanded already does an rstrip() internally?

kmvanbrunt: Aside from `Statement.raw`, its other members are built using tokens produced by shlex. Therefore all leading/trailing whitespace is gone.

The code changed in this Pull Request is referencing `HistoryItem.expanded` which just returns `Statement.expanded_command_line`. So no whitespace needs to be stripped.

PR #839, opened 2020-01-06
Add missing setuptools dependency

### LINKED ISSUE #837: Implicit dependency on setuptools is missing
`cmd2` has an implicit runtime dependency on `setuptools` due to use of `setuptools_scm` for versioning.  However, we don't explicitly call out this dependency in **setup.py**.  This could present a problem on any system which doesn't have `setuptools` installed.

tleonhardt:
`cmd2` has an implicit runtime dependency on [setuptools](https://pypi.org/project/setuptools/) picked up via the build dependency on [setuptools_scm](https://pypi.org/project/setuptools-scm/).  This is because the `cmd2/__init__.py` file imports from **pkg_resources**.  The [pkg_resources](https://setuptools.readthedocs.io/en/latest/pkg_resources.html) module distributed with `setuptools` provides an API for Python libraries to access their resource files.

This PR adds an explicit dependency on `setuptools` to prevent `cmd2` not working in any environments which don't happen to have `setuptools` already installed.  Additionally, it adds minimum required versions for both `setuptools` and `setuptools_scm` as follows:
* setuptools_scm >= 3.0.0
* setuptools >= 34.4

Closes #837

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 0.9.23 (TBD, 2019)
 * Bug Fixes
     * Fixed bug where startup script containing a single quote in its file name was incorrectly quoted
+    * Added missing implicit dependency on `setuptools` due to build with `setuptools_scm`
 
 ## 0.9.22 (December 9, 2019)
 * Bug Fixes
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,6 +51,7 @@ The tables below list all prerequisites along with the minimum required version 
 | [attrs](https://github.com/python-attrs/attrs)      | `16.3`          |
 | [colorama](https://github.com/tartley/colorama)     | `0.3.7`         |
 | [pyperclip](https://github.com/asweigart/pyperclip) | `1.6`           |
+| [setuptools](https://pypi.org/project/setuptools/)  | `34.4`          |
 | [wcwidth](https://pypi.python.org/pypi/wcwidth)     | `0.1.7`         |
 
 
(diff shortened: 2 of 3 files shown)


kotfu:
Agree with everything in this PR and the solution, but have two questions:

1. I had a quick look at `setuptools_scm` source code and it doesn't look like they declare a dependency on `setuptools`. Which seems curious to me.
2. `pip` installs `setuptools`, so it would be a pretty odd situation to have `cmd2` installed without `pip` and `setuptools`. Did you observe/discover a situation where `cmd2` was broken because `setuptools` was not installed?

tleonhardt:
@kotfu
1. Weird.  I'll look into that a little more and maybe post an issue to `setuptools_scm` 
2. Yes I encountered a very peculiar situation where a large enterprise has their own way of installing Python packages which doesn't involve `pip` (nominally for security reasons I believe).

tleonhardt:
@kotfu 
`setuptools_scm` declares a dependency on `setuptools` in it's [pyproject.toml](https://github.com/pypa/setuptools_scm/blob/master/pyproject.toml)

kotfu:
@tleonhardt 
Ahhh, I didn't look there. If that's the case, and if we declare a dependency on `setuptools_scm`, why didn't `setuptools` get installed?

kotfu:
Umm...answered my own question. They define a build-time dependency on `setuptools`, but not a run-time dependency.

tleonhardt:
We had previously declared a **build** dependency on `setuptools_scm`, not a **runtime** one:
```sh
SETUP_REQUIRES = ['setuptools_scm >= 3.0.0']
INSTALL_REQUIRES = ['attrs >= 16.3.0', 'colorama >= 0.3.7', 'pyperclip >= 1.6', 'setuptools >= 34.4', 'wcwidth >= 0.1.7']
```

PR #852, opened 2020-01-22
Import ABC from collections.abc instead of collections for Python 3.9 compatibility

### LINKED ISSUE #851: Import ABC from collections.abc instead of collections for Python 3.9 compatibility.
Import ABC from collections.abc instead of collections for Python 3.9 compatibility.

https://github.com/python-cmd2/cmd2/blob/e4bc87da34bb6fc645f54a4467dc7d0e32dfaa3b/cmd2/utils.py#L60

tirkarthi:
Fixes #851

### DIFF
--- a/cmd2/utils.py
+++ b/cmd2/utils.py
@@ -9,6 +9,7 @@ import subprocess
 import sys
 import threading
 import unicodedata
+import collections.abc as collections_abc
 from enum import Enum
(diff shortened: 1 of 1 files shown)


tirkarthi:
This seems to affect dependencies like cliff as reported in https://bugzilla.redhat.com/show_bug.cgi?id=1793645

PR #864, opened 2020-01-28
Reduce settable attributes and renamed locals_in_py

### LINKED ISSUE #861: Reconsider which attributes should be settable by default
Anything which is settable by using the **set** command can easily be changed at runtime by end users.

I'm wondering if we might provide a better end-user experience by leaving these attributes out of the `settable` dictionary by default:

- continuation_prompt
- locals_in_py
- prompt

My reasoning is that for `prompt` and `continuation_prompt` - altering the changes the look and feel of an application and also might possibly have unintended consequences.    While for `locals_in_py`, developers might not want their end users to be able to easily enable this at runtime.

tleonhardt:
Renamed `locals_in_py` to `self_in_py` to more accurately reflect what it controls and made the following public attributes of `cmd2.Cmd` no longer settable at runtime by end users:
- continuation_prompt
- self_in_py
- prompt

This closes #861


CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## 0.9.26 (January 26, 2020)
```
kmvanbrunt: Change date to TBD

cmd2/cmd2.py
```diff
@@ -181,28 +181,25 @@ def __init__(self, completekey: str = 'tab', stdin=None, stdout=None, *,
         super().__init__(completekey=completekey, stdin=stdin, stdout=stdout)
 
         # Attributes which should NOT be dynamically settable via the set command at runtime
-        # To prevent a user from altering these with the py/ipy commands, remove locals_in_py from the
+        # To prevent a user from altering these with the py/ipy commands, remove self_in_py from the
```
kmvanbrunt: This comment no longer applies

kotfu:
I think the changelog notes are misleading for what was changed. The changelog makes it sound like users can not modify those settings. That isn't quite true. If the developer added `continuation_prompt` back into the `settable` array, then end users could totally modify the setting. What we have done with this change is made these three settings only available to developers **in the default configuration**.

tleonhardt:
@kotfu Good call.  How would you recommend we reword the changelog notes to make it more clear?

Issue #871, opened 2020-01-31
Tab complete settable attribute values


bambu:
It would be nice to have the ability to tab complete settable attribute values. The use case i have is there is a setting that takes a string and I would like the user to complete from a generated list of available options.

While I could create a custom command for this and have the completion that way, it seems kinda hackish and I would have to reimplement the same thing for any other future settings where the answer is not a True/False, numeric value, or arbitrary string.

tleonhardt:
@bambu Thanks for the suggestion.  I agree that this would be nice.  This isn't a trivial change, but should be doable.  We will have to discuss the design for the best way to do this.

If you have a good idea for how to implement this in a maintainable fashion, then please feel free to submit a PR.

kotfu:
I have many ideas for improvements to the settings in `cmd2` and this is another great addition to the list, but I think this should probably wait for after 1.0.

tleonhardt:
The only reason I was thinking of doing this before 1.0 is that there is a high probability it might involve changes which break existing `cmd2` applications that alter `self.settable`.

kotfu:
@tleonhardt I agree with you that I don't think this change can be done without breaking existing apps which alter `self.settable`. So the question becomes, are we going to add yet another breaking change before 1.0, or will we postpone this for after 1.0?

kmvanbrunt:
@tleonhardt @kotfu
The last couple of years have seen these major enhancements (forgive me if I forgot some):

- @tleonhardt got it working in Python 3 and has made numerous enhancements and bug fixes
- @kotfu rewrote the parser and added `argparse` integration
- I worked on tab completion
- @anselor integrated tab completion and `argparse` via the `AutoCompleter`
- We worked together to rewrite history and pyscript functionatiliy

Since @tleonhardt became maintainer, a very talented core group has come together to give **cmd2** an amazing overhaul. The settable feature is really the only thing we haven't touched. I say why not just do it before 1.0? Otherwise we risk maintaining a 1.0 and 2.0 branch very early into 1.0's lifetime.

With that, I have an idea that #870 and this issue could be solved by using `argparse`. It already has validation for basic types and it's easy to add custom validation functions. Additionally, **cmd2** added tab completion to it. I don't have all of it fleshed out, but I'm sure it's doable.

Thoughts?

tleonhardt:
[...]

PR #872, opened 2020-01-31
Settable attributes with original value `None` can be set to a str

### LINKED ISSUE #870: setting settable attribute with default type None does not change
It appears that when a `settable` attribute is set that a call to `utils.cast` is made and attempts to cast the new `value` to the type of `orig_value`. When the original value is `None` the result of that value is also `None` and, in my case, gets passed to my setter as `None`.

While I understand that this could easily resolved by setting the default value of my attribute to '', it visually looks odd when every other settable is filled besides the one that was not set.

``` text
...
allow_style: Terminal
att: 
continuation_prompt: > 
...
```

Likewise, setting the default value to `'None'` does not seem like a good answer to resolving the issue. 

In this particular case, I think I would prefer if the original value is 'None' then the str that was passed is set

tleonhardt:
Settable attributes with original value `None` can now be set to a string

Also:
- Moved unit tests of cast utility function from test_cmd2.py to test_utils.py

Closes #870

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 0.9.26 (TBD, 2020)
 * Enhancements
     * Changed the default help text to make `help -v` more discoverable
+    * Settable attributes with original value `None` can now be set to a string
 * Breaking changes
     * Renamed `locals_in_py` attribute of `cmd2.Cmd` to `self_in_py`
     * The following public attributes of `cmd2.Cmd` are no longer settable at runtime by default:
--- a/cmd2/utils.py
+++ b/cmd2/utils.py
@@ -93,8 +93,12 @@ def cast(current: Any, new: str) -> Any:
 
     :param current: current value for the parameter, type varies
     :param new: new value
-    :return: new value with same type as current, or the current value if there was an error casting
+    :return: new value with same type as current, or the current value if there was an error casting or original was None
     """
+    # Prevent an attempt at casting to NoneType
+    if current is None:
+        return new
+
     typ = type(current)
     orig_new = new
 
--- a/tests/test_cmd2.py
+++ b/tests/test_cmd2.py
@@ -118,45 +118,6 @@ def test_base_show_readonly(base_app):
     assert out == expected
 
 
-def test_cast():
-    # Boolean
-    assert utils.cast(True, True) == True
-    assert utils.cast(True, False) == False
-    assert utils.cast(True, 0) == False
-    assert utils.cast(True, 1) == True
-    assert utils.cast(True, 'on') == True
-    assert utils.cast(True, 'off') == False
-    assert utils.cast(True, 'ON') == True
-    assert utils.cast(True, 'OFF') == False
-    assert utils.cast(True, 'y') == True
-    assert utils.cast(True, 'n') == False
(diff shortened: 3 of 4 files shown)


kotfu:
What if the original issue by @bambu asked about changing a default settable of `None` to a number? Maybe there is a use case for doing so. It seems like we ought to accommodate both scenarios. I see three options:

1. Change the casting code so a value of `None` can be changed to any other type (str, int, list, etc)
2. Add documentation that says if you want an "no value" settable that can only be changed to a string, make it be an empty string or `None`; if you want it to be only changed to an integer, make it be `0`, if you want it only changable to a float make it be `0.0`, and if you want it only changeable to a list make it`[]`.
3. Add documentation describing how to do setting validation by implementing the setting as a python property. We may need to update the `do_set` method so it gracefully handles a thrown `ValueError`, which I think would be the right exception to raise from the property method if user input did not meet requirements.

Maybe we should do both 2 and 3.

tleonhardt:
@kotfu I was considering combining a more general solution in when we address the 2nd issue submitted by @bambu #871. 

For that one I was thinking of expanding the values stored in the `self.settable` dictionary to be more than just a string description.  Tentatively I was thinking of making it optionally a tuple where the first element would be the string description but then there would be optional 2nd and 3rd elements where the 2nd element if present would be a recommend type to use when casting and the 3rd element if present would be values and/or a function for tab-completing values when calling **set** on this attribute.  What do you think of this approach?

The problem with solution 1 above is that we don't have enough information right now in regards to expect type to know what to try to cast it to.  Approach 2 seems reasonable - whatever we do we should document it.  Approach 3 could also work and seems like the most standard OO approach to this.

bambu:
FWIW,

The `ValueError` handling in 3 sounds like a good idea and would be a better implementation than reporting an error and leaving the value as is in the setter. 

When I wrote the issue, I was originally thinking of the solution being similar to what @tleonhardt just proposed. However, did not suggest it as it could break compatibility. Instead of a `Tuple` I might suggest a `dataclass` or similar to allow for growth. Having it set to a `Tuple` could cause ordering issues and may get difficult to manage if you end up adding more features in the future.

Roughly something like this:
```python
@dataclass
class Settable:
    desc: str
    type: Optional[Callable[[str], _T]] = None
    complete: Optional[Union[Iterable[Any], Callable[[..], Iterable[..] = None 

# in the Cmd instance
self.settable['sunny'] = 'Is it sunny outside?'
self.settable['food'] = Settable(
    desc='food item',
    complete=['Pizza', 'Ham', 'Ham Sandwich', 'Potato']
)
```

tleonhardt:
@bambu It is important to us to support multiple versions of Python - generally all that are currently officially supported or not end-of-life.  Currently we support Python 3.5+.  I don't believe dataclasses were added until Python 3.7, otherwise they would be an excellent solution.

Perhaps we could use a NamedTuple with default values as a poor man's dataclass.

bambu:
I completely understand. The dataclass was just an example. NamedTuple with defaults works and looks like Cmd2 already has an implementation of it to account for defaults not being available pre3.7.

tleonhardt:
I'm going to close this PR without merging because I believe a better solution is in the works.

PR #873, opened 2020-02-05
Updated set command to support tab completion of values

### LINKED ISSUE #870: setting settable attribute with default type None does not change

kmvanbrunt:
Closes #870 #871

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 0.9.26 (TBD, 2020)
 * Enhancements
     * Changed the default help text to make `help -v` more discoverable
+    * Added `add_settable()` and `remove_settable()` convenience methods to update `self.settable` dictionary
 * Breaking changes
     * Renamed `locals_in_py` attribute of `cmd2.Cmd` to `self_in_py`
     * The following public attributes of `cmd2.Cmd` are no longer settable at runtime by default:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
         * `continuation_prompt`
         * `self_in_py`
         * `prompt`
+    * `self.settable` changed to `self.settables`
+        * It is now a Dict[str, Settable] instead of Dict[str, str]
+        * **set** command now supports tab-completion of values
+    * Removed `cast()` utility function
     
 ## 0.9.25 (January 26, 2020)
 * Enhancements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.9.26 (TBD, 2020)
+## 1.0.0-rc1 (TBD, 2020)
 * Enhancements
     * Changed the default help text to make `help -v` more discoverable
     * Added `add_settable()` and `remove_settable()` convenience methods to update `self.settable` dictionary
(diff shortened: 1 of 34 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
         * `continuation_prompt`
         * `self_in_py`
         * `prompt`
+    * `self.settable` changed to `self.settables`
```
tleonhardt: We should probably add something under `Enhancements` about the `add_settable()` and `remove_settable()` methods

kotfu:
This all looks pretty good to me. I'll add one suggestion.

If the list of valid values is not known at initialization time (which is when settables would typically get defined), you have to do really awkward stuff in order to validate the new value of a setting when a user changes it. The ``onchange_cb`` doesn't have access to `self` (i.e. the `cmd2` subclass), so my callback doesn't have access to all the data in my app.

If I somehow manage to say the input isn't valid, there is no way for my callback to to say "hey this isn't a valid value, I don't want to allow the user to change it" back to `cmd2`. I think throwing a `ValueError` exception would be ideal here, and `cmd2` could then catch it and not allow the change in the value.

docs/features/settings.rst
```diff
@@ -165,4 +165,4 @@ the :ref:`features/settings:debug` setting. To do so, remove it from the
 
     def __init__(self):
         super().__init__()
-        self.settable.pop('debug')
+        self.settables.pop('debug')
```
tleonhardt: This should be updated to use the new `self.remove_settable()` method

examples/remove_settable.py
```diff
@@ -0,0 +1,19 @@
...
+A sample application for cmd2 demonstrating how to remove one of the built-in runtime settable parameters.
+"""
+import cmd2
+
+
+class MyApp(cmd2.Cmd):
+
+    def __init__(self):
+        super().__init__()
+        self.settables.pop('debug')
```
tleonhardt: This should be updated to use the new `self.remove_settable()` method

kmvanbrunt:
@kotfu The `onchange_cb` can have access to `self` if you set it to a bound method of your app.

```Python
onchange_cb = self.val_changed
```

As far as validation, that's the job of `Settable.val_type`. It corresponds to the `type` field in the `argparse.add_argument()` function. Any exception it raises will be sent to `perror()` and the value will not be changed.

kotfu:
Go it. I missed that `val_type` was a callable. You already have implemented my suggestion. This looks good to me.

kmvanbrunt:
@kotfu I update the docstring for `Settable` to be more clear.

kotfu:
hold up before we merge this, I need to fix the settings.rst documentation and update examples/environment.py

docs/conf.py
```diff
@@ -25,9 +25,6 @@
 # Import for custom theme from Read the Docs
 import sphinx_rtd_theme
 
-sys.path.insert(0, os.path.abspath('..'))
```
tleonhardt: I think you need to delete the import of `os` now that it is unused to make `flake8` happy

CHANGELOG.md
```diff
@@ -1,13 +1,20 @@
-## 0.9.26 (TBD, 2020)
+## 1.0.0-rc1 (TBD, 2020)
```
tleonhardt: Should we maybe go with something like `0.99.0` instead to technically stick within SemVer format?

kotfu: By my interpretation of https://semver.org/#spec-item-9 it is a valid SemVer format.

tleonhardt: Ok cool, I wasn't aware of that.  Sounds good.

tleonhardt:
This looks good so I'm going to merge it

PR #877, opened 2020-02-07
Changelog fix for removed ansi.FG_COLORS and ansi.BG_COLORS

tleonhardt:
Updated changelog to address removal of `ansi.FG_COLORS` and `ansi.BG_COLORS` and mention their replacement by `ansi.fg` and `ansi.bg` enums

Also:
- Use `ansi.fg` in **async_printing.py** and **README.md**
- The fact that the set command now supports tab-completion of values is an enhancement, not a breaking change

This addresses a deficiency called out in a review of  #876 after it merged.

If we absolutely didn't want to remove these, they could be easily generated from the enums like so:
```Python
FG_COLORS = {c.name: c.value for c in fg}
```

### DIFF
--- a/cmd2/ansi.py
+++ b/cmd2/ansi.py
@@ -62,7 +62,7 @@ class fg(Enum):
     @staticmethod
     def get_value(name: str) -> str:
         """Retrieve color code by name string."""
-        return fg.__members__[name].value
+        return fg[name].value
 
 
 # Background colors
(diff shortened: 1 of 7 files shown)


tleonhardt:
We should create some unit tests for the newly added functionality, but other than that this looks good.

cmd2/ansi.py
```diff
@@ -27,11 +27,47 @@
 ANSI_STYLE_RE = re.compile(r'\x1b\[[^m]*m')
 
 
+class ColorBase(Enum):
```
tleonhardt: I like how you extracted the common functionality to an Enum base class and also added overrides for the `+` operator.

PR #882, opened 2020-02-10
do_set fix

kmvanbrunt:
Corrected issue where the actual new value was not always being printed in do_set. This occurred in cases where the typed value differed from what the setter had converted it to.


cmd2/cmd2.py
```diff
@@ -2886,8 +2885,8 @@ def do_set(self, args: argparse.Namespace) -> None:
                 # Try to update the settable's value
                 try:
                     orig_value = getattr(self, args.param)
-                    new_value = settable.val_type(args.value)
-                    setattr(self, args.param, new_value)
+                    setattr(self, args.param, settable.val_type(args.value))
+                    new_value = getattr(self, args.param)
```
tleonhardt: Good catch!

PR #886, opened 2020-02-11
truncate_line not handling style sequences

kmvanbrunt:
Fixed bug where ANSI style sequences were not correctly handled in utils.truncate_line()

### DIFF
--- a/cmd2/constants.py
+++ b/cmd2/constants.py
@@ -14,6 +14,7 @@ COMMENT_CHAR = '#'
 MULTILINE_TERMINATOR = ';'
 
 LINE_FEED = '\n'
+HORIZONTAL_ELLIPSIS = '\N{HORIZONTAL ELLIPSIS}'
 
 DEFAULT_SHORTCUTS = {'?': 'help', '!': 'shell', '@': 'run_script', '@@': '_relative_run_script'}
 
--- a/cmd2/constants.py
+++ b/cmd2/constants.py
@@ -14,7 +14,9 @@ COMMENT_CHAR = '#'
 MULTILINE_TERMINATOR = ';'
 
 LINE_FEED = '\n'
-HORIZONTAL_ELLIPSIS = '\N{HORIZONTAL ELLIPSIS}'
+
+# One character ellipsis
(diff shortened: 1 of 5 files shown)


cmd2/utils.py
```diff
@@ -11,7 +11,7 @@
 import threading
 import unicodedata
 from enum import Enum
-from typing import Any, Callable, Iterable, List, Optional, TextIO, Union
+from typing import Any, Callable, Iterable, List, OrderedDict, Optional, TextIO, Union
```
tleonhardt: `OrderedDict` was not added to the `typing` module until [Python 3.7.2](https://docs.python.org/3/library/typing.html)

You should be able to directly use [collections.OrderedDict](https://docs.python.org/2/library/collections.html#collections.OrderedDict) in type hints.

This is causing failures on all 3 CI systems.  This is a must fix.

tleonhardt:
The python 3.9-dev failures on TravisCI are not our fault.  They are a temporary problem with Travis due to a bug in Virtualenv which has since been fixed: https://travis-ci.community/t/python-development-versions-no-longer-include-pip-due-to-virtualenv-20-x/7180/9

cmd2/constants.py
```diff
@@ -14,6 +14,7 @@
 MULTILINE_TERMINATOR = ';'
 
 LINE_FEED = '\n'
+HORIZONTAL_ELLIPSIS = '\N{HORIZONTAL ELLIPSIS}'
```
tleonhardt: Can you explain how this works?  Like is it filling in an ellipsis `...`?  If so where is that coming from?

kmvanbrunt: It's an escape sequence for the one-character ellipsis. (…) It's easier type than the real thing, but I will alter the code to make it clearer.

cmd2/constants.py
```diff
@@ -15,6 +15,9 @@
 
 LINE_FEED = '\n'
 
+# One character ellipsis
+HORIZONTAL_ELLIPSIS = '…'
```
tleonhardt: Ok, this makes more sense to me now ;-)

PR #887, opened 2020-02-13
Completion docs

kmvanbrunt:
Updated the tab completion examples


CHANGELOG.md
```diff
@@ -9,7 +9,7 @@
 ## 0.10.0 (February 7, 2020)
 * Enhancements
     * Changed the default help text to make `help -v` more discoverable
-    * **set** command now supports tab-completion of values
+    * **set** command now supports tab completion of values
```
tleonhardt: So are we now standardizing on "tab completion" instead of "tab-completion"?  Works for me.

examples/argparse_completion.py
```diff
@@ -0,0 +1,122 @@
+#!/usr/bin/env python
+# coding=utf-8
+"""
+A simple example demonstrating how to integrate tab completion with argparse-based commands.
```
tleonhardt: Thanks for creating this much simpler and easier to grok tab completion example

examples/tab_autocompletion.py
```diff
@@ -1,268 +0,0 @@
-#!/usr/bin/env python3
```
tleonhardt: I'd recommend double-checking that you removed all references to these two deleted files.

PR #899, opened 2020-02-23
API documentation

### LINKED ISSUE #880: API Documentation Approach
I've been working on the API documentation. We have lots of constants, functions, and classes which we use internally and which are not really intended to be part of the public API. For example, in `constants.py` the `COMMAND_FUNC_PREFIX` constant is only for internal use.

Somewhere/somehow we need to indicate whether this item is part of the public API. We could:

1. Go through a renaming exercise and rename everything that isn't public to start with an underscore.
2. Declare that the documentation, not the source code, is the authoritative declaration of the public api.
2a. Don't use the `:members:` Sphinx directive; it will include everything that doesn't start with an underscore. Instead, individually add methods, functions, and constants that are part of the public API to the documentation. This is how `decorators.py` is built now. It's OK when there are only a handful of items, but it would be quite unwieldy for `cmd2.Cmd`, which has nearly a hundred methods.
2b. I haven't experimented with this, but we could use something like https://public.readthedocs.io/en/latest/README.html, a decorator which magically populates `__all__`, and which it appears that the Sphinx `:members:` directive honors.
2c. Wait for Sphinx 3.0 to be released, which includes a `:meta private:` info field. By adding this to the docstring of a method, you can tell Sphinx that it's private and it will be excluded from `:members:`.

The nice thing about 1 is that it's abundantly clear from the source code what's public and what isn't. This makes it more visible when modifying or enhancing so a developer can be more thoughtful about breaking changes. The downside to renaming everything is that the potential for breaking downstream consumers is high.

If we choose the documentation to be the definition of the public API, then we have to manually include or exclude items when creating and maintaining the documentation.

If there is single function in a module that we don't want to be public, and which doesn't begin with an underscore, we can't use the sphinx `:members:` directive to generate the documentation, we will have to list all the functions out one by one. For example, I could argue that the `cmd_func()` method in the `cmd2.Cmd` class should not be part of the public API. If we decided that it shouldn't be, then when building and maintaining the documentation for the `cmd2.Cmd` class you have to manually create a list of the approx 100 other methods in that class instead of just putting `:members:`

This seems like a bit of work, but doable. To reduce that work we could experiment with some sort of `@public` decorator, or wait for Sphinx 3.0 (who know how long that will be) and exclude things using the private info field in the docstrings.

Any preferences on what to do here?

### LINKED ISSUE #896: Should we implement hook methods from `cmd.Cmd` so we can document them?
We are not consistent in defining and documenting methods from `cmd.Cmd` which we use/rely-on, but which perform the same function as they do in our superclass. For example, `cmd2.Cmd` provides an implementation of the `precmd()` method from our superclass, which does exactly the same thing as the superclass method. We do not implement `postcmd()` or `preloop()` or `postloop()`, which are no-ops in the superclass, i.e. they are stub methods waiting to be over-ridden.

Providing an implementation of these methods allows us to reference them in our documentation, which is nice.

I don't care whether we do or don't provide an implementation for these methods, but whichever we choose, we ought to do it the same for all four of these hook-type methods.

Any preference?

kotfu:
Fixes #896 and fixes #880.

All methods, functions, classes, constants, and attributes that are part of the public API for this project should now be documented in `docs/api`. I have made some judgement calls on certain functions and classes to be included or excluded. We need to validate and confirm these decisions.


cmd2/argparse_custom.py
```diff
@@ -1,127 +1,150 @@
 # coding=utf-8
 """
-This module adds capabilities to argparse by patching a few of its functions. It also defines a parser
-class called Cmd2ArgumentParser which improves error and help output over normal argparse. All cmd2 code uses
-this parser and it is recommended that developers of cmd2-based apps either use it or write their own parser
-that inherits from it. This will give a consistent look-and-feel between the help/error output of built-in
-cmd2 commands and the app-specific commands. If you wish to override the parser used by cmd2's built-in
-commands, see override_parser.py example.
+This module adds capabilities to argparse by patching a few of its functions.
```
tleonhardt: Fixing this so that it can be rendered by Sphinx is a vast improvement.  Nice work!

cmd2/cmd2.py
```diff
@@ -1508,13 +1523,54 @@ def sigint_handler(self, signum: int, frame) -> None:
...
 
     def precmd(self, statement: Statement) -> Statement:
-        """Hook method executed just before the command is processed by ``onecmd()`` and after adding it to the history.
+        """Hook method executed just before the command is executed by
+        :meth:`~cmd2.Cmd.onecmd` and after adding it to history.
 
         :param statement: subclass of str which also contains the parsed input
         :return: a potentially modified version of the input Statement object
+
+        See :meth:`~cmd2.Cmd.register_postparsing_hook` and
```
tleonhardt: Excellent job throughout improving the formatting and cross-referencing.

cmd2/decorators.py
```diff
@@ -17,12 +31,22 @@ def cat_decorator(func):
...
+    decorated method will receive a list of arguments parsed from user input.
 
-    :param args: Single-element positional argument list containing do_* method this decorator is wrapping
-    :param preserve_quotes: if True, then argument quotes will not be stripped
+    :param args: Single-element positional argument list containing ``do_*`` method
+                 this decorator is wrapping
+    :param preserve_quotes: if ``True``, then argument quotes will not be stripped
     :return: function that gets passed a list of argument strings
+
+    :Example:
```
tleonhardt: I really like the couple examples added to the decorators.  Consider adding examples to the other two decorators.

docs/api/index.rst
```diff
@@ -1,11 +1,30 @@
...
+If a release of this library changes any of the items documented here, the
+version number will be incremented according to the `Semantic Version
+Specification <https://semver.org>`_.
+
+This documentation is for version |version| of ``cmd2``.
+
 .. toctree::
    :maxdepth: 1
 
    cmd
```
tleonhardt: Should we attempt to order this list in any way?

cmd2/cmd2.py
```diff
@@ -374,24 +387,26 @@ def __init__(self, completekey: str = 'tab', stdin=None, stdout=None, *,
...
         """
         try:
             del self.settables[name]
         except KeyError:
             raise KeyError(name + " is not a settable parameter")
 
-    def build_settables(self):
-        """Populates self.add_settable with parameters that can be edited via the set command"""
+    def _build_settables(self):
+        """Construct the default settings"""
```
kmvanbrunt: Consider: Construct the default settable list

tleonhardt:
I'm going to go ahead and merge this large PR in.  If anyone wants to address comments, that can be done in another PR.

PR #900, opened 2020-02-24
Updated python scripting exmaple to illustrate how

### LINKED ISSUE #898: Python scripting example is functional but not correct
Our python scripting example is functional but misses the power of the python bridge generating the CommandResult for you. Need to update the `python_scripting.py` and `conditional.py` examples.

anselor:
stdout/stderr is automatically captured in CommandResult during python scripting.
Fixes #898


examples/python_scripting.py
```diff
@@ -77,13 +75,12 @@ def do_cd(self, arglist):
             except Exception as ex:
                 err = '{}'.format(ex)
             else:
-                out = 'Successfully changed directory to {!r}\n'.format(path)
-                self.stdout.write(out)
+                self.poutput('Successfully changed directory to {!r}\n'.format(path))
```
kmvanbrunt: You can remove `\n` since poutput() already adds one.

anselor: Removed that. Also changed everything to in the file to use f-strings.

kmvanbrunt: We have to support Python 3.5. Our examples can't have f-strings, as nice as they are.

tleonhardt: As long as it is only in example code, I wouldn't strongly object to begin using f-strings.  But if we do that, I would want at nice warning in the docstring at the top of the file saying something like **WARNING: This example requires Python 3.6+***

examples/python_scripting.py
```diff
@@ -29,12 +29,11 @@ def __init__(self):
         super().__init__(use_ipython=True)
         self._set_prompt()
         self.intro = 'Happy 𝛑 Day.  Note the full Unicode support:  😇 💩'
-        self.self_in_py = True
 
     def _set_prompt(self):
         """Set prompt so it displays the current working directory."""
         self.cwd = os.getcwd()
-        self.prompt = ansi.style('{!r} $ '.format(self.cwd), fg='cyan')
+        self.prompt = ansi.style(f'{self.cwd} $ ', fg='cyan')
```
tleonhardt: We should remove the f-strings so that the examples work with all currently supported versions of Python (3.5+).  

Other than that, this looks good.

tleonhardt:
@kmvanbrunt Said he is OK leaving the f-strings in examples since we will be ending support for Python 3.5 this fall once Python 3.5 hits EOL.

So I will go ahead and merge this in.

PR #906, opened 2020-03-05
Parsing exception

### LINKED ISSUE #905: Postcommand Hooks and argparse
Hi

I am using argparse for autocompletion and with sub-commands. 
I always annotate my command methods with `@with_argparser(parser)` which i think is the right way to use the argparser with cmd2. 
I did experiment with postcommand hooks using the `register_postcmd_hook` method. 
The documentation states, that the registered postcommand is only executed when the command method was called (in case of a raised exception the postcommand it is not executed). 

Now, if argparse can't parse the submitted args (for example when there are too few), the command method is not executed, but the postcommand hooks will be executed anyway. 

I understand, that in this context, the `@with_argparser(parser)` belongs to the method and thus one could argument, the method was called and didn't raise an exception, but on the other side, in most cases (i think), it makes not a lot of sense to execute postcommands if argparse can't parse the arguments of the command method. 
Is this behaviour intended? And if yes, is there a possibility to change it? I know i could ditch the annotations, use the argparser inside the command method and raise an exception if argparse fails, but this seems to be a lot of overhead (and i would lose the autocomplete stuff).

Could i possibly "monkey patch" the `with_argparser` method so a exception will be raised? (Also not a very elegant solution i think)

Thanks for any help.

kmvanbrunt:
Fixes #905 

Before we merge this, I need to update `test_debug_not_settable()` since missing a quote no longer calls `pexcept()`.

We also need a unit test to make sure postcmd hooks aren't called if `argparse` fails.

I have 2 questions:
1. Is `CmdLineError` the right name for this exception? My docstring says it's an error that occurs during parsing. But that's slightly misleading since `StatementParser` already parsed the command line. `CmdLineErrors` are raised by our `argparse` decorators which actually parse after postparsing hooks.

2. In `onecmd_plus_hooks`, should I remove "Invalid Syntax: " before the `shlex` error? I'm assuming that `ValueError` will always be a syntax error, but that may not be correct.

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -50,7 +50,7 @@ from . import utils
 from .argparse_custom import CompletionItem, DEFAULT_ARGUMENT_PARSER
 from .clipboard import can_clip, get_paste_buffer, write_to_paste_buffer
 from .decorators import with_argparser
-from .exceptions import Cmd2ArgparseException, EmbeddedConsoleExit, EmptyStatement
+from .exceptions import Cmd2ArgparseError, Cmd2ShlexError, EmbeddedConsoleExit, EmptyStatement
 from .history import History, HistoryItem
 from .parsing import StatementParser, Statement, Macro, MacroArg, shlex_split
 from .rl_utils import rl_type, RlType, rl_get_point, rl_set_prompt, vt100_support, rl_make_safe_prompt, rl_warning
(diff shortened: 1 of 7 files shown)


cmd2/cmd2.py
```diff
@@ -50,7 +50,7 @@
 from .argparse_custom import CompletionItem, DEFAULT_ARGUMENT_PARSER
 from .clipboard import can_clip, get_paste_buffer, write_to_paste_buffer
 from .decorators import with_argparser
-from .exceptions import EmbeddedConsoleExit, EmptyStatement
+from .exceptions import CmdLineError, EmbeddedConsoleExit, EmptyStatement
```
tleonhardt: Maybe we should call this something like `Cmd2ArgparseException`?

cmd2/cmd2.py
```diff
@@ -1599,12 +1599,11 @@ def onecmd_plus_hooks(self, line: str, *, add_to_history: bool = True, py_bridge
         stop = False
         try:
             statement = self._input_line_to_statement(line)
-        except EmptyStatement:
+        except (EmptyStatement, ValueError) as ex:
+            if isinstance(ex, ValueError):
```
tleonhardt: Is there a specific place in the code we expect shlex.split to potentially throw a ValueError exception?  If there are one or two such places, I would recommend catching the exception there and re-raising it as a custom exception called something like **Cmd2ShlexException**.

PR #915, opened 2020-04-09
Added documentation about decorator order and updated an example

### LINKED ISSUE #913: argparse decorator command documentation needs warning about order
A basic fact of life is that the order of Python decorators matters.  However, decorators are a somewhat advanced feature of Python that not all developers grok, especially in the scenario of multiple decorators applied to the same function.

To help make things more clear, we should add a warning to the documentation for `@cmd2.with_argparser()` to explain that if developers are using it in conjunction with a custom decorator, say `@my_decorator`, then the order of that decorator relative to `@cmd2.with_argparser` matters when it comes to runtime behavior.

If the developer does not want the custom decorator runtime behavior to occur in the case of an `argparse` error, then that decorator needs to go **after** the `argparse` one, e.g.:
```Python
@cmd2.with_argparser(foo_parser)
@my_decorator
def do_foo(self, args: argparse.Namespace) -> None:
    """foo docs"""
    pass
```

However, if the developer does wnat the customer decorator runtime behavior to occur even in the case of an `argparse` error, then that decorator needs to go **before** the `arpgarse` one, e.g.:
```Python
@my_decorator
@cmd2.with_argparser(bar_parser)
def do_bar(self, args: argparse.Namespace) -> None:
    """bar docs"""
    pass
```

### LINKED ISSUE #914: Fix Sphinx 3.0.0 autodoc warnings
Sphinx 3.0.0 was released on April 5, 2020 and it has some significant improvements such as automatically adding links to classes specific in type hints in autodoc API documentation .

However, at least at this time this ends up generating a ton of warnings because it can't find documentation for classes, especially ones from the `typing` module.

One way around this is to add ` `nitpick_ignore` to the **docs/config.py** file like so:
```Python
# Ignore nitpicky warnings from autodoc which are occurring for very new versions of Sphinx and autodoc
# They seem to be happening because autodoc is now trying to add hyperlinks to docs for typehint classes
nitpick_ignore = [
    ('py:class', 'Callable[[None], None]'),
    ('py:class', 'cmd2.cmd2.Cmd'),
    ('py:class', 'cmd2.parsing.Statement'),
    ('py:class', 'IO'),
    ('py:class', 'None'),
    ('py:class', 'Union[None, Iterable, Callable]'),
]
```

But this would actually need to be significantly longer than shown above.

I'm not sure if there is a better approach.  Temporarily, we could limit to Sphinx <= 2.4.4 if we just need to stop the bleeding of failed doc builds and such.

tleonhardt:
Added documentation about decorator order and updated an example

Also:
- Deal with warnings when building docs with Sphinx 3.0.0

Closes #913 
Closes #914

### DIFF
--- a/docs/features/argument_processing.rst
+++ b/docs/features/argument_processing.rst
@@ -347,3 +347,38 @@ class which inherits from ``argparse.ArgumentParser`` and improves error and
 help output.
 
 
+Decorator Order
+---------------
+
+If you are using custom decorators in combination with either
+``@cmd2.with_argparser`` or ``@cmd2.with_argparser_and_unknown_args``, then the
+order of your custom decorator(s) relative to the ``cmd2`` decorator matters
+when it comes to runtime behavior and ``argparse`` errors.  There is nothing
+``cmd2``-specific here, this is just a side-effect of how decorators work in
+Python.  To learn more about how decorators work, see decorator_primer_.
+
+If you want your custom decorator's  runtime behavior to occur in the case of
+an ``argparse`` error, then that decorator needs to go **after** the
+``argparse`` one, e.g.::
+
+    @cmd2.with_argparser(foo_parser)
+    @my_decorator
+    def do_foo(self, args: argparse.Namespace) -> None:
+        """foo docs"""
+        pass
+
+However, if you do NOT want the customer decorator runtime behavior to occur
+even in the case of an `argparse` error, then that decorator needs to go
+**before** the ``arpgarse`` one, e.g.::
+
+    @my_decorator
+    @cmd2.with_argparser(bar_parser)
+    def do_bar(self, args: argparse.Namespace) -> None:
+        """bar docs"""
+        pass
+
+The help_categories_ example demonstrates both above cases in a concrete
+fashion.
+
+.. _decorator_primer: https://realpython.com/primer-on-python-decorators
+.. _help_categories: https://github.com/python-cmd2/cmd2/blob/master/examples/help_categories.py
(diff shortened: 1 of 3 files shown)


docs/features/argument_processing.rst
```diff
@@ -347,3 +347,38 @@ class which inherits from ``argparse.ArgumentParser`` and improves error and
...
+---------------
+
+If you are using custom decorators in combination with either
+``@cmd2.with_argparser`` or ``@cmd2.with_argparser_and_unknown_args``, then the
+order of your custom decorator(s) relative to the ``cmd2`` decorator matters
+when it comes to runtime behavior and ``argparse`` errors.  There is nothing
+``cmd2``-specific here, this is just a side-effect of how decorators work in
+Python.  To learn more about how decorators work, see decorator_primer_.
+
+If you want your custom decorator's  runtime behavior to occur in the case of
```
kmvanbrunt, line 360: Looks like you have an extra space before runtime

tleonhardt: Good catch.  I pushed a fix.

docs/features/argument_processing.rst
```diff
@@ -347,3 +347,38 @@ class which inherits from ``argparse.ArgumentParser`` and improves error and
...
+an ``argparse`` error, then that decorator needs to go **after** the
+``argparse`` one, e.g.::
+
+    @cmd2.with_argparser(foo_parser)
+    @my_decorator
+    def do_foo(self, args: argparse.Namespace) -> None:
+        """foo docs"""
+        pass
+
+However, if you do NOT want the customer decorator runtime behavior to occur
```
kotfu, line 370: Typo on this line: should be "custom" not "customer"

tleonhardt: Doh!  Good catch.  I'll probably push trivial typo fix directly to master.

PR #916, opened 2020-04-09
Initial commit of table creation API

kmvanbrunt:
Adding table creation API

It offers the following advantages over other table creation libraries
1. Already styled text is preserved, even when wrapping over multiple lines in a cell
2. Tables can be built 1 row at a time. This helps when you have large data sets that you don't want to hold in memory or when you receive portions of the data set incrementally.
3. Text wrapping handles ANSI style sequences and characters with display widths greater than 1

The main class is `TableCreator`. It has 1 public method: `generate_row()`

This function and the `Column` class provide all features needed to build tables with headers, borders, colors, horizontal and vertical alignment, and wrapped text. However, it's generally easier to inherit from this class and implement a more granular API rather than use `TableCreator` directly. There are ready-to-use examples of provided: `SimpleTable`, `BorderedTable`, `AlternatingTable`.

It doesn't have certain advance features provided by the Python library [TableFormatter](https://github.com/python-tableformatter/tableformatter) like:
1. Assigning a data formatter function to a column
2. Handling of Panda and NumPy structures

This is intentional to keep the `TableCreator` class generic. However a child class or wrapper could be written to support these functions.


cmd2/table_creator.py
```diff
@@ -0,0 +1,748 @@
+# coding=utf-8
+"""
+cmd2 table creation API
+This API is built upon two core classes: Column and TableCreator
+The general use case is to inherit from TableCreator to create a table class with custom formatting options.
+There are already implemented and ready-to-use examples of this below TableCreator's code.
+"""
```
tleonhardt, line 7: The code, example, and tests all look good and the test coverage is excellent.  This looks like a nice new feature.  

What is missing is documentation and a changelog note.  I also since this feature is brand  new and significant, I think we should mark this feature as being in an "alpha" or "beta" state and subject to change without it counting as a backwards-incompatible change which would trigger a major version update.  Perhaps we need to agree on good way to do this for features.  Previously we discussed ways to make a feature as deprecated but we haven't discussed ways of marking them as in-development.  @kotfu Do you have any suggestions in this regard?

I think the following should be added:
1. CHANGELOG entry for an enhancement describing this new feature which also flags it as in-development, beta, or alpha or something like that
2. Sphinx doc entry of some kind, probably feature and API - the feature one can be brief and link to the example
3. We should probably convert the existing table_display.py example to use this new built-in capability

cmd2/table_creator.py
```diff
@@ -0,0 +1,748 @@
...
+            table_buf.write(header)
+        else:
+            top_border = self.generate_table_top_border()
+            table_buf.write(top_border)
+
+        for index, row_data in enumerate(table_data):
+            # Apply appropriate background color, but don't change the original
+            to_display = list()
+            for col_index, col in enumerate(row_data):
+                to_display.append(self._apply_bg_color(col))
```
kotfu, line 742: `index` and `col_index` are unused, and could be replaced with `_` to eliminate pylon errors.

examples/table_creation.py
```diff
@@ -0,0 +1,77 @@
...
+        return "${:,.2f}".format(self.val)
+
+
+# Text styles used in the data
+bold_yellow = functools.partial(ansi.style, fg=ansi.fg.bright_yellow, bold=True)
+blue = functools.partial(ansi.style, fg=ansi.fg.bright_blue)
+green = functools.partial(ansi.style, fg=ansi.fg.green)
+
+# Table Columns
+columns: List[Column] = list()
```
kotfu, line 28: This type syntax doesn't work with python 3.5. Since we officially support that version, recommend changing so the example works in all supported versions.

cmd2/table_creator.py
kotfu: Agree with @tleonhardt, @kmvanbrunt has made a nice new feature here. I made a few minor comments, but this is a really good PR. Agree with the checklist above to complete before merging.

Re alpha and beta state: the only way I know to do that is to make a feature branch that we tag with releases like "cmd2-1.1b1". We keep master as is so we could release "cmd2-1.0.x" versions before "cmd2-1.1" is completed.

cmd2/table_creator.py
kmvanbrunt: This code has been updated

examples/table_creation.py
```diff
@@ -0,0 +1,77 @@
...
+columns.append(Column("Name", width=20))
+columns.append(Column("Address", width=38))
+columns.append(Column("Income", width=14,
```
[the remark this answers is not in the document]
kmvanbrunt: The chosen widths actually had no significance. But it's a good idea to add a comment explaining width only applies to the data and does not account for borders or padding. My latest commit has documented this.

examples/table_creation.py
kmvanbrunt: I actually argued this same point in PR #900. However, it was decided, and I agreed, that supporting Python 3.5 in examples is no longer required.

cmd2/table_creator.py
kmvanbrunt: I just removed the `tableformatter` example instead of converting it. The `TableCreator` example shows off our features pretty well.

PR #917, opened 2020-04-11
app() echo updates

kmvanbrunt:
- Made some optional arguments with defaults keyword-only.
- Added unit test for echo argument to pyscript app() command.

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -213,7 +213,7 @@ class Cmd(cmd.Cmd):
         self.self_in_py = False
 
         # Commands to exclude from the help menu and tab completion
-        self.hidden_commands = ['eof', '_relative_load', '_relative_run_script']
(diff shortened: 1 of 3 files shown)


cmd2/cmd2.py
```diff
@@ -213,7 +213,7 @@ def __init__(self, completekey: str = 'tab', stdin=None, stdout=None, *,
         self.self_in_py = False
 
         # Commands to exclude from the help menu and tab completion
-        self.hidden_commands = ['eof', '_relative_load', '_relative_run_script']
+        self.hidden_commands = ['eof', '_relative_run_script']
```
tleonhardt, line 216: Good catch.  We were hiding a command which no longer exists ;-)

kmvanbrunt:
Messed up some commits because I had stuff stashed. I'm going to redo this.

PR #931, opened 2020-04-25
Added info on SemVer and branching strategy to CONTRIBUTING.md

### LINKED ISSUE #713: Determine future branching strategy
When we get to 1.0.0 and are ready to really do semantic versioning, our current branch/merge approach will not be sufficient. After 1.0.0 is tagged, we could have potentially three active development branches:

- 1.0.1
- 1.1.0
- 2.0.0

We will need to rethink how we use **master**. Maybe we need to only merge to master when we cut a release, so **master** HEAD contains the most "recent" (according to semver) release. We probably will also need to start using a **develop** branch as a place to integrated all the hotfix, minor (i.e. 1.x.0) and major (i.e. x.0.0) branches.

We are not the first people to have to solve this problem, after a little digging, I found
[Hendrick Prinsloo's description of his approach](https://medium.com/@HendrikPrinsZA/gitflow-with-semver-dde625429aca). I think this is a reasonably sound method, and propose that we adopt it.

Opinions and debate welcome.

tleonhardt:
Added info on semantic versioning and branching strategy to CONTRIBUTING.md

Also:
- Added `isort` to Pipenv dev
- Added setup.cfg to make it easy to run `flake8`, `doc8`, and `isort` directly from the command line without using `invoke`
- Ran `isort` to sort includes

The majority of the changes in this PR were automatic changes made by [isort](https://github.com/timothycrosley/isort) to sort imports to match idiomatic Python standards.

However, the main purpose was to add info on SerVer and branching strategy to the contributor's guide.  I just got carried away with a couple other minor things I've been intending to do for a long while and hadn't found the time.

Closes #713

### DIFF
--- a/cmd2/ansi.py
+++ b/cmd2/ansi.py
@@ -6,10 +6,10 @@ setting the window title, and asynchronous alerts.
 import functools
 import re
 from enum import Enum
-from typing import Any, IO, List, Union
+from typing import IO, Any, List, Union
 
 import colorama
-from colorama import Fore, Back, Style
+from colorama import Back, Fore, Style
 from wcwidth import wcswidth
 
 # On Windows, filter ANSI escape codes out of text sent to stdout/stderr, and replace them with equivalent Win32 calls
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -65,7 +65,7 @@ The tables below list all prerequisites along with the minimum required version 
 #### Additional prerequisites to build cmd2 documentation
 | Prerequisite                                | Minimum Version |
 | ------------------------------------------- | --------------- |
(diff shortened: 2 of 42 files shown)


cmd2/ansi.py
```diff
@@ -6,10 +6,10 @@
 import functools
 import re
 from enum import Enum
-from typing import Any, IO, List, Union
+from typing import IO, Any, List, Union
```
kmvanbrunt, line 9: This is the only import I see where it wasn't made alphabetical. Is this the isort version? Does it place all uppercase imports first?

tleonhardt: Yeah this is how isort does it.  It assumes that anything in all-caps is a constant per Pep8 naming conventions and it puts these first before non-constants.

PR #933, opened 2020-04-27
Fix #932

### LINKED ISSUE #932: Version 1.0.2 changed the way SystemExit exceptions raised from do_command methods are handled
I have a cmd2 application which does it's own argparsing inside of the `do_command()` methods, i.e. it does not use any of the cmd2 built-in arg parsing functionality. This is done by calling `argparse.parse_args()` as one of the first lines of code inside the `do_command()` method.

In 1.0.1 if a `do_command()` method raises a `SystemExit` exception, as `parse_args()` does if given the '-h' option or if a syntax error occurs, that exception would be swallowed by `cmd2` and the command prompt would be printed again.

In 1.0.2, that exception is not swallowed and the `cmd2` app terminates.

You can replicate this issue by adding a `raise SystemExit` to the `do_speak()` method in `examples/first_app.py`, and running it with the two versions of `cmd2`.

The behavior of other raised exceptions does not appear to have changed between 1.0.1 and 1.0.2. For example if you raise a ValueError, you get:
```
EXCEPTION of type 'ValueError' occurred with message: ''
To enable full traceback, run the following command: 'set debug true'
```
in both versions of `cmd2`.

This change in behavior is caused by the change in the logic on the last two lines of the `onecmd_plus_hooks()` method made between 1.0.1 and 1.0.2.

In the short term, I request that we quickly revert this change in behavior. After that, I suggest we discuss the desired behavior is for `cmd2` when a `SystemExit` exception is raised, and then intentionally implement and document that behavior.

kotfu:


cmd2/cmd2.py
```diff
@@ -1680,9 +1680,7 @@ def onecmd_plus_hooks(self, line: str, *, add_to_history: bool = True,
         except Exception as ex:
             self.pexcept(ex)
         finally:
-            stop = self._run_cmdfinalization_hooks(stop, statement)
-
-        return stop
+            return self._run_cmdfinalization_hooks(stop, statement)
```
tleonhardt, line 1683: It looks like this change is causing two unit tests to fail:
1. `test_runcmds_plus_hooks_ctrl_c`
2. `test_generate_transcript_stop`

tleonhardt:
@kotfu @kmvanbrunt 
Should we close this PR without merging because it is now OBE?

tleonhardt:
This appears to be OBE due to #937 

So I will close.  If I am incorrect, please feel free to re-open

PR #937, opened 2020-05-07
Exception handling

### LINKED ISSUE #932: Version 1.0.2 changed the way SystemExit exceptions raised from do_command methods are handled

kmvanbrunt:
This is my attempt to address #932 
If everyone agrees, then I'll finish writing unit tests and update the change log.

The question remains about what we do with a `SystemExit`. This PR prints to the user and sets `stop` to True. I'm also OK if it behaves like all other exceptions. In that case, we would change our except statement from:

`except Exception` to `except BaseException`

That would be consistent with the documentation @kotfu pointed to:

> If the command method raises an exception, cmd2 will catch it and display it for you

It also preserves exactly 1 method to exit the command loop (i.e. return True from a command)

Closes #932

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 1.0.3 (TBD, 2020)
+## 1.1.0 (TBD, 2020)
 * Bug Fixes
     * Fixed issue where subcommand usage text could contain a subcommand alias instead of the actual name
 * Enhancements
@@ -14,6 +14,15 @@
           documentation for an overview.
         * See [table_creation.py](https://github.com/python-cmd2/cmd2/blob/master/examples/table_creation.py)
           for an example.
+    * Added the following exceptions to the public API
+        * `SkipPostcommandHooks` - Custom exception class for when a command has a failure bad enough to skip
+          post command hooks, but not bad enough to print the exception to the user.
+        * `Cmd2ArgparseError` - A `SkipPostcommandHooks` exception for when a command fails to parse its arguments.
+          Normally argparse raises a `SystemExit` exception in these cases. To avoid stopping the command
+          loop, catch the `SystemExit` and raise this instead. If you still need to run post command hooks
+          after parsing fails, just return instead of raising an exception.
+    * Added explicit handling of `SystemExit`. If a command raises this exception, the command loop will be
+      gracefully stopped.
     
 ## 1.0.2 (April 06, 2020)
 * Bug Fixes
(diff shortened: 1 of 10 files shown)


cmd2/__init__.py
```diff
@@ -25,6 +25,7 @@
 from .cmd2 import Cmd
 from .constants import COMMAND_NAME, DEFAULT_SHORTCUTS
 from .decorators import with_argument_list, with_argparser, with_argparser_and_unknown_args, with_category
+from .exceptions import Cmd2ArgparseError, SkipPostcommandHooks
```
tleonhardt, line 28: These changes definitely need some sort of CHANGELOG entry

kotfu: - Suggest updating https://cmd2.readthedocs.io/en/latest/features/commands.html#exception-handling to describe the `BaseException` exclusion for error handling, and the behavior when raising `Cmd2ArgparseError` or `SkipPostcommandHooks` exceptions.
- Also suggest updating https://cmd2.readthedocs.io/en/latest/features/hooks.html to describe this new exception handling behavior
- At a minimum, this change should bump our version to 1.1, since we have added new exceptions to the public API. However, I could make a strong case that change to the `do_command` API (changed exception handling behavior) merits a bump to 2.0. We should discuss.

tleonhardt: Updating those docs and bumping to version to 1.1 sounds like a good idea

CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-## 1.0.3 (TBD, 2020)
+## 1.1.0 (TBD, 2020)
```
tleonhardt, line 1: Thanks for updating the CHANGELOG appropriately

Issue #943, opened 2020-06-09
Allow modular command definition


anselor:
Allow commands to be split into multiple classes that are registered with cmd2 instead of requiring all commands to be defined in one giant class.

Breakdown:
- Define a base class for these satellite command definitions.
- Everywhere in cmd2 where it introspects for `do_`, `complete_`, or `help_` also search in the satellite command objects.
- Stretch - auto-detect all satellite command definitions (by subclass) and load them automatically on init of cmd2.Cmd.
- Example of loading a satellite command object

muhrin:
Thanks a lot for introducing this feature.  I, for one, would find this to be very useful.

In general, I think it's also a good thing for `cmd2` to move away (or at minimum provide alternatives to) to the monolithic model of putting everything in one class as this severely restricts modularity and leads to a very complex state.

If I can be of any help with testing or development, do let me know, I'd be happy to pitch in some of my time.

muhrin:
I've been playing around with this new API and I think it's very nice.  It brings some very helpful modularity to `Cmd2`.

There's one thing that I find a little awkward though, which is the interface for the `do/help/complete` functions in `CommandSet`.  Placing the `CommandSet` as the second positional argument removes some of the flexibility of being able to move a command function (without modification) from being a free function, `Cmd` member function or `CommandSet` member function.  If, however, the `CommandSet` was supplied as a keyword argument it would be possible to use the same function interchangably, e.g.:

```
def do_echo(app: Cmd, statement: Statement, *, command_set: CommandSet=None):
    ....
```
This can now be used in the following flexible ways:
```
# 1
class App(Cmd):
    <Copy/paste do_echo here>
# 2
setattr(cmd, 'do_echo', do_echo)
# 3
class Commands(CommandSet):
    <Copy/paste do_echo here>
# 4
setattr(command_set, 'do_echo', do_echo)
```
And there are no doubt others.  Point being, that there would be one unique signature for each `do/help/complete` (of course a user implementing a `Cmd` member could choose to omit the `command_set` completely knowing it will never be passed).

[...]

PR #945, opened 2020-06-15
Load/Unload commands at runtime - Implements #943

anselor:
- Group commands in CommandSets to be installed/uninstalled as a batch. Commands within a CommandSet can be optionally tagged with at default category. Individual commands may override the default category.
- Install/Uninstall individual command functions at runtime
- Tag command functions for automatic discovery and load
- Automatic discovery of complete_ and help_ functions that are contained in the same module as command functions.
- Automatic discovery and load of CommandSets and registered command functions that can be disabled in the constructor

### DIFF
--- a/cmd2/__init__.py
+++ b/cmd2/__init__.py
@@ -28,6 +28,7 @@ if cmd2_parser_module is not None:
 # Get the current value for argparse_custom.DEFAULT_ARGUMENT_PARSER
 from .argparse_custom import DEFAULT_ARGUMENT_PARSER
 from .cmd2 import Cmd
+from .command_definition import CommandSet, with_default_category, register_command
 from .constants import COMMAND_NAME, DEFAULT_SHORTCUTS
 from .decorators import with_argument_list, with_argparser, with_argparser_and_unknown_args, with_category
 from .exceptions import Cmd2ArgparseError, SkipPostcommandHooks
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -471,7 +471,7 @@ class Cmd(cmd.Cmd):
 
     def uninstall_command_set(self, cmdset: CommandSet):
         """
-        Uninstalls an CommandSet and unloads all associated commands
+        Uninstalls a CommandSet and unloads all associated commands
         :param cmdset: CommandSet to uninstall
         """
         if cmdset in self._installed_command_sets:
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -264,14 +264,14 @@ class Cmd(cmd.Cmd):
         self.last_result = None
 
         # Used by run_script command to store current script dir as a LIFO queue to support _relative_run_script command
-        self._script_dir = []  # type: List[AnyStr]
+        self._script_dir = []  # type: List[str]
 
         # Context manager used to protect critical sections in the main thread from stopping due to a KeyboardInterrupt
         self.sigint_protection = utils.ContextFlag()
 
         # If the current command created a process to pipe to, then this will be a ProcReader object.
         # Otherwise it will be None. It's used to know when a pipe process can be killed and/or waited upon.
-        self._cur_pipe_proc_reader = None
+        self._cur_pipe_proc_reader = None  # type: Optional[utils.ProcReader]
 
         # Used to keep track of whether we are redirecting or piping output
         self._redirecting = False
@@ -295,7 +295,7 @@ class Cmd(cmd.Cmd):
         self.broken_pipe_warning = ''
 
         # Commands that will run at the beginning of the command loop
-        self._startup_commands = []
+        self._startup_commands = []  # type: List[str]
 
         # If a startup script is provided and exists, then execute it in the startup commands
         if startup_script:
@@ -304,7 +304,7 @@ class Cmd(cmd.Cmd):
                 self._startup_commands.append("run_script {}".format(utils.quote_string(startup_script)))
 
         # Transcript files to run instead of interactive command loop
-        self._transcript_files = None
+        self._transcript_files = None  # type: Optional[List[str]]
 
         # Check for command line args
         if allow_cli_args:
@@ -2088,7 +2088,7 @@ class Cmd(cmd.Cmd):
                                                         self._cur_pipe_proc_reader, self._redirecting)
 
         # The ProcReader for this command
-        cmd_pipe_proc_reader = None
+        cmd_pipe_proc_reader = None  # type: Optional[utils.ProcReader]
 
         if not self.allow_redirection:
             # Don't return since we set some state variables at the end of the function
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -46,7 +46,7 @@ from typing import Any, AnyStr, Callable, Dict, Iterable, List, Mapping, Optiona
 from . import ansi, constants, plugin, utils
 from .argparse_custom import DEFAULT_ARGUMENT_PARSER, CompletionItem
 from .clipboard import can_clip, get_paste_buffer, write_to_paste_buffer
-from .command_definition import _UNBOUND_COMMANDS, CommandSet, _partial_passthru
+from .command_definition import _REGISTERED_COMMANDS, CommandSet, _partial_passthru
 from .constants import COMMAND_FUNC_PREFIX, COMPLETER_FUNC_PREFIX, HELP_FUNC_PREFIX
 from .decorators import with_argparser
 from .exceptions import Cmd2ShlexError, EmbeddedConsoleExit, EmptyStatement, RedirectionError, SkipPostcommandHooks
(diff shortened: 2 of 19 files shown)


tleonhardt:
On my laptop I'm getting the following error trying to build the Sphinx docs using Python 3.8.2 and Sphinx 3.1.1:
```sh
  File "/Users/toddleonhardt/src/cmd2/.venv/lib/python3.8/site-packages/sphinx/util/logging.py", line 421, in filter
    raise exc
sphinx.errors.SphinxWarning: /Users/toddleonhardt/src/cmd2/cmd2/cmd2.py:docstring of cmd2.Cmd::py:class reference target not found: cmd2.command_definition.CommandSet
```

cmd2/cmd2.py
```diff
@@ -37,14 +37,17 @@
...
 from code import InteractiveConsole
 from collections import namedtuple
 from contextlib import redirect_stdout
-from typing import Any, Callable, Dict, Iterable, List, Mapping, Optional, Tuple, Type, Union
+from typing import Any, AnyStr, Callable, Dict, Iterable, List, Mapping, Optional, Tuple, Type, Union
 
 from . import ansi, constants, plugin, utils
 from .argparse_custom import DEFAULT_ARGUMENT_PARSER, CompletionItem
 from .clipboard import can_clip, get_paste_buffer, write_to_paste_buffer
+from .command_definition import _REGISTERED_COMMANDS, CommandSet, _partial_passthru
```
tleonhardt, line 49: importing a couple protected members feels off

cmd2/cmd2.py
```diff
@@ -130,7 +133,9 @@ def __init__(self, completekey: str = 'tab', stdin=None, stdout=None, *,
                  startup_script: str = '', use_ipython: bool = False,
                  allow_cli_args: bool = True, transcript_files: Optional[List[str]] = None,
                  allow_redirection: bool = True, multiline_commands: Optional[List[str]] = None,
-                 terminators: Optional[List[str]] = None, shortcuts: Optional[Dict[str, str]] = None) -> None:
+                 terminators: Optional[List[str]] = None, shortcuts: Optional[Dict[str, str]] = None,
+                 command_sets: Optional[Iterable[CommandSet]] = None,
+                 auto_load_commands: bool = True) -> None:
```
tleonhardt, line 138: Is there any risk of backwards incompatibility with the default value of `auto_load_commands` being `True`?

I'm thinking we should make the default `False` for safe backwards compatibility.

cmd2/cmd2.py
```diff
@@ -249,7 +264,7 @@ def __init__(self, completekey: str = 'tab', stdin=None, stdout=None, *,
         self.last_result = None
 
         # Used by run_script command to store current script dir as a LIFO queue to support _relative_run_script command
-        self._script_dir = []
+        self._script_dir = []  # type: List[AnyStr]
```
tleonhardt, line 267: Should it be `str` or `AnyStr`?  Would there ever be a case where we assign a `bytes`?

cmd2/command_definition.py
```diff
@@ -0,0 +1,131 @@
...
+)
+from .constants import COMMAND_FUNC_PREFIX
+
+# Allows IDEs to resolve types without impacting imports at runtime, breaking circular dependency issues
+try:  # pragma: no cover
+    from typing import TYPE_CHECKING
+    if TYPE_CHECKING:
+        from .cmd2 import Cmd, Statement
+        import argparse
+except ImportError:
```
tleonhardt, line 22: When would this `ImportError` ever occur?  `TYPE_CHECKING` was added in 3.5.2.  Theoretically we say we support Python `3.5+`.  But I'd be OK bumping that to `3.5.2+` if need be.

cmd2/command_definition.py
```diff
@@ -0,0 +1,131 @@
...
+    :param func: wrapped function
+    :param args: positional arguments
+    :param kwargs: keyword arguments
+    :return: partial function that exposes attributes of wrapped function
+    """
+    def __getattr__(self, item):
+        return getattr(self.func, item)
+
+    def __setattr__(self, key, value):
+        return setattr(self.func, key, value)
```
tleonhardt, line 45: Recommend adding unit test for this method

cmd2/command_definition.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def __dir__(self) -> Iterable[str]:
+        return dir(self.func)
```
tleonhardt, line 48: Recommend adding unit test for this method

anselor:
> On my laptop I'm getting the following error trying to build the Sphinx docs using Python 3.8.2 and Sphinx 3.1.1:
> 
> ```shell
>   File "/Users/toddleonhardt/src/cmd2/.venv/lib/python3.8/site-packages/sphinx/util/logging.py", line 421, in filter
>     raise exc
> sphinx.errors.SphinxWarning: /Users/toddleonhardt/src/cmd2/cmd2/cmd2.py:docstring of cmd2.Cmd::py:class reference target not found: cmd2.command_definition.CommandSet
> ```

That's weird. I ran sphinx on mine and didn't have any problems. I'm running Ubuntu 20.04 through WSL with the latest. I'll have to check when I get back on that machine what the versions were.

cmd2/cmd2.py
anselor: I made them protected to indicate to external users that they shouldn't directly access these. It's for internal cmd2 use only.

cmd2/cmd2.py
anselor: No risk to backwards compatibility here. The option only uses the new decorators and base classes that didn't exist before. I want the default behavior to be that things 'just work' by defining functions and classes using the decorators and base class provided. The extra work of disabling this is for people that have more specific behavior they want to control. For projects that existed prior to this they wouldn't be using the new decorator and base class so they'll be unaffected.

cmd2/cmd2.py
anselor: I don't know. Was just trying to give a type to an untyped variable. PyCharm auto-detected this field as being likely List[AnyStr] so I went with it.

cmd2/command_definition.py
anselor: That's the only reason for this check, for python earlier than 3.5.2.

cmd2/command_definition.py
anselor: I only really needed the getattr to work to pass through the docstring from the wrapped function. I added setattr for completeness but I don't really know how to test this in a sensible fashion.

cmd2/command_definition.py
anselor: Same as above.

examples/modular_commands/commandset_basic.py
```diff
@@ -0,0 +1,121 @@
+# coding=utf-8
```
tleonhardt, line 1: Creating an empty __init__.py file in this directory might be a good enough hack to get the examples to run but probably not to work nicely in an IDE

cmd2/__init__.py
```diff
@@ -23,6 +23,7 @@
 # Get the current value for argparse_custom.DEFAULT_ARGUMENT_PARSER
```
tleonhardt, line 23: Before merging in this would need a CHANGELOG entry for version 1.2.0 or whatever.

cmd2/cmd2.py
```diff
@@ -381,6 +396,192 @@ def __init__(self, completekey: str = 'tab', stdin=None, stdout=None, *,
...
+        Load modular command definitions.
+        :return: None
+        """
+
+        # start by loading registered functions as commands
+        for cmd_name in _REGISTERED_COMMANDS.keys():
+            self.install_registered_command(cmd_name)
+
+        # Search for all subclasses of CommandSet, instantiate them if they weren't provided in the constructor
+        all_commandset_defs = CommandSet.__subclasses__()
```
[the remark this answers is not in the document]
anselor: This is specifically subclasses of this class. It's not a name based match.

cmd2/cmd2.py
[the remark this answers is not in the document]
anselor: It would only affect your code if you modified it to add support for this. Legacy code would be unaffected because all of this is new and didn't exist before. The goal was to make the default case easier to use.

cmd2/cmd2.py
```diff
@@ -200,7 +205,7 @@ def __init__(self, completekey: str = 'tab', stdin=None, stdout=None, *,
         self.max_completion_items = 50
 
         # A dictionary mapping settable names to their Settable instance
-        self.settables = dict()
+        self.settables = dict()  # type: Dict[str, Settable]
```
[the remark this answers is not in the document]
anselor: Yeah, PyCharm treats these comments same as if you had declared a type hint. Mypy does as well.

cmd2/cmd2.py
```diff
@@ -381,6 +396,192 @@ def __init__(self, completekey: str = 'tab', stdin=None, stdout=None, *,
...
         self.matches_sorted = False
 
+    def _autoload_commands(self) -> None:
+        """
+        Load modular command definitions.
+        :return: None
+        """
+
+        # start by loading registered functions as commands
+        for cmd_name in _REGISTERED_COMMANDS.keys():
```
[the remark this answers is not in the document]
anselor: `REGISTERED_COMMANDS` is in a static context. It's where command functions are registered as available to load with the new decorator. `self.__installed_functions` is inside the instance and are there specific commands that are currently active.

examples/modular_commands/commandset_basic.py
anselor: Yeah, the layout of the examples doesn't really support this very well.

examples/modular_commands_main.py
```diff
@@ -0,0 +1,127 @@
+#!/usr/bin/env python
```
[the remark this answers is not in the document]
anselor: That's weird. It shows up as executable for me.

anselor:
Accepting #961 instead.

Issue #644, opened 2019-03-10
Improve pyscript documentation


kmvanbrunt:
The section in Sphinx  on `pyscript` command needs a bit of updating.
Topics to cover:

- Running commands with `app()`
- Developing a basic scripting API that only relies on stdout and stderr in `CommandResult`
- Developing an advanced scripting API that includes populating `cmd2.last_result` which becomes `CommandResult.data`
- Utilizing `StdSim.pause_storage` to avoid duplicating data already stored in `CommandResult.data`
- Processing a `CommandResult` object returned from app()
- Exposing custom data values with `self.py_locals`.

tleonhardt:
Given the power of ``pyscript``, our documentation has been sorely lacking in helping making this discoverable to end users

kotfu:
We have some basic pyscript documentation now. I propose we move this issue out of the 1.0.0 milestone so that we can release. Any objections?

tleonhardt:
@kotfu Sure moving this issue out of 1.0.0 sounds fine.  We really could benefit from some improved pyscript documentation at some point, but it can happen after the 1.0 release.

@kmvanbrunt Do you have any pyscript documentation that you have been working on?

anselor:
It's probably also worth nothing that the external test plugin simply opens up the pyscript as a cmd2 member function to make it easier to unit test cmd2 application commands.

jayrod:
I've written a fair bit of pyscript and would gladly add some documentation. Is this still a requirement?

anselor:
@jayrod Yes, if you're willing to improve this area of documentation and/or provide additional examples of how this can be used it would be appreciated!

jayrod:
I'm actually not sure what is meant by "Processing a CommandResult object returned from app()"  I presume you mean when the application returns a complex object via last_result that the script can then use for data?

jayrod:
After looking at the code and tinkering around a bit this also makes me scratch my head a bit.

Utilizing StdSim.pause_storage to avoid duplicating data already stored in CommandResult.data

i.e. what *is* a good use case for StdSim.pause_storage?

jayrod:
PR submitted. https://github.com/python-cmd2/cmd2/pull/1187

doesn't include anything about StdSim however.

kmvanbrunt:
@jayrod Thanks for taking the time to write documentation. I'm just finishing my vacation, so please allow me a few days to read over your PR.

jayrod:
[...]

PR #947, opened 2020-06-16
Further additions to pyscript documenntation for Issue #644

anselor:


docs/features/scripting.rst
```diff
@@ -94,6 +94,24 @@ where:
 * ``command`` and ``args`` are entered exactly like they would be entered by
   a user of your application.
 
+Advanced Support
+~~~~~~~~~~~~~~~~
+
+When implementing a command, setting ``self.__last_result`` allows for application-specific
```
tleonhardt, line 100: I think this is incorrect and should be `self.last_result` (applies below as well)

docs/features/scripting.rst
```diff
@@ -94,6 +94,24 @@ where:
 * ``command`` and ``args`` are entered exactly like they would be entered by
   a user of your application.
 
+Advanced Support
```
tleonhardt, line 97: This section should go at the end of the file.  As is, there is an extra line that is completely out of place.

anselor: I don't know what you mean by this. This is already the last section in this file.

tleonhardt: I pushed a small change to address what I was talking about

PR #949, opened 2020-07-07
Swap out tox for nox

anselor:
Nox is like tox but implemented in the form of a python file with the full power of python.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -1,0 +1,18 @@
+import nox
+
+
+@nox.session(python=['3.7'])
+def docs(session):
+    session.install('sphinx', 'sphinx-rtd-theme', '.')
+    session.chdir('docs')
+    tmpdir = session.create_tmp()
+
+    session.run('sphinx-build', '-a', '-W', '-T', '-b', 'html',
+                '-d', '{}/doctrees'.format(tmpdir), '.', '{}/html'.format(tmpdir))
+
+
+@nox.session(python=['3.5', '3.6', '3.7', '3.8', '3.9'])
+def tests(session):
+    session.install('invoke', './[test]')
+    session.run('invoke', 'pytest', '--junit', '--no-pty')
+    session.run('codecov')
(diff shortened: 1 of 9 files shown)


tleonhardt:
@anselor I have never heard of nox before.  Can you please elaborate on what you think the advantages are in comparison to tox?  Are you aware of any down sides?

@kotfu Do you have any experience with nox?

anselor:
It's inspired by tox but is fully python. Allows you to have control over the generation of venvs and run commands inside them. This would allow control over which specific wheels are installed in each venv and allow tests to be written to validate behavior with and without plug-ins.

tleonhardt:
Ok, that sounds pretty decent.  Are there any known disadvantages relative to tox?

tleonhardt:
Assuming we do make the switch, there are a few additional files which should be updated.  GitHub search in our repo for tox shows 9 files.  There are additional files like tasks.py, MANIFEST.in, .gitignore, setup.cfg, etc.  we should also delete the tox.ini unless we want to leave it around in-case.

anselor:
My understanding is tox has lots of magic it does that makes the simplest cases very simple but becomes really doesn't let you customize and stray from the path they've defined. 

Nox I think is a slightly higher learning curve to set up, but you really have full access to executing commands as if you're on the console in the venv.

Also, if you look at my other branch I've already bright in the external test plugin and gotten it running under nox.

tleonhardt:
I like that the nox config file is in Python - that seems like a plus.  If you could tie up the loose ends in terms of modifying the remaining files which use tox for equivalent nox usage I’m good to approve.

noxfile.py
```diff
@@ -0,0 +1,18 @@
+import nox
+
+
+@nox.session(python=['3.7'])
```
tleonhardt, line 4: Nice work!  Do like the format of this file better.  I find it more readable and less magical for general Python developers who aren't experts in `tox`.

PR #951, opened 2020-07-08
Moved external test plugin in as a first-class plugin

anselor:
The PR moves the [cmd2-ext-test](https://github.com/python-cmd2/cmd2-ext-test) plugin within the main `cmd2` repo to facilitate a combined CI process to help ensure that changes in `cmd2` do not break this plugin.

This is a bold new experiment at attempting to support a very limited number of "maintained" plugins and tweak our CI process to support this.

Immediately after this merges in we should update the README for the old repo to state that the plugin code is now being maintained in the main repo and add a link to the plugin directory.  Then some period of time after this merges in, if/when we have been comfortable with this change, we should delete the old repository.


plugins/ext_test/CHANGELOG.md
```diff
@@ -0,0 +1,12 @@
+# Changelog
```
tleonhardt, line 1: I'd recommend adding a README.md in the **plugins** directory to just give some basic information on:
- What a `cmd2` plugin is
- How to use a `cmd2` plugin
- How to create your own `cmd2` plugin

It can be brief as long as it provides links to where to get more info.

noxfile.py
```diff
@@ -15,4 +15,8 @@ def docs(session):
 def tests(session):
     session.install('invoke', './[test]')
     session.run('invoke', 'pytest', '--junit', '--no-pty')
+
+    session.chdir('./plugins/ext_test')
+    session.install('.[test]')
+    session.run('invoke', 'pytest', '--junit', '--no-pty', '--append-cov')
     session.run('codecov')
```
tleonhardt, line 22: Something is wrong with `codecov` in this PR - it isn't reporting data.  I think you need to move `session.run('codecov')` back above the `chdir` command.

plugins/ext_test/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
tleonhardt, line 1: Having a 2nd license might prove problematic within the same Git repo might prove problematic for some large enterprise organizations.  However, since they are both MIT it might be OK.

plugins/tasks.py
```diff
@@ -0,0 +1,138 @@
...
+
+#####
+#
+# pytest, tox, pylint, and codecov
+#
+#####
+
+
+@invoke.task()
+def pytest(_, junit=False, pty=True):
```
tleonhardt, line 32: This task is broken for me, I get a traceback like the following one:
```sh
inv pytest 
Traceback (most recent call last):
  File "/Users/toddleonhardt/src/cmd2/.venv/bin/inv", line 8, in <module>
    sys.exit(program.run())
  File "/Users/toddleonhardt/src/cmd2/.venv/lib/python3.8/site-packages/invoke/program.py", line 384, in run
    self.execute()
  File "/Users/toddleonhardt/src/cmd2/.venv/lib/python3.8/site-packages/invoke/program.py", line 566, in execute
    executor.execute(*self.tasks)
  File "/Users/toddleonhardt/src/cmd2/.venv/lib/python3.8/site-packages/invoke/executor.py", line 129, in execute
    result = call.task(*args, **call.kwargs)
  File "/Users/toddleonhardt/src/cmd2/.venv/lib/python3.8/site-packages/invoke/tasks.py", line 127, in __call__
    result = self.body(*args, **kwargs)
  File "/Users/toddleonhardt/src/cmd2/plugins/tasks.py", line 36, in pytest
    iexec.execute([
  File "/Users/toddleonhardt/src/cmd2/.venv/lib/python3.8/site-packages/invoke/executor.py", line 88, in execute
    calls = self.normalize(tasks)
  File "/Users/toddleonhardt/src/cmd2/.venv/lib/python3.8/site-packages/invoke/executor.py", line 154, in normalize
    name, kwargs = task
ValueError: not enough values to unpack (expected 2, got 1)
```

plugins/ext_test/CHANGELOG.md
```diff
@@ -0,0 +1,12 @@
...
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
+and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
+
+## 1.0.0 (2020-03-09)
```
tleonhardt, line 7: You might want to bump the version of this plugin since it looks like you are changing to versioning based on setuptools_scm

noxfile.py
anselor: Oh, I think I need to chdir back to the root.

tleonhardt: I don’t think we want to measure code coverage of plugins or at least we don’t want to report that as part of the cmd2 coverage.

plugins/ext_test/CHANGELOG.md
anselor: Something about the tag filtering falls apart on the CI servers. I don't know what the problem is. Options are either to switch to hard-coding the version or just keeping the version in lock-step with cmd2 proper.

plugins/tasks.py
anselor: I need to just delete this. I was trying to see if I could trick invoke into passing the command-line parameters down.

noxfile.py
anselor: I don't see why that's a problem. These plugins are optionally installed first-class components maintained by cmd2.

plugins/ext_test/CHANGELOG.md
anselor: @kotfu has a plugin template in a separate repo with detailed instructions. I think maybe it makes sense to go ahead and bring that in here as well.

tleonhardt: Yeah it would be reasonable to do that.

noxfile.py
tleonhardt: I guess it is fine as long as it is clear exactly what coverage is being measured

plugins/ext_test/CHANGELOG.md
tleonhardt: I feel strongly that we don't want to 100% tie the release cycle in precision lockstep, so it would be preferable to hard-code the versions of the plugins. 

You could do something like make a single plugin version in a file if you want and have the version read from there when building.

plugins/ext_test/CHANGELOG.md
anselor: Yeah, I think me plan is to leave this directory as-is for this PR. Next, bring @kotfu's plugin template in in a subsequent PR and link up the tests and coverage and add a README here.

plugins/ext_test/CHANGELOG.md
anselor: I think I'm going to hard-code the setup.py to 0.2.0 and put in an issue to figure out why the tag filtering fails on the CI system. It may be possible to switch out how it does the tag filtering depending on whether we're running it on our system vs the CI system. We only really need the tag filtering to work when we invoke the task to publish to pypi and tag it. Doesn't matter in the context of the CI reporting.

noxfile.py
tleonhardt: This seems to be fixed

PR #953, opened 2020-07-12
replace pkg_resources with importlib.metadata

dhellmann:
Importing pkg_resources has a side-effect of scanning every installed
distribution on sys.path to load the metadata, especially the entry
points defined in the packages. This can have a significant
launch-time cost for command line applications when there are a lot of
distributions to scan.

Since cmd2 is only using pkg_resources to find the version of the
installed package, pkg_resources can be replaced with
importlib.metadata. The implementation in the new library is
significantly faster because it goes immediately to the metadata file
for the requested distribution, instead of scanning all of them. There
are also no import-time side-effects.

importlib.metadata is a new standard library module starting with
python 3.8. For earlier versions, a compatible library has been
released to PyPI as 'importlib_metadata'. This change adds the new
dependency with a qualifier so that it is only applied to older
versions of python, and then updates the places that were importing
pkg_resources to look for the different versions of the new library
instead. The documentation configuration is changed to import cmd2
itself to get its version, since the package has to be installed for
the metadata to be available anyway.

### DIFF
--- a/cmd2/__init__.py
+++ b/cmd2/__init__.py
@@ -3,10 +3,15 @@
 # flake8: noqa F401
 """This simply imports certain things for backwards compatibility."""
 
-from pkg_resources import get_distribution, DistributionNotFound
 try:
-    __version__ = get_distribution(__name__).version
-except DistributionNotFound:
+    # For python 3.8 and later
+    import importlib.metadata as importlib_metadata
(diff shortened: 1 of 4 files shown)


dhellmann:
Thanks for the quick review, @tleonhardt !

PR #957, opened 2020-07-16
Initial checkin of Modular Command definitions as a plugin. #943

anselor:


anselor:
Accepting #961 instead.

PR #959, opened 2020-07-17
Fixed hook documentation

### LINKED ISSUE #958: Registering a preloop hook without return type hint throws an error
As is discussed [here](https://cmd2.readthedocs.io/en/latest/features/hooks.html?highlight=preloop#application-lifecycle-hooks), I should be able to declare and register a preloop hook as follows:

```
class App(cmd2.Cmd):
    def __init__(self, *args, *kwargs):
        super().__init__(*args, **kwargs)
        self.register_preloop_hook(self.myhookmethod)

    def myhookmethod(self):
        self.poutput("before the loop begins")
```

However, running the above will throw an error: ```TypeError: initialize_preloop must declare return a return type of 'None'```. In order to fix the error, you need to provide a return type hint:

```
class App(cmd2.Cmd):
    def __init__(self, *args, *kwargs):
        super().__init__(*args, **kwargs)
        self.register_preloop_hook(self.myhookmethod)

    def myhookmethod(self) -> None:
        self.poutput("before the loop begins")
```

Versions:
- cmd2: 1.2.0
- python: 3.7

kmvanbrunt:
Closes #958

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -4281,11 +4281,11 @@ class Cmd(cmd.Cmd):
         signature = inspect.signature(func)
         _, param = list(signature.parameters.items())[0]
         if param.annotation != plugin.CommandFinalizationData:
-            raise TypeError("{} must have one parameter declared with type "
-                            "'cmd2.plugin.CommandFinalizationData'".format(func.__name__))
+            raise TypeError("{} must have one parameter declared with type {}".format(func.__name__,
+                                                                                      plugin.CommandFinalizationData))
         if signature.return_annotation != plugin.CommandFinalizationData:
-            raise TypeError("{} must declare return a return type of "
-                            "'cmd2.plugin.CommandFinalizationData'".format(func.__name__))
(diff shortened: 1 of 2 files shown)


cmd2/cmd2.py
```diff
@@ -4281,11 +4281,11 @@ def _validate_cmdfinalization_callable(cls, func: Callable[[plugin.CommandFinali
         signature = inspect.signature(func)
         _, param = list(signature.parameters.items())[0]
         if param.annotation != plugin.CommandFinalizationData:
-            raise TypeError("{} must have one parameter declared with type "
-                            "'cmd2.plugin.CommandFinalizationData'".format(func.__name__))
+            raise TypeError("{} must have one parameter declared with type {}".format(func.__name__,
+                                                                                      plugin.CommandFinalizationData))
         if signature.return_annotation != plugin.CommandFinalizationData:
```
tleonhardt, line 4286: Should we be comparing both `param.annotation` and `signature.return_annotation` to `plugin.CommandFinalizationData`?

PR #960, opened 2020-07-21
Plugins follow-up. Adds plugin coverage. Minor cleanup.

anselor:
Added type annotations for fields in cmd2.Cmd.
Cleaned up docstring in external test plugin
Updated some initialization to match new approach for version info discovery.
Tagged some IDE-only lines as no-cover
Adds plugin coverage reporting.

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -200,7 +200,7 @@ class Cmd(cmd.Cmd):
         self.max_completion_items = 50
 
         # A dictionary mapping settable names to their Settable instance
-        self.settables = dict()
+        self.settables = dict()  # type: Dict[str, Settable]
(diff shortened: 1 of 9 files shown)


cmd2/cmd2.py
```diff
@@ -2694,16 +2694,31 @@ def do_help(self, args: argparse.Namespace) -> None:
 
     def _help_menu(self, verbose: bool = False) -> None:
         """Show a list of commands which help can be displayed for"""
+        cmds_cats, cmds_doc, cmds_undoc, help_topics = self._build_command_info()
```
tleonhardt, line 2697: Is there a reason we are changing cmd2.py code in this CR?

PR #972, opened 2020-08-11
Python352

anselor:
For Python 3.5.2 pinned nox to an older version that doesn't exercise bug typing bug introduced in 3.5.2.
Changed a couple type hints in utils.py to also not exercise the 3.5.2 typing bugs.

### DIFF
--- a/cmd2/utils.py
+++ b/cmd2/utils.py
@@ -13,7 +13,7 @@ import sys
 import threading
 import unicodedata
 from enum import Enum
-from typing import Any, Callable, Dict, Iterable, List, Optional, TextIO, Type, Union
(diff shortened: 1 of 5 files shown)


cmd2/utils.py
```diff
@@ -13,7 +13,7 @@
 import threading
 import unicodedata
 from enum import Enum
-from typing import Any, Callable, Dict, Iterable, List, Optional, TextIO, Type, Union
+from typing import Any, Callable, Dict, Iterable, List, Optional, TextIO, Union
```
tleonhardt, line 16: It would be extremely helpful if you can start documenting all the crap we can and should rip out once we deprecate support for 3.5 in the open issue for 3.5 deprecation.

PR #973, opened 2020-08-12
Commandset simplified

anselor:
Removed cmd2.Cmd parameter from CommandSet do/help/complete functions.
Added support for ns_provider to be in a CommandSet.

Relates to #943

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
     * Added explicit testing against python 3.5.2 for Ubuntu 16.04, and 3.5.3 for Debian 9
     * Added fallback definition of typing.Deque (taken from 3.5.4)
     * Removed explicit type hints that fail due to a bug in 3.5.2 favoring comment-based hints instead
+    * When passing a ns_provider to an argparse command, will now attempt to resolve the correct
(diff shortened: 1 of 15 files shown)


CHANGELOG.md
```diff
@@ -1,8 +1,15 @@
 ## 1.3.3 (TBD)
+* Breaking changes
```
tleonhardt, line 2: In the future whenever we release a big new feature like commands sets we should advertise that feature as being in some kind of alpha/beta/release-candidate phase where it is subject to change and keep it that way for a few minor releases until we feel it is stable.  We should come up with a formalized way of doing this.

cmd2/cmd2.py
```diff
@@ -4663,3 +4656,50 @@ def register_cmdfinalization_hook(self, func: Callable[[plugin.CommandFinalizati
         """Register a hook to be called after a command is completed, whether it completes successfully or not."""
         self._validate_cmdfinalization_callable(func)
         self._cmdfinalization_hooks.append(func)
+
+    def _resolve_func_self(self, cmd_support_func: Callable, cmd_self: Union[CommandSet, 'Cmd']) \
+            -> object:
```
tleonhardt, line 4661: Recommend not splitting line with a `\` but splitting at one of the arguments

PR #975, opened 2020-08-12
Fix tuple metavar crash in Cmd2HelpFormatter and ArgparseCompleter

### LINKED ISSUE #974: Bug in Cmd2HelpFormatter and ArgparseCompleter when metavar is a tuple
The [Cmd2HelpFormatter](https://github.com/python-cmd2/cmd2/blob/master/cmd2/argparse_custom.py#L600-L716) class has a bug which causes a crash in the following circumstances:
- `nargs` is an int and > 1
- [metavar](https://docs.python.org/3/library/argparse.html#metavar) is a tuple

Tested with `cmd2` 1.2.0 but verified the bug exists in the current version as well.

In my case, the crash occurs on line 780 in the current code with the [_format_args](https://github.com/python-cmd2/cmd2/blob/master/cmd2/argparse_custom.py#L763-L784) method.

kmvanbrunt:
Fixes #974

### DIFF
--- a/cmd2/argparse_custom.py
+++ b/cmd2/argparse_custom.py
@@ -733,7 +733,8 @@ class Cmd2HelpFormatter(argparse.RawTextHelpFormatter):
                 return ', '.join(action.option_strings) + ' ' + args_string
             # End cmd2 customization
 
-    def _metavar_formatter(self, action, default_metavar) -> Callable:
+    def _determine_metavar(self, action, default_metavar) -> Union[str, Tuple]:
+        """Custom method to determine what to use as the metavar value of an action"""
         if action.metavar is not None:
             result = action.metavar
         elif action.choices is not None:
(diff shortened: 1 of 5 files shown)


cmd2/argparse_custom.py
```diff
@@ -743,38 +744,46 @@ def _metavar_formatter(self, action, default_metavar) -> Callable:
...
+                return (metavar, ) * tuple_size
         return format
 
     # noinspection PyProtectedMember
     def _format_args(self, action, default_metavar) -> str:
-        get_metavar = self._metavar_formatter(action, default_metavar)
-        # Begin cmd2 customization (less verbose)
-        nargs_range = getattr(action, ATTR_NARGS_RANGE, None)
+        """Customized to handle ranged nargs and make other output less verbose"""
+        metavar = self._determine_metavar(action, default_metavar)
```
anselor, line 763: We're calling `self._determine_metavar()` twice in a row here. Maybe pass it into `self._metavar_formatter()` or have `self._metavar_formatter()` return a tuple?

cmd2/argparse_custom.py
```diff
@@ -743,38 +744,46 @@ def _metavar_formatter(self, action, default_metavar) -> Callable:
             # End cmd2 customization
         else:
             result = default_metavar
+        return result
+
+    def _metavar_formatter(self, action, default_metavar) -> Callable:
```
anselor, line 749: The name makes me think the callable returned here is the output formatter, but it's just trying to get the data into a standard data structure. Maybe call something like `_normalize_metavar()`?

cmd2/argparse_custom.py
```diff
@@ -743,38 +744,46 @@ def _metavar_formatter(self, action, default_metavar) -> Callable:
...
+
+        # Make this output less verbose. Do not customize the output when metavar is a
+        # tuple of strings. Allow argparse's formatter to handle that instead.
+        elif isinstance(metavar, str):
+            if action.nargs == ZERO_OR_MORE:
+                return '[%s [...]]' % metavar_formatter(1)
+            elif action.nargs == ONE_OR_MORE:
+                return '%s [...]' % metavar_formatter(1)
+            elif isinstance(action.nargs, int) and action.nargs > 1:
+                return '{}{{{}}}'.format('%s' % metavar_formatter(1), action.nargs)
```
anselor, line 784: Every call to `metavar_formatter()` passes in 1. Is something supposed to pass in something other than 1? If not, do we need that function argument?

cmd2/argparse_custom.py
kmvanbrunt: `self._metavar_formatter()` is an overrided `argparse` function. I can't change its signature.

cmd2/argparse_custom.py
kmvanbrunt: `self._metavar_formatter()` is an overrided `argparse` function. It expects the size of tuple being created as an argument.

cmd2/argparse_custom.py
kmvanbrunt: I can't rename this since it's an overrided `argparse` method.

anselor:
Looks good.

PR #977, opened 2020-08-13
Refactoring completion for 2.0.0 release

kmvanbrunt:
**Note: This PR is merging into 2.0 branch. Not master.**

* Breaking changes
    * Argparse Completion / Settables
        * Replaced `choices_function` / `choices_method` with `choices_provider`.
        * Replaced `completer_function` / `completer_method` with `completer`.
        * ArgparseCompleter now always passes `cmd2.Cmd` or `CommandSet` instance as the `self argument` to choices_provider and completer functions.
    * Moved `basic_complete` from utils into `cmd2.Cmd` class.
    * Moved `CompletionError` to exceptions.py

### DIFF
--- a/cmd2/argparse_completer.py
+++ b/cmd2/argparse_completer.py
@@ -24,8 +24,8 @@ from .argparse_custom import (
     generate_range_error,
 )
 from .command_definition import CommandSet
+from .exceptions import CompletionError
 from .table_creator import Column, SimpleTable
-from .utils import CompletionError, basic_complete
 
 # If no descriptive header is supplied, then this will be used instead
 DEFAULT_DESCRIPTIVE_HEADER = 'Description'
@@ -459,7 +459,7 @@ class ArgparseCompleter:
                 if action.help != argparse.SUPPRESS:
                     match_against.append(flag)
 
-        return basic_complete(text, line, begidx, endidx, match_against)
+        return self._cmd2_app.basic_complete(text, line, begidx, endidx, match_against)
 
     def _format_completions(self, arg_state: _ArgumentState, completions: List[Union[str, CompletionItem]]) -> List[str]:
         # Check if the results are CompletionItems and that there aren't too many to display
@@ -533,7 +533,7 @@ class ArgparseCompleter:
                     return completer.complete_subcommand_help(tokens[token_index:], text, line, begidx, endidx)
                 elif token_index == len(tokens) - 1:
                     # Since this is the last token, we will attempt to complete it
-                    return basic_complete(text, line, begidx, endidx, self._subcommand_action.choices)
+                    return self._cmd2_app.basic_complete(text, line, begidx, endidx, self._subcommand_action.choices)
                 else:
                     break
         return []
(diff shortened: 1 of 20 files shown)


examples/modular_commands/commandset_complex.py
```diff
@@ -42,7 +41,7 @@ def do_durian(self, args: List[str]):
         self._cmd.poutput(', '.join(['{}']*len(args)).format(*args))
 
     def complete_durian(self, text: str, line: str, begidx: int, endidx: int) -> List[str]:
-        return utils.basic_complete(text, line, begidx, endidx, ['stinks', 'smells', 'disgusting'])
+        return self._cmd.basic_complete(text, line, begidx, endidx, ['stinks', 'smells', 'disgusting'])
```
tleonhardt, line 44: It feels awkward to be accessing a protected member in example code.

kmvanbrunt: `_cmd` is a protected member of `CommandSet`. If we want to change it to public, then it should first be changed in 1.3 and merged into 2.0.

kmvanbrunt: @anselor What are your thoughts of making `CommandSet._cmd` public?

anselor: In general, protected members imply they're intended to be accessible by subclasses, while accessing it external to the class is discouraged. This is an example which extends CommandSet. The variable will be set or be None depending on whether the CommandSet is registered. Wasn't really intending it to be a public way to access to the cmd2 app.

If you really think they should be public, I guess that's fine. It'll just be None sometimes, depending on the state.

kmvanbrunt: I'm OK leaving it protected since it really is only meant to be used as @anselor described.

cmd2/argparse_completer.py
```diff
@@ -577,16 +577,15 @@ def _complete_for_arg(self, arg_state: _ArgumentState,
...
-                # need to find an instance at runtime that matches the types during declaration
-                cmd_set = self._cmd2_app._resolve_func_self(arg_choices.to_call, cmd_set)
-                if cmd_set is None:
-                    # No cases matched, raise an error
-                    raise CompletionError('Could not find CommandSet instance matching defining type for completer')
+            # The completer may or may not be defined in the same class as the command. Since completer
+            # functions are registered with the command argparser before anything is instantiated, we
+            # need to find an instance at runtime that matches the types during declaration
+            self_arg = self._cmd2_app._resolve_func_self(arg_choices.to_call, cmd_set)
+            if self_arg is None:
```
anselor, line 584: Do we still support standalone functions? If not, I think we should. If self is `None` here, we can pass the cmd2 app by default. This would require standalone functions to take `cmd2.Cmd` as the first parameter.

We also wouldn't need to move the utility completers if we do this.

Edit:
Nevermind, I look at the resolve method again. This should never be None.

cmd2/utils.py
```diff
@@ -129,14 +102,11 @@ def __init__(self, name: str, val_type: Callable, description: str, *,
...
-        :param completer_function: tab completion function that provides choices for this argument
-        :param completer_method: cmd2-app tab completion method that provides choices
-                                 for this argument (See note below)
+        :param choices_provider: function that provides choices for this argument
+        :param completer: tab completion function that provides choices for this argument
 
         Note:
-        For choices_method and completer_method, do not set them to a bound method. This is because
+        For choices_provider and completer, do not set them to a bound method. This is because
         ArgparseCompleter passes the self argument explicitly to these functions.
```
anselor, line 110: We mention this multiple times. I don't know if we have a place for this, but documenting the error complaining about parameter count may help someone debug this if they run into it.

PR #985, opened 2020-08-25
Added Cmd2AttributeWrapper class

kmvanbrunt:
* Breaking Changes
    * The functions cmd2 adds to Namespaces (`get_statement()` and `get_handler()`) are now `Cmd2AttributeWrapper` objects named `cmd2_statement` and `cmd2_handler`. This makes it easy to filter out which attributes in an `argparse.Namespace` were added by `cmd2`.
* Deprecations
    * ``Namespace.__statement__`` will be removed in `cmd2` 2.0.0. Use `Namespace.get_statement()` going forward.

### DIFF
--- a/cmd2/decorators.py
+++ b/cmd2/decorators.py
@@ -4,6 +4,7 @@ import argparse
 from typing import TYPE_CHECKING, Any, Callable, Dict, Iterable, List, Optional, Tuple, Union
 
 from . import constants
+from .argparse_custom import Cmd2AttributeWrapper
 from .exceptions import Cmd2ArgparseError
 from .parsing import Statement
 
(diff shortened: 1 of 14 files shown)


cmd2/decorators.py
```diff
@@ -300,11 +300,16 @@ def cmd_wrapper(*args: Any, **kwargs: Dict[str, Any]) -> Optional[bool]:
             else:
                 # Add statement to Namespace and a getter function for it
                 setattr(ns, constants.NS_ATTR_STATEMENT, statement)
-                setattr(ns, 'get_statement', lambda: statement)
+                setattr(ns, 'get_statement', Cmd2AttributeWrapper(statement))
```
tleonhardt, line 303: Would `Cmd2NamespaceWrapper` be a better name?

kmvanbrunt: It wraps an attribute that `cmd2` adds to the `Namespace`. Since it's callable, I've considered `Cmd2AttributeGetter`.

tleonhardt: Yeah I like `Cmd2AttributeGetter` better than I do the `Wrapper` name

kmvanbrunt: After talking to @anselor, I'm gonna make a change so it's no longer callable. Instead the class will implement `__get__()` and `__set()`. This will function like a normal `Namespace` attribute but will still be a `Cmd2AttributeWrapper` type so you can easily filter them out.

kmvanbrunt: Nevermind. `__get__()` and `__set__()` only work on class attributes, not instance attributes. Since we aren't adding our `Cmd2AttributeWrappers` to the `Namespace` as class attributes, this won't work. The other approach is to add `get()` and `set()` methods to `Cmd2AttributeWrapper`.

Then a developer would use it this way:
```Python
# Get the handler
handler = args.cmd2_handler.get()
handler(args)

# Overwrite the handler
args.cmd2_handler.set(new_handler)
```

PR #988, opened 2020-08-28
Removed need to set add_help to False for @as_subcommand_to decorator

kmvanbrunt:
* Bug Fixes
    * Fixed issue where subcommand added with `@as_subcommand_to` decorator did not display help when called with `-h/--help`.
* Enhancements
    * `add_help=False` no longer has to be passed to parsers used in `@as_subcommand_to` decorator.
      Only pass this if your subcommand should not have the `-h/--help` help option (as stated in argparse documentation).


anselor:
Looks good

PR #989, opened 2020-09-01
Run Shell without arguments

ralequi:
In some scenarios, you would like to allow the user to have a native (bash/zsh/sh) shell, rather than simply run a command.
I think it would be simpler for the final user to just run `shell` or `!` to directly enter into a native shell rather than `!bash` or `shell sh`

Please, if you find some easier way to do this (and, maybe, make it more flexible such as allow to modify the shell, etc) I would love to hear.
Also, I'm constrained to use the keyword "shell" to support this functionality, so create an alias such as "bash" would be nice but does not fits my current requirements :disappointed: .

Thank you !!

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -3375,7 +3375,7 @@ class Cmd(cmd.Cmd):
                 self.poutput(result_str)
 
     shell_parser = DEFAULT_ARGUMENT_PARSER(description="Execute a command as if at the OS prompt")
-    shell_parser.add_argument('command', help='the command to run', completer_method=shell_cmd_complete)
+    shell_parser.add_argument('command', nargs='?', default='bash', help='the command to run', completer_method=shell_cmd_complete)
     shell_parser.add_argument('command_args', nargs=argparse.REMAINDER, help='arguments to pass to command',
                               completer_method=path_complete)
 


cmd2/cmd2.py
```diff
@@ -3375,7 +3375,7 @@ def do_set(self, args: argparse.Namespace) -> None:
                 self.poutput(result_str)
 
     shell_parser = DEFAULT_ARGUMENT_PARSER(description="Execute a command as if at the OS prompt")
-    shell_parser.add_argument('command', help='the command to run', completer_method=shell_cmd_complete)
+    shell_parser.add_argument('command', nargs='?', default='bash', help='the command to run', completer_method=shell_cmd_complete)
```
tleonhardt, line 3378: cmd2 is a fully cross-platform framework which supports Windows, macOS, and Linux.  It is not OK to make an assumption like this that won’t work on Windows.

ralequi: Thank for your comments.
Yes, this edit is not cross-platform... the bash should be replaced somehow with the default OS shell. Don't know how to do that, and either have a Mac to test. Didn't notice that the project was multiplatform, my apologizes.

The issue creating an alias such as "shell=shell bash" would imply that when someone wants to run a command over a shell it will spawn a bash and then run the command on that bash.... I don't like that much, but probably the best/easiest solution I could get.

ralequi: BTW, I can't create an alias such as `shell=shell bash`...
What I get there is: `Alias cannot have the same name as a command`

tleonhardt: aliases are intended for end users and prevent overwriting commands.  But as an application developer it is easy to create an alias of a command.

PR #991, opened 2020-09-01
Read input enhancements

### LINKED ISSUE #982: Enhance read_input()
Add following feature to `cmd2.Cmd.read_input()`.

1. Ability to tab complete custom values
    * Support passing in list of choices
    * Support passing in argparse.ArgumentParser which tab completes similar to our commands
    * Preserve support for doing command tab completion since this is called by `cmd2.Cmd._read_command_line()`.
    * Default will be no tab completion

2. Support passing in custom up-arrow history

We need to hammer out an interface to `read_input()`. This will be a 2.0 feature since its signature will change.

kmvanbrunt:
Added support for custom tab completion and up-arrow input history to `cmd2.Cmd2.read_input()`
Closes #982

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -1618,7 +1618,10 @@ class Cmd(cmd.Cmd):
         :param endidx: the ending index of the prefix text
         :param custom_settings: optional prepopulated completion settings
         """
+        from .argparse_completer import ArgparseCompleter
+
         unclosed_quote = ''
+        command = None
 
         # If custom_settings is None, then we are completing a command's arguments
         if custom_settings is None:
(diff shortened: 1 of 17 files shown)


cmd2/cmd2.py
```diff
@@ -95,6 +97,40 @@
     ipython_available = False
 
 
+class CompletionMode(enum.Enum):
```
tleonhardt, line 100: Do these two new classes belong in cmd2.py?  Not sure.  But given the size of cmd2.py, consider a potentially better location.

tleonhardt (approved):

cmd2/cmd2.py
kmvanbrunt: I moved them to utils.py

PR #994, opened 2020-09-05
Group flags in argparse tab completion

kmvanbrunt:
* Enhancements
    * argparse tab completion now groups flag names which run the same action. Optional flags are wrapped in brackets like it is done in argparse usage text.

This still needs unit tests, but I want to make sure the new behavior is agreed upon first.

To exercise this, just tab complete flags on any command. Common flags will be grouped together. Optional flags will be in brackets. The history command is a good one to test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
     * Added user-settable option called `always_show_hint`. If True, then tab completion hints will always
     display even when tab completion suggestions print. Arguments whose help or hint text is suppressed will
     not display hints even when this setting is True.
+    * argparse tab completion now groups flag names which run the same action. Optional flags are wrapped
+    in brackets like it is done in argparse usage text.
 * Bug Fixes
     * Fixed issue where flag names weren't always sorted correctly in argparse tab completion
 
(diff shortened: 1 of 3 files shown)


tleonhardt:
I think this is a great enhancement.  I experimented with it and think its a very nice improvement.  

One thing that it would be cool to add if it isn't too hard is display of mutually-exclusive arguments , perhaps separated by pipes within the same group or something similar?  The `history` command would also be a great candidate to test that with.

kmvanbrunt:
@tleonhardt I can't easily group the mutually-exclusive arguments in this release. It would end up with some ugly wrapping if there was a long chain of pipe-delimited flags.

For 2.0, I am working on supporting more elaborate tables in tab completion. This would include the support of newlines and text style. Since we rely on readline to print our tables in the 1.0 branch, we cannot currently do this. Once that feature is added, I'll update the 2.0 branch to better organize flag names in tab completion.

tleonhardt:
@kmvanbrunt Your plan for updating further in 2.0 sounds great.

tleonhardt (approved):

anselor (approved):

anselor (approved):

PR #999, opened 2020-09-11
Added way to silence alias/macro creation

kmvanbrunt:
Added `--silent` flag to `alias/macro create`.
Added `--with_silent` flag to `alias/macro list`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@
     not display hints even when this setting is True.
     * argparse tab completion now groups flag names which run the same action. Optional flags are wrapped
     in brackets like it is done in argparse usage text.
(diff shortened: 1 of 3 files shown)


tleonhardt (dismissed):

anselor (commented):
Is this really needed? Can't you just pipe the output to /dev/null in the startup script?

anselor (dismissed):

kmvanbrunt:
@anselor This feature is a convenience when making alias startup scripts. Instead of manually adding `> /dev/null` to each line, you can just run `alias list --with_silent > aliases.txt`. That will output all aliases in the syntax `alias create --silent name value` to the alias file.

PR #1006, opened 2020-10-01
Remove unnecessary runtime dependency on setuptools

mgorny:
The package does not seem to use setuptools or pkg_resources anywhere,
and the dependency is probably leftover after migration to
importlib.metadata.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,14 +31,13 @@ Programming Language :: Python :: Implementation :: CPython
 Topic :: Software Development :: Libraries :: Python Modules
 """.splitlines())))  # noqa: E128
 
-SETUP_REQUIRES = ['setuptools_scm >= 3.0']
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -31,14 +31,13 @@
 Topic :: Software Development :: Libraries :: Python Modules
 """.splitlines())))  # noqa: E128
 
-SETUP_REQUIRES = ['setuptools_scm >= 3.0']
+SETUP_REQUIRES = ['setuptools >= 34.4', 'setuptools_scm >= 3.0']
```
tleonhardt, line 34: Good call, after moving to `importlib.metadata`/`importlib_metadata` we don't need a runtime dependency on `setuptools` anymore.  

Thanks for the PR!

tleonhardt (approved):

PR #1011, opened 2020-10-31
Fixed error in set --long, which should be set -v

KyleKing:
I think I found a minor bug in the documentation:

```
(Cmd) set --long
Usage: set [-h] [-v] [param] [value]
Error: unrecognized arguments: --long
```

Should be:

```
(Cmd) set -v | grep sunny
sunny: False              # Is it sunny outside?
```

### DIFF
--- a/docs/features/settings.rst
+++ b/docs/features/settings.rst
@@ -134,9 +134,9 @@ changes a setting, and will receive both the old value and the new value.
 
 .. code-block:: text
 
-   (Cmd) set --long | grep sunny
+   (Cmd) set -v | grep sunny
    sunny: False                # Is it sunny outside?
-   (Cmd) set --long | grep degrees
+   (Cmd) set -v | grep degrees
    degrees_c: 22               # Temperature in Celsius
    (Cmd) sunbathe
    Too dim.
--- a/docs/features/settings.rst
+++ b/docs/features/settings.rst
@@ -134,9 +134,9 @@ changes a setting, and will receive both the old value and the new value.
 
 .. code-block:: text
 
-   (Cmd) set -v | grep sunny
+   (Cmd) set --verbose | grep sunny
    sunny: False                # Is it sunny outside?
-   (Cmd) set -v | grep degrees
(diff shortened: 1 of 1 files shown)


docs/features/settings.rst
```diff
@@ -134,9 +134,9 @@ changes a setting, and will receive both the old value and the new value.
 
 .. code-block:: text
 
-   (Cmd) set --long | grep sunny
+   (Cmd) set -v | grep sunny
```
kmvanbrunt, line 137: I'm thinking we should use the long option name here (--verbose) to be more explicit in the documentation.
@tleonhardt Thoughts?

kmvanbrunt:
@KyleKing Thanks for the documentation correction.

docs/features/settings.rst
KyleKing: Yeah, I agree that the full option name would be better. I'll go ahead and make the change

I also noticed that there were some references to `with_argparser_and_unknown_args` that I think should be `with_argparser(parser, with_unknown_args=True)`

KyleKing:
Also, I just want to say that the docs have been great! I read or skimmed most of the documentation yesterday and left feeling like I had a good understanding of how cmd2 works and how to build applications

I evaluated several alternatives (clint, cliff, click, cement, prompt-toolkit, etc.) for a class project and ended up finding that cmd2 was the best fit. Finding [alternatives](https://cmd2.readthedocs.io/en/latest/overview/alternatives.html) earlier would have saved me some time 😆

tleonhardt (approved):

docs/features/settings.rst
tleonhardt: @kmvanbrunt  Yeah I like the `--verbose` because it makes it very clear what the intent is.

@KyleKing Good point about the outdated `with_argparser_and_unknown_args` docs.  That is an older form that is now deprecated.  You are correct that it should be `with_argparser(parser, with_unknown_args=True)`.  If you have the time and inclination, please submit another PR to fix that ;-)

Issue #1016, opened 2020-11-16
Traceback on import cmd2 in centos:7 docker container


gmmephisto:
Hi! I've got an error while running tests using `stestr` tool for some OpenStack project in fresh `centos:7` docker container. There is an original traceback:
```
upstream run-test: commands[1] | stestr run --concurrency 2
Traceback (most recent call last):
  File "/builds/.../.tox/upstream/bin/stestr", line 5, in <module>
    from stestr.cli import main
  File "/builds/.../.tox/upstream/lib/python3.6/site-packages/stestr/cli.py", line 15, in <module>
    from cliff import app
  File "/builds/.../.tox/upstream/lib/python3.6/site-packages/cliff/app.py", line 24, in <module>
    import cmd2
  File "/builds/.../.tox/upstream/lib/python3.6/site-packages/cmd2/__init__.py", line 30, in <module>
    from .cmd2 import Cmd
  File "/builds/.../.tox/upstream/lib/python3.6/site-packages/cmd2/cmd2.py", line 121, in <module>
    class Cmd(cmd.Cmd):
  File "/builds/.../.tox/upstream/lib/python3.6/site-packages/cmd2/cmd2.py", line 129, in Cmd
    DEFAULT_EDITOR = utils.find_editor()
  File "/builds/.../.tox/upstream/lib/python3.6/site-packages/cmd2/utils.py", line 385, in find_editor
    if which(editor):
  File "/builds/.../.tox/upstream/lib/python3.6/site-packages/cmd2/utils.py", line 201, in which
    exe_path = subprocess.check_output(['which', exe_name], stderr=subprocess.STDOUT).strip()
  File "/usr/lib64/python3.6/subprocess.py", line 336, in check_output
    **kwargs).stdout
  File "/usr/lib64/python3.6/subprocess.py", line 403, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/usr/lib64/python3.6/subprocess.py", line 709, in __init__
    restore_signals, start_new_session)
  File "/usr/lib64/python3.6/subprocess.py", line 1344, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'which': 'which'
```
Looks like `stestr` uses `cmd2` under the hood and `cmd2` tries to detect the default editor using `which` tool, but can not find it. And `EDITOR` env variable is not set in the container by default.

Steps to reproduce the:
[...]

Issue #1017, opened 2020-11-16
Incorrect detection of default editor in ubuntu:bionic docker container


gmmephisto:
Hi! I found a corner case in the default editor detection where neither `EDITOR` variable set nor any editor installed in the system. The excellent environment to reproduce this case is `ubuntu:bionic` docker container.

Steps to reproduce:
* `$ docker run -ti --rm ubuntu:bionic`
* `(container) # apt update && apt install python3 python3-pip`
* `(container) # pip3 install cmd2`
* `(container) # python3 -c "import cmd2; print(cmd2.Cmd.DEFAULT_EDITOR)`

Result:
```
root@820b205efe1f:/# python3 -c "import cmd2; print(cmd2.Cmd.DEFAULT_EDITOR)"
atom
```

Expected result:
* Empty value ?
* Raise an EnvironmentError ?

tleonhardt:
This was closed by #1018

PR #1018, opened 2020-11-16
feat(utils): probe editors in system path

gmmephisto:
Closes: #1016 #1017


tleonhardt (approved):
Changes LGTM.  Is the fundamental problem that `which` is not available out-of-the-box on some modern Linux systems?

tleonhardt:
@gmmephisto Thanks for the PR!  I never thought I would see the day where a POSIX OS ships without `which`!

PR #1021, opened 2020-11-19
Updated utils.find_editor() to include more Windows editors

kmvanbrunt:


cmd2/utils.py
```diff
@@ -357,32 +357,30 @@ def expand_user_in_tokens(tokens: List[str]) -> None:
...
+    :return: Default editor or None
+    """
     editor = os.environ.get('EDITOR')
     if not editor:
         if sys.platform[:3] == 'win':
-            editor = 'notepad'
+            editors = ['code.cmd', 'notepad++.exe', 'notepad.exe']
         else:
-            editor = probe_editors()
+            editors = ['vim', 'vi', 'emacs', 'nano', 'pico', 'gedit', 'kate', 'subl', 'geany', 'atom']
```
tleonhardt, line 371: Recommend adding `code` for VSCode to Linux/Mac list as well.

tleonhardt (dismissed):

cmd2/utils.py
kmvanbrunt: I added it right before subl. Is that OK?

tleonhardt: yup thats great

tleonhardt (approved):

tests/test_utils.py
```diff
@@ -641,55 +641,18 @@ def test_str_to_bool_bad_input():
...
-
-    with mock.patch.dict(os.environ, {'PATH': str(path)}):
-        with mock.patch(
-            'cmd2.utils.is_executable',
-            mock_is_executable
-        ):
-            editor = cu.probe_editors()
-    assert editor == vi_path
+
+def test_find_editor_not_specified():
```
tleonhardt, line 650: Are we losing any code coverage by simplifying the unit tests like this?

tleonhardt (approved):

cmd2/utils.py
```diff
@@ -357,32 +357,30 @@ def expand_user_in_tokens(tokens: List[str]) -> None:
...
+    :return: Default editor or None
+    """
     editor = os.environ.get('EDITOR')
     if not editor:
         if sys.platform[:3] == 'win':
-            editor = 'notepad'
+            editors = ['code.cmd', 'notepad++.exe', 'notepad.exe']
         else:
-            editor = probe_editors()
+            editors = ['vim', 'vi', 'emacs', 'nano', 'pico', 'gedit', 'kate', 'code', 'subl', 'geany', 'atom']
```
anselor, line 371: Since we're adding a bunch of editors - can we add `joe` to the list for linux?

tleonhardt: Seems reasonable to me.  Should we add it right after say `pico`?

tests/test_utils.py
```diff
@@ -641,55 +641,18 @@ def test_str_to_bool_bad_input():
...
+    # Use existing path env setting. Something in the editor list should be found.
+    editor = cu.find_editor()
+    assert editor
+
+    # Overwrite path env setting with invalid path. No editor should be found.
+    with mock.patch.dict(os.environ, {'PATH': 'fake_dir'}):
+        editor = cu.find_editor()
+    assert editor is None
```
tleonhardt, line 658: This test is failing for me when manually testing on macOS and needs to be refactored.

tleonhardt (changes requested):

cmd2/utils.py
tleonhardt: I added `joe` and updated the order of the graphical editors at the end to be consistent with my recent experience.

kmvanbrunt:
@tleonhardt @anselor While I understand the intention of adding more editors to the Linux/Mac list, the algorithm is set up to return the first one found. The chances that `find_editor()` will even get past vim is very low. This function only sets the default editor. The user is always free to run `set editor joe` in a startup file.

tests/test_utils.py
kmvanbrunt: I don't have a Mac to test on. Manual tests in Linux are working fine.

tleonhardt: It was failing because I have the `EDITOR` environment variable set on my system.  I fixed the test so it is robust to that.

Issue #952, opened 2020-07-10
Figure out why CI system can't use filtered git version tags


anselor:
For the ext_test plugin, the setup.py now has the version hard-coded to 0.2.0 and will require a manual update for each new version. The issue is the custom filters for `git describe` aren't executing correctly on the CI nodes. It may be possible to switch out how it does the tag filtering depending on whether we're running it on our system vs the CI system. We only really need the tag filtering to work when we invoke the task to publish to pypi and tag it. Doesn't matter in the context of the CI reporting.

anselor:
More of a note to self - it's possible that we need to specify newer versions:
setuptools >= 42
setuptools_scm >= 3.4.

anselor:
Resolved via #1034

PR #1034, opened 2021-01-05
Plugin version

anselor:
Updates our plugin setup.py to use git tag for setting the library version. Needed to specify a newer version of setuptools to make this work.


tleonhardt (approved):

tleonhardt:
We need to fix unit tests on Python 3.8 and 3.9, other than that this looks OK

anselor:
This closes #952

PR #1038, opened 2021-01-11
Added py.typed to signal that we support type annotations for PEP-561

### LINKED ISSUE #1037: PEP 561 Compliance
We added type declarations but didn't meet PEP 561 so mypy validation against anything that uses cmd2 causes a lot of warnings.

anselor:
…. Fixes #1037


cmd2/py.typed
```diff
@@ -0,0 +1 @@
+# PEP 561
```
tleonhardt, line 1: So what does adding this file with this comment do?

setup.py
```diff
@@ -82,6 +86,7 @@
     url='https://github.com/python-cmd2/cmd2',
     license='MIT',
     platforms=['any'],
+    package_data=PACKAGE_DATA,
```
tleonhardt, line 89: What effect does adding this have?

cmd2/py.typed
anselor: The comment doesn't matter. But the file is what's defined in PEP-561 to signal that we support type annotations. Without this, any project that depends on us will get a bunch of warnings and errors using mypy because it'll assume everything from our package is coming in as `Any`.

setup.py
anselor: This causes py.typed to be included in the wheel.

tleonhardt (dismissed):

tleonhardt:
LGTM.  We need to fix unit test on Python 3.8 and 3.9, but this PR looks fine.

kmvanbrunt (approved):

PR #1045, opened 2021-01-22
Added option to run startup scripts silently

### LINKED ISSUE #1010: Please consider the ability to suppress diagnostics output of "set" command
Currently when a Settable's value is changed via "set" or via startup script, there is a small amount of diagnostics output that is always generated regardless of the "quiet" setting.  For example:

> set root '/.more-space'
root - was: '/.space'
now: '/.more-space'

This is somewhat jarring if we use a lot of "set" commands in startup script.  Would you please consider suppressing this output?

kmvanbrunt:
Closes #1010


anselor (approved):

tleonhardt (dismissed):
Strongly recommend updating one of the examples which use `startup_script` to also use this new option to help with discoverability - [basic.py](https://github.com/python-cmd2/cmd2/blob/master/examples/basic.py) would be a good candidate

tleonhardt (approved):

PR #1049, opened 2021-01-28
Fixed width calculation when table column headers had tabs

kmvanbrunt:
Also fixed `FileNotFoundError` if you ran back-to-back `history --clear` commands when using a history file.

### DIFF
--- a/cmd2/table_creator.py
+++ b/cmd2/table_creator.py
@@ -83,7 +83,7 @@ class Column:
         :param header: label for column header
         :param width: display width of column. This does not account for any borders or padding which
                       may be added (e.g pre_line, inter_cell, and post_line). Header and data text wrap within
-                      this width using word-based wrapping (defaults to width of header or 1 if header is blank)
+                      this width using word-based wrapping (defaults to actual width of header or 1 if header is blank)
         :param header_horiz_align: horizontal alignment of header cells (defaults to left)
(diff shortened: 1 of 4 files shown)


cmd2/table_creator.py
```diff
@@ -141,8 +136,20 @@ def __init__(self, cols: Sequence[Column], *, tab_width: int = 4) -> None:
...
+            # For headers with the width not yet set, use the width of the
+            # widest line in the header or 1 if the header has no width
+            if col.width is None:
+                line_widths = [ansi.style_aware_wcswidth(line) for line in col.header.splitlines()]
+                line_widths.append(1)
+                col.width = max(line_widths)
+
     @staticmethod
-    def _wrap_long_word(word: str, max_width: int, max_lines: Union[int, float], is_last_word: bool) -> Tuple[str, int, int]:
+    def _wrap_long_word(word: str, max_width: int, max_lines: Union[int, float],
```
tleonhardt, line 151: I haven't tested it yet, but I read recently that instead of `Union` you can use syntax like the folliwing:
```Python
def _wrap_long_word(word: str, max_width: int, max_lines: int | float):
```

I'm not sure if it requires a newer version of Python or not, but sometime we should try it out - I think it is a lot cleaner and more readable than using `Union[int, float]` and is also easier when people are moving between Python and TypeScript.

tleonhardt (approved):

PR #1065, opened 2021-03-01
History fix

### LINKED ISSUE #1064: History repeats index numbers
Once a history file reaches the max limit of items, cmd2 truncates it before closing. When you reopen cmd2, the history file is read and all `HistoryItems` are restored with their previous index numbers. Then `cmd2.History` sets the starting index based on the length of the list, which could be an index already used by a restored `HistoryItem`.

I propose removing the index from `HistoryItem` and instead use its actual index in the `History` list (converted to 1-based as it has been).

kmvanbrunt:
Fixes #1064

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -787,7 +787,7 @@ class Cmd(cmd.Cmd):
                             if choice_name == cur_subcmd:
                                 return find_subcommand(choice, subcmd_names)
                         break
-                raise CommandSetRegistrationError('Could not find sub-command "{}"'.format(full_command_name))
+                raise CommandSetRegistrationError('Could not find subcommand "{}"'.format(full_command_name))
 
             target_parser = find_subcommand(command_parser, subcommand_names)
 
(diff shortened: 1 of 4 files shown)


cmd2/cmd2.py
```diff
@@ -4230,13 +4230,13 @@ def do_history(self, args: argparse.Namespace) -> Optional[bool]:
                 self.perror("Cowardly refusing to run all previously entered commands.")
                 self.perror("If this is what you want to do, specify '1:' as the range of history.")
             else:
-                return self.runcmds_plus_hooks(history)
+                return self.runcmds_plus_hooks(list(history.values()))
```
tleonhardt, line 4233: Something just smells off to me.  The `History` class is still just a `List[HistoryItem]` with some added methods.  But the `cmd2.Cmd._get_history()` method returns a `Dict[int, HistoryItem]` and does so using the `self.history` History list.  

It feels awkward and like we don't have a data structure or algorithm right somewhere.

Should the `History` class be modified to be a dictionary?  Or should we add a convenience method to it so we don't need this method that returns an intermediate dicionary and then re-converts that back to a list?

tleonhardt: Ok, I've slept on this and don't have any brilliant ideas.

Perhaps if we just encapsulated more of the weird functionality within the `History` class it would feel less awkward?  Not sure.  What do you think?

tleonhardt (approved):

cmd2/cmd2.py
kmvanbrunt: Yeah, I couldn't think of a better structure either. We need the results and we need to know what their indexes are. Since the user is searching all of the history, results won't be necessarily be contiguous from the main list. At least the dictionary makes it easy to treat the values as a list. And thankfully the dictionaries are ordered by index numbers.

PR #1074, opened 2021-03-15
CommandSet Settables

### LINKED ISSUE #965: Allow CommandSets to have their own user-settable settings
Currently `do_set` is limited to attributes of `cmd2.Cmd` classes. Make it possible for attributes of `CommandSet` to also be settable.

anselor:
- Each CommandSet has its own settables.
- Renamed attribute settables to _settables and added a new settables property that combines cmd2's _settables with the ones from all CommandSets
- New cmd2.Cmd attribute always_prefix_settables will prepend all settables from CommandSets with a settable_prefix provided by each CommandSet
- Settable now allows you to route to any attribute on any object - it doesn't have to be an attribute of Cmd or CommandSet. It also allows the name presented to the user to be different from the name of the underlying attribute.

Also make some progress addressing mypy validation issues in a few files. 

Closes #965

### DIFF
--- a/cmd2/ansi.py
+++ b/cmd2/ansi.py
@@ -230,7 +230,7 @@ def style_aware_write(fileobj: IO[str], msg: str) -> None:
     fileobj.write(msg)
 
 
-def fg_lookup(fg_name: Union[str, fg]) -> Fore:
+def fg_lookup(fg_name: Union[str, fg]) -> str:
     """
     Look up ANSI escape codes based on foreground color name.
 
@@ -239,16 +239,16 @@ def fg_lookup(fg_name: Union[str, fg]) -> Fore:
     :raises: ValueError: if the color cannot be found
     """
     if isinstance(fg_name, fg):
-        return fg_name.value
+        return str(fg_name.value)
 
     try:
         ansi_escape = fg[fg_name.lower()].value
     except KeyError:
         raise ValueError('Foreground color {!r} does not exist; must be one of: {}'.format(fg_name, fg.colors()))
-    return ansi_escape
+    return str(ansi_escape)
 
 
-def bg_lookup(bg_name: Union[str, bg]) -> Back:
+def bg_lookup(bg_name: Union[str, bg]) -> str:
     """
     Look up ANSI escape codes based on background color name.
 
@@ -257,13 +257,13 @@ def bg_lookup(bg_name: Union[str, bg]) -> Back:
     :raises: ValueError: if the color cannot be found
     """
     if isinstance(bg_name, bg):
-        return bg_name.value
+        return str(bg_name.value)
 
     try:
         ansi_escape = bg[bg_name.lower()].value
     except KeyError:
         raise ValueError('Background color {!r} does not exist; must be one of: {}'.format(bg_name, bg.colors()))
-    return ansi_escape
+    return str(ansi_escape)
 
 
 # noinspection PyShadowingNames
@@ -292,13 +292,13 @@ def style(
     :return: the stylized string
     """
     # List of strings that add style
-    additions = []
+    additions: List[str] = []
 
     # List of strings that remove style
-    removals = []
+    removals: List[str] = []
 
     # Convert the text object into a string if it isn't already one
-    text = "{}".format(text)
+    text_formatted = "{}".format(text)
 
     # Process the style settings
     if fg:
@@ -322,7 +322,7 @@ def style(
         removals.append(UNDERLINE_DISABLE)
 
     # Combine the ANSI style sequences with the text
-    return cast(str, "".join(additions) + text + "".join(removals))
+    return "".join(additions) + text_formatted + "".join(removals)
 
 
 # Default styles for printing strings of various types.
--- a/cmd2/ansi.py
+++ b/cmd2/ansi.py
@@ -13,15 +13,16 @@ from typing import (
     Any,
     List,
     Union,
+    cast,
 )
 
-import colorama
+import colorama  # type: ignore [import]
 from colorama import (
     Back,
     Fore,
     Style,
 )
-from wcwidth import (
+from wcwidth import (  # type: ignore [import]
     wcswidth,
 )
 
@@ -86,14 +87,14 @@ class ColorBase(Enum):
         Support building a color string when self is the left operand
         e.g. fg.blue + "hello"
         """
-        return str(self) + other
+        return cast(str, str(self) + other)
 
     def __radd__(self, other: Any) -> str:
         """
         Support building a color string when self is the right operand
         e.g. "hello" + fg.reset
         """
-        return other + str(self)
+        return cast(str, other + str(self))
 
     @classmethod
     def colors(cls) -> List[str]:
@@ -194,7 +195,7 @@ def style_aware_wcswidth(text: str) -> int:
              then this function returns -1. Replace tabs with spaces before calling this.
     """
     # Strip ANSI style sequences since they cause wcswidth to return -1
-    return wcswidth(strip_style(text))
+    return cast(int, wcswidth(strip_style(text)))
 
 
 def widest_line(text: str) -> int:
@@ -217,7 +218,7 @@ def widest_line(text: str) -> int:
     return max(lines_widths)
 
 
-def style_aware_write(fileobj: IO, msg: str) -> None:
+def style_aware_write(fileobj: IO[str], msg: str) -> None:
     """
     Write a string to a fileobject and strip its ANSI style sequences if required by allow_style setting
 
@@ -229,7 +230,7 @@ def style_aware_write(fileobj: IO, msg: str) -> None:
     fileobj.write(msg)
 
 
-def fg_lookup(fg_name: Union[str, fg]) -> str:
+def fg_lookup(fg_name: Union[str, fg]) -> Fore:
     """
     Look up ANSI escape codes based on foreground color name.
 
@@ -247,7 +248,7 @@ def fg_lookup(fg_name: Union[str, fg]) -> str:
     return ansi_escape
 
 
-def bg_lookup(bg_name: Union[str, bg]) -> str:
+def bg_lookup(bg_name: Union[str, bg]) -> Back:
     """
     Look up ANSI escape codes based on background color name.
 
@@ -321,7 +322,7 @@ def style(
         removals.append(UNDERLINE_DISABLE)
 
     # Combine the ANSI style sequences with the text
-    return "".join(additions) + text + "".join(removals)
+    return cast(str, "".join(additions) + text + "".join(removals))
 
 
 # Default styles for printing strings of various types.
@@ -400,4 +401,4 @@ def set_title_str(title: str) -> str:
     :param title: new title for the window
     :return: string to write to sys.stderr in order to set the window title to the desired test
     """
-    return colorama.ansi.set_title(title)
+    return cast(str, colorama.ansi.set_title(title))
--- a/cmd2/argparse_custom.py
+++ b/cmd2/argparse_custom.py
@@ -203,12 +203,15 @@ from gettext import (
     gettext,
 )
 from typing import (
+    IO,
     Any,
     Callable,
+    Dict,
     Iterable,
     List,
     NoReturn,
     Optional,
+    Sequence,
     Tuple,
     Type,
     Union,
@@ -219,6 +222,15 @@ from . import (
     ansi,
     constants,
 )
+
+try:
+    from typing import (
+        Protocol,
+    )
+except ImportError:
+    from typing_extensions import (  # type: ignore[misc]
+        Protocol,
+    )
 
 ############################################################################################################
 # The following are names of custom argparse argument attributes added by cmd2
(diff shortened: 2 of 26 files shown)


cmd2/cmd2.py
```diff
@@ -575,6 +580,15 @@ def register_command_set(self, cmdset: CommandSet) -> None:
         if type(cmdset) in existing_commandset_types:
             raise CommandSetRegistrationError('CommandSet ' + type(cmdset).__name__ + ' is already installed')
 
+        if self.always_prefix_settables:
+            if len(cmdset.settable_prefix.strip()) == 0:
```
kmvanbrunt, line 584: Can this be `if not cmdset.settable_prefix.strip()` instead?

cmd2/cmd2.py
```diff
@@ -575,6 +580,15 @@ def register_command_set(self, cmdset: CommandSet) -> None:
...
+                raise CommandSetRegistrationError('CommandSet settable prefix must not be empty')
+        else:
+            all_settables = self.settables
+            for key in cmdset.settables.keys():
```
kmvanbrunt, line 588: Can this be combined into the following?
`if any(key in all_settables for key in cmdset.settables.keys()):`

cmd2/cmd2.py
```diff
@@ -888,13 +902,30 @@ def _unregister_subcommands(self, cmdset: Union[CommandSet, 'Cmd']) -> None:
                     action.remove_parser(subcommand_name)
                     break
 
+    @property
+    def settables(self) -> Mapping[str, Settable]:
```
kmvanbrunt, line 906: This needs a docstring

cmd2/cmd2.py
```diff
@@ -888,13 +902,30 @@ def _unregister_subcommands(self, cmdset: Union[CommandSet, 'Cmd']) -> None:
...
+
     def add_settable(self, settable: Settable) -> None:
         """
         Convenience method to add a settable parameter to ``self.settables``
 
         :param settable: Settable object being added
         """
-        self.settables[settable.name] = settable
+        if not self.always_prefix_settables:
+            if settable.name in self.settables.keys() and settable.name not in self._settables.keys():
```
kmvanbrunt, line 924: Doesn't `self.settables` also include `self._settables`? Can the name be in one and not the other?

cmd2/cmd2.py
```diff
@@ -3712,29 +3743,23 @@ def do_set(self, args: argparse.Namespace) -> None:
         if args.param:
             try:
                 settable = self.settables[args.param]
+                if settable.settable_obj is None:
```
kmvanbrunt, line 3746: Why would settable_obj be None?

cmd2/command_definition.py
```diff
@@ -126,3 +133,36 @@ def on_unregistered(self) -> None:
...
+        Convenience method to add a settable parameter to the CommandSet
+
+        :param settable: Settable object being added
+        """
+        if self._cmd and not self._cmd.always_prefix_settables:
+            if settable.name in self._cmd.settables.keys() and settable.name not in self._settables.keys():
+                raise KeyError(f'Duplicate settable: {settable.name}')
+        if settable.settable_obj is None:
+            settable.settable_obj = self
+        self._settables[settable.name] = settable
```
kmvanbrunt, line 156: Does this consider the case where two CommandSets used the same prefix and the same setting name?

cmd2/utils.py
```diff
@@ -93,18 +97,20 @@ def str_to_bool(val: str) -> bool:
...
         *,
-        onchange_cb: Callable[[str, Any, Any], Any] = None,
-        choices: Iterable = None,
-        choices_provider: Optional[Callable] = None,
-        completer: Optional[Callable] = None
+        settable_object: Optional[object] = None,
+        settable_attrib_name: Optional[str] = None,
+        onchange_cb: Optional[Callable[[str, Any, Any], Any]] = None,
+        choices: Optional[Iterable[Any]] = None,
+        choices_provider: Optional[Callable[[], List[str]]] = None,
```
kmvanbrunt, line 112: Same issue about arg_tokens dictionary. See above.

cmd2/utils.py
```diff
@@ -93,18 +97,20 @@ def str_to_bool(val: str) -> bool:
...
+        completer: Optional[Callable[[str, str, int, int], List[str]]] = None
```
kmvanbrunt, line 113: Same issue about arg_tokens dictionary. See above.

cmd2/utils.py
```diff
@@ -93,18 +97,20 @@ def str_to_bool(val: str) -> bool:
...
         name: str,
-        val_type: Callable,
+        val_type: Union[Type[Any], Callable[[Any], Any]],
         description: str,
         *,
-        onchange_cb: Callable[[str, Any, Any], Any] = None,
-        choices: Iterable = None,
-        choices_provider: Optional[Callable] = None,
-        completer: Optional[Callable] = None
+        settable_object: Optional[object] = None,
```
kmvanbrunt, line 108: What is the use case for this being None?

cmd2/utils.py
```diff
@@ -210,13 +213,16 @@ def is_text_file(file_path: str) -> bool:
...
+
+
+def remove_duplicates(list_to_prune: List[_T]) -> List[_T]:
     """Removes duplicates from a list while preserving order of the items.
 
     :param list_to_prune: the list being pruned of duplicates
     :return: The pruned list
     """
-    temp_dict = collections.OrderedDict()
+    temp_dict: collections.OrderedDict[_T, Any] = collections.OrderedDict()
```
kmvanbrunt, line 225: We can remove OrderedDict usage now that we are on Python 3.6.

cmd2/utils.py
```diff
@@ -441,7 +447,7 @@ class StdSim:
     Stores contents in internal buffer and optionally echos to the inner stream it is simulating.
     """
 
-    def __init__(self, inner_stream, *, echo: bool = False, encoding: str = 'utf-8', errors: str = 'replace') -> None:
+    def __init__(self, inner_stream: TextIO, *, echo: bool = False, encoding: str = 'utf-8', errors: str = 'replace') -> None:
```
kmvanbrunt, line 450: inner_stream can also be a StdSim

cmd2/utils.py
```diff
@@ -509,11 +515,11 @@ def line_buffering(self) -> bool:
         when running unit tests because pytest sets stdout to a pytest EncodedFile object.
         """
         try:
-            return self.inner_stream.line_buffering
+            return bool(self.inner_stream.line_buffering)
```
kmvanbrunt, line 518: Isn't line_buffering already a bool?

cmd2/utils.py
```diff
@@ -1054,13 +1060,13 @@ def categorize(func: Union[Callable, Iterable[Callable]], category: str) -> None
         for item in func:
             setattr(item, constants.CMD_ATTR_HELP_CATEGORY, category)
     else:
-        if inspect.ismethod(func):
-            setattr(func.__func__, constants.CMD_ATTR_HELP_CATEGORY, category)
+        if inspect.ismethod(func) and hasattr(func, '__func__'):
```
kmvanbrunt, line 1063: Why this extra check?

kmvanbrunt (changes requested):

cmd2/utils.py
anselor: Apparently it's `int` on one of the versions of python we're using.

cmd2/ansi.py
```diff
@@ -229,7 +230,7 @@ def style_aware_write(fileobj: IO, msg: str) -> None:
     fileobj.write(msg)
 
 
-def fg_lookup(fg_name: Union[str, fg]) -> str:
+def fg_lookup(fg_name: Union[str, fg]) -> Fore:
```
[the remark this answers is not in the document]
anselor: We weren't returning a string. I can change it to return a string.

cmd2/argparse_custom.py
```diff
@@ -287,7 +292,11 @@ class ChoicesCallable:
     While argparse has the built-in choices attribute, it is limited to an iterable.
     """
 
-    def __init__(self, is_completer: bool, to_call: Callable):
+    def __init__(
+        self,
+        is_completer: bool,
+        to_call: Union[Callable[[], List[str]], Callable[[str, str, int, int], List[str]]],
```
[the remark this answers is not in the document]
anselor: ok, to support this I'm going to need to declare a callback Protocol I think.

cmd2/cmd2.py
anselor: Yes, because self.settables is now a property that aggregates settables in cmd2 and in each CommandSet.

cmd2/cmd2.py
anselor: The Settable class doesn't require the settable_obj to be set. For legacy purposes, anything without a defined object is assumed to be the the thing the settable is added to.

cmd2/command_definition.py
anselor: No. We can either add a check to prevent 2 commandsets from using the same prefix (easy) or do more complex checking.

cmd2/utils.py
anselor: All legacy settables have this set to None.

cmd2/utils.py
anselor: I can take that out. I was trying to get around a mypy error and ended up just telling it to ignore it.

cmd2/utils.py
kmvanbrunt: Nevermind, we should keep OrderedDict for non-CPython environments where dictionaries may not preserve insertion order.

cmd2/cmd2.py
anselor: It can - but it makes 'key' inaccessible to the exception being raised.

cmd2/cmd2.py
anselor: Thinking about this more, this is a lot less likely since I 'broke' the existing API. However I think this will still be necessary in case people don't use the `add_settable` function.

tleonhardt (approved):

cmd2/ansi.py
```diff
@@ -86,14 +87,14 @@ def __add__(self, other: Any) -> str:
         Support building a color string when self is the left operand
         e.g. fg.blue + "hello"
         """
-        return str(self) + other
+        return cast(str, str(self) + other)
```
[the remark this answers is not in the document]
anselor: Python cast is just a function that returns your value with a matching type decorator. No operation is performed.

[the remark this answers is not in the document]
tleonhardt: Interesting.  Well, I learned something today ;-)

tleonhardt (changes requested):
The CHANGELOG needs an update related to backwards incompatible changes present in this PR

cmd2/cmd2.py
```diff
@@ -3712,29 +3802,22 @@ def do_set(self, args: argparse.Namespace) -> None:
         if args.param:
             try:
                 settable = self.settables[args.param]
+                # For legacy reasons, in case developers don't switch to use the add_settable function
```
kmvanbrunt, line 3805: Is this comment necessary?

kmvanbrunt (changes requested):

cmd2/cmd2.py
anselor: The settables property returns all settables in Cmd as well as all CommandSets. `self._settables` is only what's Cmd.

cmd2/ansi.py
```diff
@@ -291,13 +292,13 @@ def style(
...
-    additions = []
+    additions: List[str] = []
 
     # List of strings that remove style
-    removals = []
+    removals: List[str] = []
 
     # Convert the text object into a string if it isn't already one
-    text = "{}".format(text)
+    text_formatted = "{}".format(text)
```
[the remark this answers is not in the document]
anselor: Yes, the variable can only have one type associated. Mypy doesn't like the type changing partway through.

cmd2/cmd2.py
kmvanbrunt: Sorry, I get it now.

cmd2/argparse_custom.py
```diff
@@ -281,13 +296,37 @@ def __init__(self, value: object, desc: str = '', *args) -> None:
 ############################################################################################################
 # Class and functions related to ChoicesCallable
 ############################################################################################################
+
+
+class ChoicesCallableFunc(Protocol):
+    def __call__(self, *, arg_tokens: Dict[str, List[str]]) -> List[str]:
```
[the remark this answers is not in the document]
anselor: We will likely need to refine this as we get further along in getting all of the annotations correct.

cmd2/cmd2.py
anselor: no

cmd2/argparse_custom.py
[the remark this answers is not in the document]
anselor: There are now 2 forms of each of these.

kmvanbrunt (approved):

PR #1076, opened 2021-03-19
Pass SystemExit code to self.exit_code

jedie:
The exit code will be missed, if it's not passed to `self.exit_code`


kmvanbrunt (approved):
I like this change. We'll want to add some unit tests which call `sys.exit()` or raise `SystemExit` in `do_*` functions. Also we should update our examples and docs about `self.exit_code`. While it is true that a developer can manually set `self.exit_code` and return `True` from their `do_*` function, calling `sys.exit()` needs to be supported.

kmvanbrunt:
@jedie Thanks for this PR. I'm gonna close this version of it and make a few tweaks. I'll create a new PR in a bit.

kmvanbrunt:
@jedie See #1078 for current PR.

PR #1080, opened 2021-03-24
ipy command now includes all of self.py_locals in the IPython environment

kmvanbrunt:


tleonhardt (dismissed):

cmd2/cmd2.py
```diff
@@ -168,7 +168,7 @@
 try:
     # noinspection PyUnresolvedReferences,PyPackageRequirements
     from IPython import (  # type: ignore[import]
-        embed,
+        start_ipython,
```
tleonhardt, line 171: Just curious, what is the difference between `embed` and `start_ipython`?

tleonhardt (dismissed):

cmd2/cmd2.py
kmvanbrunt: Since `embed()` runs within the scope of a function, it has access to both the `globals()` and `locals()` namespaces. Limiting what's in `locals()` is easy since it only includes variables in the function where `embed()` was called. However, I couldn't find a way to limit access to `globals()` using `embed()`.

Here is the difference between it and `start_ipython()`.
https://ipython.org/ipython-doc/stable/api/generated/IPython.html#IPython.start_ipython

Basically IPython is running in a sandbox this way and only has access to what we provide it.

anselor (approved):

PR #1086, opened 2021-03-29
Mypy explicit export

anselor:
Running mypy with strict checking gets a lot of errors when using cmd2 related to implicit exports.
General mypy related validation fixes added.

Took a stab at creating a type declaration for matching argparse command functions. Will likely need further refinement.

### DIFF
--- a/cmd2/argparse_custom.py
+++ b/cmd2/argparse_custom.py
@@ -872,7 +872,7 @@ class Cmd2ArgumentParser(argparse.ArgumentParser):
         usage: Optional[str] = None,
         description: Optional[str] = None,
         epilog: Optional[str] = None,
-        parents: Sequence[argparse.ArgumentParser] = [],
+        parents: Sequence[argparse.ArgumentParser] = (),
         formatter_class: Type[argparse.HelpFormatter] = Cmd2HelpFormatter,
         prefix_chars: str = '-',
         fromfile_prefix_chars: Optional[str] = None,
@@ -989,7 +989,7 @@ class Cmd2AttributeWrapper:
     arguments from a parser and which were added by cmd2.
     """
 
-    def __init__(self, attribute: Any):
+    def __init__(self, attribute: Any) -> None:
         self.__attribute = attribute
 
     def get(self) -> Any:
--- a/cmd2/argparse_custom.py
+++ b/cmd2/argparse_custom.py
@@ -215,7 +215,6 @@ from typing import (
     Tuple,
     Type,
     Union,
-    cast,
 )
 
 from . import (
@@ -279,7 +278,7 @@ class CompletionItem(str):
     """
 
     def __new__(cls, value: object, *args: Any, **kwargs: Any) -> 'CompletionItem':
-        return cast(CompletionItem, super(CompletionItem, cls).__new__(cls, value))  # type: ignore [call-arg]
+        return super(CompletionItem, cls).__new__(cls, value)
 
     # noinspection PyUnusedLocal
     def __init__(self, value: object, desc: str = '', *args: Any) -> None:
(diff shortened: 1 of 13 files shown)


cmd2/cmd2.py
```diff
@@ -1884,17 +1892,22 @@ def _perform_completion(
             return
 
         # Determine the completer function to use
-        if command is not None:
+        if custom_settings is None:
```
kmvanbrunt, line 1895: Why was this logic changed?

kmvanbrunt (changes requested):

cmd2/decorators.py
```diff
@@ -196,13 +200,30 @@ def _set_parser_prog(parser: argparse.ArgumentParser, prog: str):
...
+ArgparseCommandFuncBoolReturn = Union[
+    Callable[['cmd2.Cmd', argparse.Namespace], bool],
+    Callable[[CommandSet, argparse.Namespace], bool],
+]
+ArgparseCommandFuncNoneReturn = Union[
+    Callable[['cmd2.Cmd', argparse.Namespace], None],
+    Callable[[CommandSet, argparse.Namespace], None],
+]
+
+
```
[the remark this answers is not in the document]
anselor: Yes. This file needs a lot more work and attention. I was hitting a few items that the other files I touched needed.

cmd2/py_bridge.py
```diff
@@ -58,7 +66,7 @@ class CommandResult(NamedTuple):
     """
 
     stdout: str = ''
-    stderr: str = ''
+    stderr: Optional[str] = None
```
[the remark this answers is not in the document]
anselor: I don't have strong opinions on this. I was just making the type match how we were using it.

cmd2/cmd2.py
kmvanbrunt: The only way for `command` to not be `None` is if `custom_settings` is `None`. The new logic is redundant.

cmd2/argparse_custom.py
```diff
@@ -872,7 +872,7 @@ def __init__(
         usage: Optional[str] = None,
         description: Optional[str] = None,
         epilog: Optional[str] = None,
-        parents: Sequence[argparse.ArgumentParser] = [],
+        parents: Sequence[argparse.ArgumentParser] = (),
```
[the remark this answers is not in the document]
anselor: Using a mutable object in an initializer is generally frowned upon because that instance is shared for every call to the function. The validators complain.

cmd2/cmd2.py
anselor: As with all changes here, to make the appease the type validator.

cmd2/decorators.py
```diff
@@ -145,9 +149,9 @@ def cmd_wrapper(*args, **kwargs: Dict[str, Any]) -> Optional[bool]:
         cmd_wrapper.__doc__ = func.__doc__
         return cmd_wrapper
 
-    if len(args) == 1 and callable(args[0]):
+    if callable(func_arg):
```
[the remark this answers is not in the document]
anselor: Yes, the function no longer accepts a variable number of positionals and instead optionally takes 1 positional.

cmd2/cmd2.py
kmvanbrunt: I'd like to refactor this code or something. The new solution ends up with a redundant block that never gets unit tested since it can't actually occur.

```Python
if command is None:
    return
```

kmvanbrunt: What was the type validator error on this code?

anselor: I put a cast where it was complaining instead.

PR #1088, opened 2021-04-03
Shared implementation of write it StdSim

anselor:


cmd2/utils.py
```diff
@@ -477,11 +477,7 @@ def write(self, s: str) -> None:
         """
         if not isinstance(s, str):
             raise TypeError('write() argument must be str, not {}'.format(type(s)))
-
-        if not self.pause_storage:
-            self.buffer.byte_buf += s.encode(encoding=self.encoding, errors=self.errors)
-        if self.echo:
-            self.inner_stream.write(s)
+        self.buffer.write(s.encode(self.encoding))
```
tleonhardt, line 480: Does dropping `errors=self.errors` have a significant effect?  I would worry about edge cases.

tleonhardt (approved):

cmd2/utils.py
anselor: Good catch. Added that back.

cmd2/utils.py
```diff
@@ -477,11 +477,7 @@ def write(self, s: str) -> None:
         """
         if not isinstance(s, str):
             raise TypeError('write() argument must be str, not {}'.format(type(s)))
-
-        if not self.pause_storage:
-            self.buffer.byte_buf += s.encode(encoding=self.encoding, errors=self.errors)
-        if self.echo:
-            self.inner_stream.write(s)
+        self.buffer.write(s.encode(self.encoding, errors=self.errors))
```
kmvanbrunt, line 480: This approach replaces `inner_stream.write()` with `inner_stream.buffer.write()`. Does writing directly to the buffer remove support for newline translation and unbuffered writes?

```Python
class io.TextIOWrapper(buffer, encoding=None, errors=None, newline=None,
                       line_buffering=False, write_through=False)
```

anselor: I don't know how to even check that. Can you think of a way a unit test could verify this?

tleonhardt:
@anselor What is the status of this PR?  Should we close it?

Is there more work that needs to be done to fix stuff before we can merge it in?

tleonhardt:
@anselor I'm going to close this PR for now since it has been left in a broken state for over a month.  I won't delete the branch immediately in case you want to continue working on it and resubmit, but will eventually delete it.

PR #1098, opened 2021-04-27
Added handling for OSError when loading GNU readline shared library

### LINKED ISSUE #1096: rl_utils trying to load .py as a shared library on PyPy3
When trying to run cmd2's tests on PyPy3, I get the following error:

```
pypy3 -m pytest -vv -ra -l
ImportError while loading conftest '/tmp/portage/dev-python/cmd2-1.5.0/work/cmd2-1.5.0/tests/conftest.py'.
tests/conftest.py:23: in <module>
    import cmd2
cmd2/__init__.py:30: in <module>
    from .cmd2 import Cmd
cmd2/cmd2.py:110: in <module>
    from .rl_utils import (
cmd2/rl_utils.py:124: in <module>
    readline_lib = ctypes.CDLL(readline.__file__)
/usr/lib/pypy3.7/lib-python/3/ctypes/__init__.py:366: in __init__
    pypy_dll = _ffi.CDLL(name, mode)
E   OSError: Cannot load library /usr/lib/pypy3.7/lib_pypy/readline.py: /usr/lib/pypy3.7/lib_pypy/readline.py: invalid ELF header
```

Apparently the cause is cmd2/rl_utils.py#L131:

```
readline_lib = ctypes.CDLL(readline.__file__)
```

which assumes that `readline` will be a shared library rather than pure Python module.

kmvanbrunt:
Fixes #1096 

PyPy has a pure Python implementation of readline and therefore does not have a shared library to load.

### DIFF
--- a/cmd2/rl_utils.py
+++ b/cmd2/rl_utils.py
@@ -129,10 +129,10 @@ elif 'gnureadline' in sys.modules or 'readline' in sys.modules:
             import ctypes
 
             readline_lib = ctypes.CDLL(readline.__file__)
-        except AttributeError:  # pragma: no cover
+        except (AttributeError, OSError):  # pragma: no cover
             _rl_warn_reason = (
                 "this application is running in a non-standard Python environment in\n"
-                "which readline is not loaded dynamically from a shared library file."
+                "which GNU readline is not loaded dynamically from a shared library file."
             )
         else:
             rl_type = RlType.GNU


cmd2/rl_utils.py
```diff
@@ -129,10 +129,10 @@ def pyreadline_remove_history_item(pos: int) -> None:
             import ctypes
 
             readline_lib = ctypes.CDLL(readline.__file__)
-        except AttributeError:  # pragma: no cover
+        except (AttributeError, OSError):  # pragma: no cover
```
tleonhardt, line 132: We shoudl consider creating a new issue to add PyPy unit testing as part of our normal CI via GitHub Actions.  However, PyPy is unique in that code coverage analysis in it is awfully slow, so we would need to run unit tests differently than all other platforms where just the tests run and not code coverage.

tleonhardt (approved):

cmd2/rl_utils.py
kmvanbrunt: The biggest hurdle is getting cmd2 to support their version of readline.

PR #1105, opened 2021-05-14
Make _run_editor() public

srowe:
It is useful for commands to be able to do the following

    with NamedTemporaryFile() as fileh:
        # convert data to text
        self._run_editor(fileh.name)
        # convert text back to data


cmd2/cmd2.py
```diff
@@ -4389,7 +4389,7 @@ def do_history(self, args: argparse.Namespace) -> Optional[bool]:
                     else:
                         fobj.write(f'{command.raw}\n')
             try:
-                self._run_editor(fname)
+                self.run_editor(fname)
```
tleonhardt, line 4392: This seems like a reasonable change.  Once it is merged in, we should update the CHANGELOG and search through the Sphinx docs to see if there is anything we need to update there as well.

tleonhardt (approved):

tleonhardt:
Thanks for the PR!

PR #1130, opened 2021-07-02
Added accessor methods for cmd2-specific attributes to the argparse.Action class.

kmvanbrunt:
* Added accessor methods for cmd2-specific attributes to the `argparse.Action` class.
* Deprecated `set_choices_provider()` and `set_completer()` functions in favor of these new methods.


tleonhardt (approved):

anselor (dismissed):
I like the new method-based access. Thanks for the change.

PR #1138, opened 2021-08-25
Refactored custom argparse completers

kmvanbrunt:
Refactored custom ArgparseCompleter functionality so they will now be set using methods on ArgumentParser objects.
This fixes issue where subcommands did not use the correct custom ArgparseCompleter type.

### DIFF
--- a/cmd2/argparse_completer.py
+++ b/cmd2/argparse_completer.py
@@ -30,7 +30,7 @@ from .constants import (
     INFINITY,
 )
 
-if TYPE_CHECKING:
+if TYPE_CHECKING:  # pragma: no cover
     from .cmd2 import (
         Cmd,
     )
--- a/cmd2/argparse_completer.py
+++ b/cmd2/argparse_completer.py
@@ -407,9 +407,15 @@ class ArgparseCompleter:
                                 if action.dest != argparse.SUPPRESS:
                                     parent_tokens[action.dest] = [token]
 
-                                completer = ArgparseCompleter(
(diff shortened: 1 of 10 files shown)


tleonhardt (dismissed):

cmd2/argparse_completer.py
```diff
@@ -407,9 +407,15 @@ def update_mutex_groups(arg_action: argparse.Action) -> None:
                                 if action.dest != argparse.SUPPRESS:
                                     parent_tokens[action.dest] = [token]
 
-                                completer = ArgparseCompleter(
-                                    self._subcommand_action.choices[token], self._cmd2_app, parent_tokens=parent_tokens
-                                )
+                                parser: argparse.ArgumentParser = self._subcommand_action.choices[token]
+                                completer_type: Optional[
+                                    Type[ArgparseCompleter]
+                                ] = parser.get_ap_completer_type()  # type: ignore[attr-defined]
```
anselor, line 413: Since we have to do these few lines everywhere we use this, would it make more sense to change `get_ap_completer_type()` to return `DEFAULT_AP_COMPLETER` if nothing has been set?

anselor (dismissed):
Just the one note. Would be a minor code reduction but not super important to be done that way.

cmd2/argparse_completer.py
kmvanbrunt: I made that code common in a cmd2 method. I prefer letting `get_ap_completer_type()` still return `None` and let the caller determine how to handle it.

anselor (approved):

PR #1187, opened 2022-01-01
Added substantial documentation concerning pyscript

jayrod:
Based on the changes I've made I didn't think it made sense to keep the blurb concerning cmd2 plugin ext testing. But hopefully I added enough examples to get people going in the right direction.

### DIFF
--- a/docs/features/scripting.rst
+++ b/docs/features/scripting.rst
@@ -84,6 +84,28 @@ command. Here's a simple example that uses the arg_printer_ script::
 system paths, and as shown above it has the ability to pass command-line
 arguments to the scripts invoked.
 
+Developing a CMD2 API 
+---------------------
+
+If you as an app designer have not explicitly disabled the run_pyscript command it must be assumed
+that your application is structured for use in higher level python scripting. The following sections
+are meant as guidelines and highlight possible pitfalls with both production and consumption
+of API functionality. For clarity when speaking of "scripter" we are referring to those writing
+scripts to be run by pyscript and "designer" as th CMD2 application author. 
+
+Basics
+~~~~~~
+
+When implementing a command, setting ``self.last_result`` allows for application-specific
+data to be returned to a python script from the command. This can allow python scripts to
+make decisions based on the result of previous application commands.
+
+The application command (default: ``app``) returns a ``cmd2.CommandResult`` for each command.
+The ``cmd2.CommandResult`` object provides the captured output to ``stdout`` and ``stderr``
+while a command is executing. Additionally, it provides the value that command stored in
+``self.last_result``.
+
+
 Python scripts executed with :ref:`features/builtin_commands:run_pyscript` can
 run ``cmd2`` application commands by using the syntax::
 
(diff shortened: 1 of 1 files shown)


docs/features/scripting.rst
```diff
@@ -96,29 +118,405 @@ where:
...
+  (Cmd) run_pyscript script.py 
+  Working
+  (Cmd) 
+
+You should notice that no error message is printed. Let's utilize the ``CommandResult`` 
+object to inspect the actual returned data.::
+
+  result = app('speak')
+  print(result)
+::
```
tleonhardt, line 275: Need to fix doc build warning:
```
Warning, treated as error:
/home/runner/work/cmd2/cmd2/docs/features/scripting.rst:275:Literal block ends without a blank line; unexpected unindent.
```

tleonhardt (changes requested):
Need to fix the strict doc build test

docs/features/scripting.rst
```diff
@@ -84,6 +84,28 @@ command. Here's a simple example that uses the arg_printer_ script::
...
 arguments to the scripts invoked.
 
+Developing a CMD2 API 
+---------------------
+
+If you as an app designer have not explicitly disabled the run_pyscript command it must be assumed
+that your application is structured for use in higher level python scripting. The following sections
+are meant as guidelines and highlight possible pitfalls with both production and consumption
+of API functionality. For clarity when speaking of "scripter" we are referring to those writing
+scripts to be run by pyscript and "designer" as th CMD2 application author.
```
kmvanbrunt, line 94: Fix "as th".

docs/features/scripting.rst
```diff
@@ -84,6 +84,28 @@ command. Here's a simple example that uses the arg_printer_ script::
...
+~~~~~~
+
+When implementing a command, setting ``self.last_result`` allows for application-specific
+data to be returned to a python script from the command. This can allow python scripts to
+make decisions based on the result of previous application commands.
+
+The application command (default: ``app``) returns a ``cmd2.CommandResult`` for each command.
+The ``cmd2.CommandResult`` object provides the captured output to ``stdout`` and ``stderr``
+while a command is executing. Additionally, it provides the value that command stored in
+``self.last_result``.
```
kmvanbrunt, line 106: Specify that the `CommandResult` object stores `self.last_result` in its `data` member.
Also mention `CommandResult.stop` which is the return value of `onecmd_plus_hooks()` after it runs the given command line. It says whether the command loop is intended to stop. (e.g. quit returns `True`)

docs/features/scripting.rst
```diff
@@ -96,29 +118,405 @@ where:
...
-Advanced Support
-~~~~~~~~~~~~~~~~
 
-When implementing a command, setting ``self.last_result`` allows for application-specific
-data to be returned to a python script from the command. This can allow python scripts to
-make decisions based on the result of previous application commands.
+Design principles
+~~~~~~~~~~~~~~~~~
+If the cmd2 application follows the unix_design_philosophy_ a scriptor will have the most flexibility
+to piece together workflows using different commands. If the designers application is more complete
```
kmvanbrunt, line 137: Make "designers" possessive.

docs/features/scripting.rst
```diff
@@ -96,29 +118,405 @@ where:
...
+::
+
+  (Cmd) run_pyscript script.py 
+  CommandResult(stdout='', stderr='Usage: speak [-h] [-p] [-s] [-r REPEAT] words [...]\nError: the following arguments are required: words\n\n', stop=False, data=None)
+
+Now we can see that there has been an error. Let's re write the script to perform error checking.::
+
+  result = app('speak')
+
+  if result.stderr:
```
kmvanbrunt, line 284: `CommandResult` implements `__bool__()`.

```Python
def __bool__(self) -> bool:
    """Returns True if the command succeeded, otherwise False"""

    # If data was set, then use it to determine success
    if self.data is not None:
        return bool(self.data)

    # Otherwise check if stderr was filled out
    else:
        return not self.stderr
````

Therefore, you don't need to check for errors this way: `if result.stderr:`
Instead just do: `if not result:`

File: docs/features/scripting.rst [hunk 6]
```diff
@@ -96,29 +118,405 @@ where:
...
+    dir_parser.add_argument('-l', '--long', action='store_true', help="display in long format with one item per line")
+
+    @cmd2.with_argparser(dir_parser, with_unknown_args=True)
+    def do_dir(self, args, unknown):
+        """List contents of current directory."""
+        # No arguments for this command
+        if unknown:
+            self.perror("dir does not take any positional arguments:")
+            self.do_help('dir')
+            self.last_result = 'Bad arguments'
```
kmvanbrunt, line 361: I'd probably set `last_result` to `None` since the presence of the error in `CommandResult.stderror` already supports using the `CommandResult` object as a bool to check for errors.

Since `CommandResult.__bool__()` gives preference to a non-None `data` member, it will treat "Bad arguments" as a boolean True value.

kmvanbrunt (changes requested):
Thanks for working on this. In regards to my comment about `CommandResult.__bool__()`, try to minimize or eliminate using `if result.stderr:` to check for errors and instead use `if not result:` if you can.

tleonhardt:
@jayrod Thanks for working to help improve the documentation! We would like to get this merged in whenever you can find the time to address the comments.  Thanks agian!

jayrod:
Oh wow.. I have not received these messages. My apologies I'll work on this tonight.

docs/features/scripting.rst
jayrod: God, that is soo helpful. Wish I had known that several months ago. I'm going to need to refactor a large chunk of my codebase ;)

docs/features/scripting.rst
jayrod: Is last_result set to None automatically?  The way I've been  writing commands is to set last_result iff there is something positive to return.

It makes it easier for the designer.

docs/features/scripting.rst
jayrod: is there an rst linter?? I was using vscode RST extension but didn't see anything wrong.

docs/features/scripting.rst
jayrod: CommandREsult.stop seems extra? how is that different than stderr or CommandResult.__bool__? I know that they are different but I'm not groking how I as a scriptor would use one over another?

jayrod: Without any work on the part of the designer, a scripter can take advantage of piecing together workflows
using simple ``app`` calls. The result of a run_pyscript app call yields a ``CommandResult`` object exposing 
four members: ``Stdout``, ``Stderr``, ``Stop``, and ``Data``. 

``Stdout`` and ``Stderr`` are fairly straightforward representations of normal data streams and accurately reflect
what is seen by the user during normal cmd2 interaction. ``Stop`` contains information about how the invoked 
command has ended its lifecycle. Lastly ``Data`` contains any information the designer sets via ``self.last_result`` 
or ``self._cmd.last_result`` if called from inside a CommandSet.

tleonhardt:
@jayrod I have invited you to be a `cmd2` contributor.  This will make it much easier for you to contribute in the future and to collaborate with other developers.  It will allow you to create branches in the main repo and create PRs directly off of those branches - so no need to use a fork from here on out.

tleonhardt (approved):

jayrod:
> @jayrod I have invited you to be a `cmd2` contributor. This will make it much easier for you to contribute in the future and to collaborate with other developers. It will allow you to create branches in the main repo and create PRs directly off of those branches - so no need to use a fork from here on out.

I appreciate the invite and I'll try to continue my contributions :)

PR #1206, opened 2022-02-22
Improving table memory usage

kmvanbrunt:
Reduced amount of style characters carried over from previous lines when aligning text.
Also reduced amount of style characters appended to truncated text.
These changes were made to reduce memory usage in certain use cases of tables (e.g. nested colored tables).

### DIFF
--- a/tests/test_ansi.py
+++ b/tests/test_ansi.py
@@ -253,3 +253,47 @@ def test_rgb_bounds(r, g, b, valid):
             ansi.RgbFg(r, g, b)
         with pytest.raises(ValueError):
             ansi.RgbBg(r, g, b)
+
+
+def test_std_color_re():
+    """Test regular expressions for matching standard foreground and background colors"""
+    for color in ansi.Fg:
+        assert ansi.STD_FG_RE.match(str(color))
+        assert not ansi.STD_BG_RE.match(str(color))
+    for color in ansi.Bg:
+        assert ansi.STD_BG_RE.match(str(color))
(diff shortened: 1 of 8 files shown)


tests/test_ansi.py
```diff
@@ -253,3 +253,41 @@ def test_rgb_bounds(r, g, b, valid):
...
         with pytest.raises(ValueError):
             ansi.RgbBg(r, g, b)
+
+
+def test_std_color_re():
+    """Test regular expressions for matching standard foreground and background colors"""
+    for color in ansi.Fg:
+        assert ansi.STD_FG_RE.match(str(color))
+    for color in ansi.Bg:
+        assert ansi.STD_BG_RE.match(str(color))
```
anselor, line 263: Would it also make sense to test the reverse for each color?
Like:
```
    for color in ansi.Bg:
        assert ansi.STD_BG_RE.match(str(color))
        assert not ansi.STD_FG_RE.match(str(color))
```

anselor (approved):

tests/test_utils.py
```diff
@@ -341,6 +344,79 @@ def test_context_flag_exit_err(context_flag):
         context_flag.__exit__()
 
 
+def test__remove_overridden_styles():
```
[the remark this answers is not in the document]
kmvanbrunt: Renamed test function

tests/test_ansi.py
kmvanbrunt: Done for all 3 functions like this

anselor (approved):

Issue #1185, opened 2021-12-22
transcripts do not catch errors


duck-rh:
Quack,

I'm using cmd2 2.3.3 on Python 3.9.9 on Debian unstable.

I am trying to test my application using transcripts. Depending on the gravity of the situation it uses `pwarning()` or `perror()` to alert the user but these calls are not caught by transcript and looking at the code of `_generate_transcript` is seems only stdout is captured.

I believe testing application checks are useful and in most test framework you can catch exceptions; would you consider capturing stderr too in the transcript?

Regards.
\\_o<

crotwell:
Possible this could be resolved by PR #1208
Also see issue #1207

kmvanbrunt:
@tleonhardt @anselor 
Can you think of a reason why capturing **stderr** would be problematic?

anselor:
@kmvanbrunt The transcripts feature is an area of cmd2 I haven't really touched. At a high level I see no reason why it would be a problem.

crotwell:
@kmvanbrunt I would be happy to see a more robust solution, but currently cmd2 transcripts do not actually capture stdout, they only capture the output of `self.poutput()`.

In other words, if you have two commands like:
```
    def do_print(self, statement):
        """print something"""
        print("printing something")

    def do_poutput(self, statement):
        """poutput something"""
        self.poutput("poutput something")
```

and run you see this:
```
(Cmd) print
printing something
(Cmd) poutput
poutput something
```
but if you create a transcript with:
```
(Cmd) history -t transcript.txt -- -2:
```
then the transcript.txt looks like:
```
(Cmd) print
(Cmd) poutput
poutput something
```
so the print output is missing.

If you can figure out how to actually capture stdout, that would be great. But if not, maybe capturing `self.perror()` as in PR #1208 is at least sufficient so transcripts fail if there are exceptions raised?

crotwell:
OK, I see the clarification. You said capturing `stderr`, but sounds like you actually meant capturing `self.stderr`.

I do not care about stderr being in the transcript, intermingling stdout and stderr is probably hard and error prone. And I guess there could be cases where self.stderr was not an actual failure, so `perror` is not the right way to go. As I said in issue #1207, what I actually care about is that a transcript test should not "pass" if there is a failure situation in cmd2. In particular, a test of a transcript that contains a command that does not actually exist should not report "ok". If it was easier or less invasive to fail on exceptions or other error handling instead of stderr, that is probably better.

[...]

PR #1210, opened 2022-03-01
use pybridge to capture stderr for transcript tests

crotwell:
use pybridge to capture stderr when running transcripts as tests. Think this addresses at least part of #1185, as stderr is part of the comparison. But this does not address outputing stderr when creating a transcript using history -t.

Possible that simple concatenation is not the right way to create the result for comparison, but at least any self.stderr output will be noticed by the test.


tleonhardt:
@crotwell Formatting is still broken due to isort.

I added a new invoke task you can run with `inv format` which will auto-format code using both black and isort.

tleonhardt:
@kmvanbrunt @anselor Now that all tests are passing, could one or both of you please review this PR?

IMO I don't like that it is only a partial solution.  On the other hand, perhaps that is better than no solution?

tleonhardt:
@crotwell Thank you very much for this PR.  We appreciate your contribution.

My understanding is that this solution is incomplete and likely not the best long-term approach. This PR has existed for almost a year and I am going to close it now.

If someone would like to work on a more complete and appropriate solution in the future, that would be great.

PR #1224, opened 2022-04-18
Moved documentation annotation stripping into utility function.

### LINKED ISSUE #1220: command function type annotations are pulled into command help
in `with_argparser`, when an explicit description is not provided we pull the functions `__doc__` property as the description. It currently pulls the entire __doc__ property including any documentation generation annotations. Would be cleaner to omit annotations.

Example:

```
@with_argparser(foo_args)
def do_foo(args: Namespace) -> None
"""
Does foo stuff

:param args: doc-annotation for args
:raises MyException: some notes about exceptions
"""
```

Results in something like:
```
> help foo
Usage: foo

    Does foo stuff

    :param args: doc-annotation for args
    :raises MyException: some notes about exceptions
```

And would be cleaner like:
```
> help foo
Usage: foo

    Does foo stuff
```

anselor:
Changed argparse decorator to automatically strip annotations. 

Resolves #1220

### DIFF
--- a/cmd2/utils.py
+++ b/cmd2/utils.py
@@ -1228,3 +1228,29 @@ class CustomCompletionSettings:
         """
         self.parser = parser
         self.preserve_quotes = preserve_quotes
+
+
+def strip_doc_annotations(doc: str) -> str:
+    """
+    Strip annotations from a docstring leaving only the text description
+
+    :param doc: documentation string
+    """
+    # Attempt to locate the first documentation block
+    cmd_desc = ''
+    found_first = False
+    for doc_line in doc.splitlines():
+        stripped_line = doc_line.strip()
+
+        # Don't include :param type lines
+        if stripped_line.startswith(':'):
+            if found_first:
+                break
+        elif stripped_line:
+            if found_first:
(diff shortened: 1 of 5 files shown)


cmd2/utils.py
```diff
@@ -1228,3 +1228,29 @@ def __init__(self, parser: argparse.ArgumentParser, *, preserve_quotes: bool = F
...
+    for doc_line in doc.splitlines():
+        stripped_line = doc_line.strip()
+
+        # Don't include :param type lines
+        if stripped_line.startswith(':'):
+            if found_first:
+                break
+        elif stripped_line:
+            if found_first:
+                cmd_desc += "\n"
```
kmvanbrunt, line 1251: Can you add a unit test for this line?

kmvanbrunt (changes requested):
Looks good, just add unit test for 100% coverage of `strip_doc_annotations()`.

cmd2/utils.py
kmvanbrunt: Thanks

kmvanbrunt (approved):

PR #1239, opened 2022-07-14
Use latest version of mypy and fix type hinting accordingly

tleonhardt:
Use latest version of mypy and fix type hinting accordingly

Also:
- Update Pipfile to never require mock since we only support Python 3.6+
- Remove Azure Pipelines badge from Readme and fix section links there
- Added an "inv format" task to run black and isort to auto-format all code before a commit

### DIFF
--- a/cmd2/argparse_custom.py
+++ b/cmd2/argparse_custom.py
@@ -1033,7 +1033,7 @@ setattr(argparse.ArgumentParser, '_check_value', _ArgumentParser_check_value)
 ############################################################################################################
 
 # noinspection PyPep8Naming,PyProtectedMember
-def _SubParsersAction_remove_parser(self: argparse._SubParsersAction, name: str) -> None:
+def _SubParsersAction_remove_parser(self: Any, name: str) -> None:
     """
     Removes a sub-parser from a sub-parsers group. Used to remove subcommands from a parser.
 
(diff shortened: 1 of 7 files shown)


cmd2/argparse_custom.py
```diff
@@ -1033,7 +1033,7 @@ def _ArgumentParser_check_value(self: argparse.ArgumentParser, action: argparse.
 ############################################################################################################
 
 # noinspection PyPep8Naming,PyProtectedMember
-def _SubParsersAction_remove_parser(self: argparse._SubParsersAction, name: str) -> None:
+def _SubParsersAction_remove_parser(self: Any, name: str) -> None:
```
kmvanbrunt, line 1036: `self` has to be an `argparse._SubParsersAction` since this is a custom method that we add to the `_SubParsersAction` class.

cmd2/argparse_custom.py
```diff
@@ -1333,7 +1333,7 @@ def __init__(
         self.set_ap_completer_type(ap_completer_type)  # type: ignore[attr-defined]
 
     # noinspection PyProtectedMember
-    def add_subparsers(self, **kwargs: Any) -> argparse._SubParsersAction:
+    def add_subparsers(self, **kwargs: Any) -> Any:
```
kmvanbrunt, line 1336: `argparse._SubParsersAction` is the correct return type.

kmvanbrunt (changes requested):
What version of mypy is giving you errors for the argparse_custom.py code?
The previous types are correct. Maybe just add a `# type ignore` instead since mypy is giving you issues.

cmd2/argparse_custom.py
tleonhardt: Yeah I'll just add `# type: ignore` instead.  Thanks

cmd2/argparse_custom.py
tleonhardt: Fixed

PR #1257, opened 2023-01-28
[WIP] Start deprecation of Python 3.6

tleonhardt:
This is a WIP PR intended to remove support for Python 3.6 and make sure support for Python 3.11 is solid.

Remaining work to be done includes:
- Replace use of `attrs` with data classes and remove dependency on `attrs`
- Revisit type hinting to make sure we are following best practices
- Make sure all CI stuff is passing on all OSes and versions supported

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -46,8 +46,8 @@ The tables below list all prerequisites along with the minimum required version 
 #### Prerequisites to run cmd2 applications
 
 | Prerequisite                                        | Minimum Version |
-| --------------------------------------------------- | --------------- |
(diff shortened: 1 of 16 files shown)


tleonhardt:
@kmvanbrunt This PR is now ready for review whenever you have time.

We should hold off a little bit on merging because I think @anselor has a minor bug fix he would like to get in before we deprecate support for Python 3.6.

kotfu (approved):
This looks pretty good to me. I expect over time we may find a few more things we can clean up now that we don't support 3.6, but this gets us formally started. I think it's ready to merge.

tleonhardt:
@anselor Did you have some bug fix you wanted to get in before this PR merges? If so please get it done within the next day.

PR #1264, opened 2023-04-07
Fix Nonetype error when building with PyInstaller and no-console flag.

Starwarsfan2099:
If building a PyInstaller application with the `--noconsole` option and if any library imports `cmd2`, the application will fail with the following traceback:

```
Traceback (most recent call last):
  File "test.py", line 2, in <module>
    import cmd2
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "PyInstaller\loader\pyimod02_importers.py", line 352, in exec_module
  File "cmd2\__init__.py", line 53, in <module>
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "PyInstaller\loader\pyimod02_importers.py", line 352, in exec_module
  File "cmd2\cmd2.py", line 125, in <module>
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "PyInstaller\loader\pyimod02_importers.py", line 352, in exec_module
  File "cmd2\rl_utils.py", line 73, in <module>
AttributeError: 'NoneType' object has no attribute 'isatty'
```
The error can be replicated using the following code and building with `pyinstaller -F --distpath . --noconsole test.py`:
```
# test.py
import cmd2
from tkinter import *

root = Tk()
a = Label(root, text ="Hello World")
a.pack()
root.mainloop()
```
This was tested and reproduced using Windows 11, `Python 3.10.4`, and `cmd2 2.4.3`. 

This simple fix checks if `sys.stdout` is `None` before checking for `sys.stdout.isatty()`.

### DIFF
--- a/cmd2/rl_utils.py
+++ b/cmd2/rl_utils.py
@@ -70,42 +70,43 @@ if 'pyreadline3' in sys.modules:
     )
 
     # Check if we are running in a terminal
-    if sys.stdout.isatty():  # pragma: no cover
-        # noinspection PyPep8Naming,PyUnresolvedReferences
-        def enable_win_vt100(handle: HANDLE) -> bool:
-            """
-            Enables VT100 character sequences in a Windows console
-            This only works on Windows 10 and up
-            :param handle: the handle on which to enable vt100
-            :return: True if vt100 characters are enabled for the handle
-            """
-            ENABLE_VIRTUAL_TERMINAL_PROCESSING = 0x0004
+    if sys.stdout:
+        if sys.stdout.isatty():  # pragma: no cover
+            # noinspection PyPep8Naming,PyUnresolvedReferences
+            def enable_win_vt100(handle: HANDLE) -> bool:
+                """
+                Enables VT100 character sequences in a Windows console
+                This only works on Windows 10 and up
+                :param handle: the handle on which to enable vt100
+                :return: True if vt100 characters are enabled for the handle
+                """
+                ENABLE_VIRTUAL_TERMINAL_PROCESSING = 0x0004
 
-            # Get the current mode for this handle in the console
-            cur_mode = DWORD(0)
-            readline.rl.console.GetConsoleMode(handle, byref(cur_mode))
+                # Get the current mode for this handle in the console
+                cur_mode = DWORD(0)
+                readline.rl.console.GetConsoleMode(handle, byref(cur_mode))
 
-            retVal = False
+                retVal = False
 
-            # Check if ENABLE_VIRTUAL_TERMINAL_PROCESSING is already enabled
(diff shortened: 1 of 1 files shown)


tleonhardt (changes requested):

kmvanbrunt:
python.org says this about `sys.stdout` being None.

> Note Under some conditions stdin, stdout and stderr as well as the original values __stdin__, __stdout__ and __stderr__ can be None. It is usually the case for Windows GUI apps that aren’t connected to a console and Python apps started with pythonw.

What is the use-case for this fix? cmd2 is a CLI library and uses `sys.stdout` extensively. Why is it being imported into an application with no `stdout`? I don't want future developers to have to worry about checking `sys.stdout` in all code which runs during imports.

Starwarsfan2099:
@tleonhardt Thank you, that is a much cleaner way of checking. I've changed the check in c576bb6.

Starwarsfan2099:
@kmvanbrunt This originated from the [Pymobiledevice3](https://github.com/doronz88/pymobiledevice3) library. I use it in a GUI application to connect to phones via AFC. The actual AFC service library I need imports `cmd2` and uses it to create an AFC shell. My application has no need for that, and I think that CLI code should be initialized with other CLI code anyway. So I submitted a [pull request there](https://github.com/doronz88/pymobiledevice3/pull/435) that fixed this by moving the AFC shell creation code to a dedicated CLI import. However, they said this was a `cmd2` problem and directed me to open a PR here.

cmd2/rl_utils.py
```diff
@@ -70,7 +70,7 @@ class RlType(Enum):
     )
 
     # Check if we are running in a terminal
-    if sys.stdout.isatty():  # pragma: no cover
+    if sys.stdout and sys.stdout.isatty():  # pragma: no cover
```
kmvanbrunt, line 73: I would prefer this to be `if sys.stdout is not None ...` just to be clearer about what you're checking.

kmvanbrunt (changes requested):

kmvanbrunt (approved):

kmvanbrunt (approved):

kmvanbrunt (approved):

PR #1267, opened 2023-04-17
rl_utils: Verify `readline.__doc__ is not None`

netanelc305:


netanelc305:
should fix the following error i encountered.

https://github.com/doronz88/pymobiledevice3/actions/runs/4718117991/jobs/8367505899

cmd2/rl_utils.py
```diff
@@ -141,7 +141,7 @@ def pyreadline_remove_history_item(pos: int) -> None:
 
 elif 'gnureadline' in sys.modules or 'readline' in sys.modules:
     # We don't support libedit. See top of this file for why.
-    if 'libedit' not in readline.__doc__:
+    if readline.__doc__ and 'libedit' not in readline.__doc__:
```
kmvanbrunt, line 144: I would prefer this to be `readline.__doc__ is not None` just to be clearer about what you're checking.

kmvanbrunt (changes requested):

cmd2/rl_utils.py
netanelc305: done

kmvanbrunt (approved):

PR #1269, opened 2023-04-26
Updated poutput to use new style_output definition. 

### LINKED ISSUE #1268: poutput() and ansi styles.
in ansi.py we have: `style_success`, `style_warning`, and `style_error` defined. It looks like `pwarning()` and `perror()` use these style but `poutput()` doesn't.

I think one of these two changes should be mode:

* Option 1: Add a new `psuccess()` that uses `style_success` and define a new `style_output` that `poutput()` uses.
* Option 2: Change `poutput()` to use `style_success`

anselor:
Closes #1268

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -54,6 +54,7 @@ from types import (
     ModuleType,
 )
 from typing import (
+    IO,
     Any,
     Callable,
     Dict,
@@ -601,11 +602,14 @@ class Cmd(cmd.Cmd):
                     raise CommandSetRegistrationError(f'Duplicate settable {key} is already registered')
 
         cmdset.on_register(self)
-        methods = inspect.getmembers(
-            cmdset,
-            predicate=lambda meth: isinstance(meth, Callable)  # type: ignore[arg-type]
-            and hasattr(meth, '__name__')
-            and meth.__name__.startswith(COMMAND_FUNC_PREFIX),
+        methods = cast(
+            List[Tuple[str, Callable[..., Any]]],
+            inspect.getmembers(
+                cmdset,
+                predicate=lambda meth: isinstance(meth, Callable)  # type: ignore[arg-type]
+                and hasattr(meth, '__name__')
+                and meth.__name__.startswith(COMMAND_FUNC_PREFIX),
+            ),
         )
 
         default_category = getattr(cmdset, CLASS_ATTR_DEFAULT_HELP_CATEGORY, None)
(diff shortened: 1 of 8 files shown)


kmvanbrunt:
I like the idea of adding a `psuccess()` method. However, I have reservations about applying style in `poutput()` for the following reasons.

1. The docstring for `ansi.style_output` says it "colors text for normal output." That implies all output will have this style. However, async alerts, tab completions, prompts, and argparse help text aren't printed with `poutput()`. A developer would need to edit other functions to acheive a matching style.

2. Adding `apply_style` to `ppaged()` and `pfeedback()` is confusing since style is only applied when the `msg` text is sent by these functions to `poutput()`.

3. Adding the `apply_style` argument to `poutput()` breaks any existing code which prints already styled text since it defaults to `True`. In practice developers won't notice a visible change since `style_output` defaults to nothing. However, their code is still wrong since they aren't setting `apply_style` to `False` when they call `poutput`.

@tleonhardt What are your thoughts?

anselor:
I feel part of the issue is our current print methods conflate where to send output with how to send output. I've made some adjustments that move in the direction of keeping things separate.

anselor:
@tleonhardt @kmvanbrunt ?

cmd2/cmd2.py
```diff
@@ -1065,44 +1102,87 @@ def poutput(self, msg: Any = '', *, end: str = '\n') -> None:
...
+
+        :param msg: object to print
+        :param end: string appended after the end of the message, default a newline
+        :param paged: If True, pass the output through the configured pager.
+        :param chop: If paged is True, True to truncate long lines or False to wrap long lines.
+        """
+        self.print_to(self.stdout, msg, end=end, style=ansi.style_success, paged=paged, chop=chop)
 
-    def pwarning(self, msg: Any = '', *, end: str = '\n', apply_style: bool = True) -> None:
+    def pwarning(
```
kmvanbrunt, line 1150: Should the signature for `pwarning()` match that of `pfailure()` (i.e. remove `apply_style`).
This seems appropriate since we now have two wrappers for `perror()`. They should work the same way.

kmvanbrunt (changes requested):

cmd2/cmd2.py
anselor: I was reluctant to make an API breaking change to pwarning(). I don't know if it's possible to mark a single argument deprecated.

anselor: I added a deprecation to the parameter documentation. We can remove it in the major version revision.

PR #1272, opened 2023-06-29
Suggest most simillar command

### LINKED ISSUE #1271: Adding a did you mean? or most similar command in case of error
Would it be possible to display the most similar command in case of error?
I think I could create a PR adding this behavior to the default method.
I'm doing something like the snippet below in a subclass of cmd2.Cmd
What do you think?

``` python
    def default(self, statement: Statement) -> Optional[bool]:
        default_error = self.default_error
        MIN_SIMIL_TO_CONSIDER = 0.7
        try:
            commands = self.get_visible_commands()
            requested_command = statement.command
            commands_similarities = {
                x:jarowinkler_similarity(x,requested_command)
                for x in commands
            }
            proposed_command = None
            best_simil = None
            for each, simil in commands_similarities.items():
                if simil > MIN_SIMIL_TO_CONSIDER and (best_simil is None or best_simil < simil):
                    best_simil = simil
                    proposed_command = each
            if proposed_command is not None:
                self.default_error = self.default_error + os.linesep + f"Did you mean {proposed_command}?"
            return super().default(statement)
        finally:
            self.default_error = default_error
```

federicoemartinez:
closes #1271 
The app can be configured to suggest the most similar command. In that case, the most similar command of the visible commands (more similar than a threshold) will be suggested and added to the default error message.

By default SequenceMatcher.ratio() from the standard library is used but a subclass can define _suggest_similar_command with a new algorithm.

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -141,6 +141,7 @@ from .utils import (
     Settable,
     get_defining_class,
     strip_doc_annotations,
+    suggest_similar,
 )
 
 # Set up readline
@@ -236,6 +237,7 @@ class Cmd(cmd.Cmd):
         command_sets: Optional[Iterable[CommandSet]] = None,
         auto_load_commands: bool = True,
         allow_clipboard: bool = True,
+        suggest_similar_command: bool = False,
     ) -> None:
         """An easy but powerful framework for writing line-oriented command
         interpreters. Extends Python's cmd package.
@@ -529,6 +531,9 @@ class Cmd(cmd.Cmd):
 
         # Add functions decorated to be subcommands
         self._register_subcommands(self)
+
+        self.suggest_similar_command = suggest_similar_command
+        self.default_suggestion_message = "Did you mean {}?"
 
     def find_commandsets(self, commandset_type: Type[CommandSet], *, subclass_match: bool = False) -> List[CommandSet]:
         """
@@ -2968,10 +2973,16 @@ class Cmd(cmd.Cmd):
             return self.do_shell(statement.command_and_args)
         else:
             err_msg = self.default_error.format(statement.command)
-
+            if self.suggest_similar_command:
+                suggested_command = self._suggest_similar_command(statement.command)
+                if suggested_command:
+                    err_msg = err_msg + ' ' + self.default_suggestion_message.format(suggested_command)
             # Set apply_style to False so default_error's style is not overridden
             self.perror(err_msg, apply_style=False)
             return None
+
+    def _suggest_similar_command(self, command: str) -> Optional[str]:
+        return suggest_similar(command, self.get_visible_commands())
 
     def read_input(
         self,
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -547,7 +547,7 @@ class Cmd(cmd.Cmd):
         return [
             cmdset
             for cmdset in self._installed_command_sets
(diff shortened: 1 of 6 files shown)


cmd2/cmd2.py
```diff
@@ -529,6 +531,9 @@ def __init__(
         # Add functions decorated to be subcommands
         self._register_subcommands(self)
 
+        self.suggest_similar_command = suggest_similar_command
+        self.default_suggestion_message = "The most similar command is {}"
```
anselor, line 535: When I doubt I tend to fall back on what bash does. How about:
', did you mean: {}'

cmd2/cmd2.py
```diff
@@ -2866,11 +2871,17 @@ def default(self, statement: Statement) -> Optional[bool]:  # type: ignore[overr
             return self.do_shell(statement.command_and_args)
         else:
             err_msg = self.default_error.format(statement.command)
-
+            if self.suggest_similar_command:
+                suggested_command = self._suggest_similar_command(statement.command)
+                if suggested_command:
+                    err_msg = err_msg + os.linesep + self.default_suggestion_message.format(suggested_command)
```
anselor, line 2877: With my alternate suggestion message, no need for line separator.

cmd2/cmd2.py
federicoemartinez: It was my first choice. I changed it because of git:
```
git comit

git: 'comit' is not a git command. See 'git --help'.

The most similar command is
	commit
```
But sure, I can change it.

cmd2/cmd2.py
federicoemartinez: I can use a space

federicoemartinez:
I see mypy and some other jobs are failing but I don't think that's related to this branch

pablopissi:
It would be great to have this merged

anselor:
My PR just merged which also addresses many of the issues you probably ran into. You may want to rebase over latest to see.

anselor:
I'm not sure what's going on with your branch but my changes are now showing up as changes in your branch. I highly suggest squashing related commits and then rebasing it all on top of latest master.

federicoemartinez:
> I'm not sure what's going on with your branch but my changes are now showing up as changes in your branch. I highly suggest sq

I think it's done now. I made a mistake rebasing, I'm much more used to mercurial to be honest.
Thanks for pointing it out.

anselor (dismissed):

federicoemartinez:
@anselor I applied the changes from the formatter jobs and the doc job. Don't know about the linter. It looks like it is not from my PR

anselor:
> @anselor I applied the changes from the formatter jobs and the doc job. Don't know about the linter. It looks like it is not from my PR

There pyinvoke commands you can run locally to check

federicoemartinez:
I'm trying 

> > @anselor I applied the changes from the formatter jobs and the doc job. Don't know about the linter. It looks like it is not from my PR
> 
> There pyinvoke commands you can run locally to check

They passed now (Though with python 3.8 in my local setup).
The linter job still fails but it is totally unrelated to my branch

anselor:
@federicoemartinez The linter problem looks spurious and not applicable to the situation. Would you mind just adding `# noqa: E731` to those lines to make sure this doesn't hide something else?

federicoemartinez:
done! 
Does this job works for master?

anselor:
As far as I'm aware it passed before the last PR. What I've sometimes see happen is the validation library updates and new errors are found that did not previously fail.

Issue #1002, opened 2020-09-28
Help text duplicated when using more than 1 CLI instance


kmvanbrunt:
```Python
import sys
from cmd2 import cmd2

instance_1 = cmd2.Cmd()
instance_2 = cmd2.Cmd()
sys.exit(instance_2.cmdloop())
```
With this program, run `alias -h`. The subcommand help gets repeated.
```subcommands:
  SUBCOMMAND
    create    create or overwrite an alias
    delete    delete aliases
    list      delete aliases
    create    create or overwrite an alias
    delete    delete aliases
    list      delete aliases
```

Since the parsers (e.g. `Cmd.alias_parser`) are class objects, they get updated with each instantiation. This problem has existed since we added the `argparse` decorators, but it's being exercised a bit more since we added `CommandSet` and `as_subcommand_to` decorators. This specific issue of repeated text was introduced when cmd2.py started using the `as_subcommand_to` decorator.

For the 2.0 release, we need to find a way to make deep copies of the parsers and tie each copy to an instance. We won't have 1 parser attached to the functions anymore. Instead the decorators will need to retrieve the instance-specific parser.

tleonhardt:
I have a very vague recollection of researching how to make a deep copy of a parser and finding that it was possible in Python 3.7+ (or something similar) but not beforehand.

tleonhardt:
Yeah, inability to deep-copy an `ArgumentParser` was fixed in Python 3.7.  The context is in older issue #522

tleonhardt:
@kmvanbrunt #1005 appears to have fixed this, at least temporarily.  Are we awaiting a more permanent fix?  Or should this issue be closed?

kmvanbrunt:
@tleonhardt The permanent fix needs to be in 2.0. We need a way to support instance-specific parsers and that will most likely be a breaking change.

kmvanbrunt:
@anselor Do you have any thoughts on this issue?

anselor:
I'm sure I do I've just been busy with other more immediate deadlines and haven't had time to come back to this and other cmd2 stuff.

kmvanbrunt:
@anselor and I agree the 2.0 release does not have to wait for this fix. The risk of someone using two cmd2 objects and editing common parsers between them is pretty low.

tleonhardt:
@kmvanbrunt Is this fixed now?  Or are we waiting on some better fix?

federicoemartinez:
[...]

PR #1278, opened 2023-09-26
Changed with_argparser() and as_subcommand_to() to accept either an A…

anselor:
…rgumentParser or a

factory callable that returns an ArgumentParser.
Changed Cmd constructor to construct an instance-specific ArgumentParser using either the factory callable or by deep-copying the provided ArgumentParser.
With this change a new argparse instance should be created for each instance of Cmd.

Addresses #1002

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -30,6 +30,7 @@ Git repository on GitHub at https://github.com/python-cmd2/cmd2
 # setting is True
 import argparse
 import cmd
+import copy
 import functools
 import glob
 import inspect
@@ -55,6 +56,7 @@ from types import (
 )
 from typing import (
     IO,
+    TYPE_CHECKING,
     Any,
     Callable,
     Dict,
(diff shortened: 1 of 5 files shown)


cmd2/cmd2.py
```diff
@@ -659,6 +668,43 @@ def register_command_set(self, cmdset: CommandSet) -> None:
             cmdset.on_unregistered()
             raise
 
+    def _build_parser(self, parent: Union['Cmd', CommandSet], parser_builder: Any) -> Optional[argparse.ArgumentParser]:
```
federicoemartinez, line 671: parser_builder shouldn't be Union[argparse.ArgumentParser, Callable] ?

anselor: No, the point is to always result in an instance of a parser after this. either by deep-copying the provided one or calling the builder.

anselor: oh wait, misread your comment.

anselor: Ah, this got messy because `staticmethod` and `classmethod` are not considered callable for some reason and there's some complicated mess where mypy wants to treat them like a generic type but at runtime this causes a python parsing error and fails..

tleonhardt (approved):

PR #1291, opened 2024-01-31
Allow scripts to bypass adding their commands to the history

laf0rge:
At least in most of my use cases, the individual commands of a script executed by the command line should not end up in the history.  If I go back in history, I want to find the commands I typed (like the `run_script`), and not the commands contained in the script I executed, or even the scripts that this script might execute itself.  Imagine this happening on your bash/tcsh/zsh: Every time you execute a shell script your history is cluttered.

I took care to preserve the existing behavior by default in these patches.  However, if the cmd2 maintainers agree, it would of course also make sense (from my point of view) to simply never add the commands form scripts to the history and avoid all the configurability this patch introuces for backwards compatibility.  Or introduce this patch, and change the default to "don't add to history".  

Also, the naming of the settable and the run_script argument could possibly be improved,  I just didn't come up with somtehing better so far.


anselor (dismissed):
Changes seem reasonable to me.

cmd2/cmd2.py
```diff
@@ -5045,6 +5049,11 @@ def _current_script_dir(self) -> Optional[str]:
         help='record the output of the script as a transcript file',
         completer=path_complete,
     )
+    run_script_parser.add_argument(
+        '--history',
+        action=argparse.BooleanOptionalAction,
```
kmvanbrunt, line 5054: `BooleanOptionalAction` was added in Python 3.9. Currently, cmd2 supports 3.7 and higher.
Consider changing this option name to `--no-history` and changing action to `store_true`.

kmvanbrunt (changes requested):

cmd2/cmd2.py
tleonhardt: @laf0rge Would it be possible for you to make this change so it can work with Python 3.8?

tleonhardt:
@kmvanbrunt  That is a philosophical question.  I can see advantages and disadvantages to either approach. Whatever works best for the majority of users is probably the right answer.

My gut instinct is we shouldn't.  But it depends on whether we want it to be more of a command history or an audit trail?

Perhaps doing whatever bash command history would do in a similar situation would be the safest approach regarding customer expectations for behavior.

kotfu:
I don't have a strong opinion on what the default should be, but I love having the setting so people can choose

kmvanbrunt:
@tleonhardt @kotfu @anselor @laf0rge 
I completed this PR by accounting for the new settable in `run_pyscript`, adding unit tests, and making documentation updates.

Since the original author of this PR wasn't fond of his settable name (`history_includes_scripts`), I changed it to `save_scripted_commands`. I also considered `log_scripted_commands`, but I didn't want to confuse those who use cmd2-based apps with a log file. I'm open to changing the settable name if you have suggestions.

I also removed his `--history` option to  `run_script` which temporarily overrides `self.save_scripted_commands`. If you think it should be restored, let me know and I will also add it to `run_pyscript`.

tleonhardt (approved):
I like the approach taken

PR #1318, opened 2024-08-19
Handle absent lzma library

stephenfin:
The lzma library may be missing in some environments as it is an optional dependency [1]. Fall back to bz2 in this instance.

[1] https://devguide.python.org/getting-started/setup-building/

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -4627,7 +4627,7 @@ class Cmd(cmd.Cmd):
             )
 
             # Start IPython
-            start_ipython(config=config, argv=[], user_ns=local_vars)
+            start_ipython(config=config, argv=[], user_ns=local_vars)  # type: ignore[no-untyped-call]
             self.poutput("Now exiting IPython shell...")
 
             # The IPython application is a singleton and won't be recreated next time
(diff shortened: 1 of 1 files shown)


cmd2/cmd2.py
```diff
@@ -4813,7 +4813,12 @@ def _initialize_history(self, hist_file: str) -> None:
                           to this file when the application exits.
         """
         import json
-        import lzma
+        try:
+            import lzma
+            bz2 = None
```
tleonhardt, line 4818: Overall your logic looks flawed.  

There appears to be no need for a `bz2` variable, but in this case you should set `lzma` to something, perhaps `True`.

Same comment applies to similar logic below.

In other case probably best to set witch to `False` instead of `None` to make this boolean and switch conditional logic as well.

tleonhardt (changes requested):

cmd2/cmd2.py
osfrickler: How about:

```
try:
    import lzma as compress_lib
except ModuleNotFoundError:  # pragma: no cover
    import bz2 as compress_lib
```

and then you can simply use `compress_lib.decompress()` unconditionally below?

tleonhardt: @osfrickler Yes I like that solution better, it is very clean.  Great suggestion.

kmvanbrunt:
@osfrickler I went ahead and made the suggested changes since the function needed to be refactored a bit.
Let me know if my changes match your intentions.

PR #1349, opened 2024-10-24
Updated pypi_test task to use --repository flag in its twine command.

kmvanbrunt:
This is to support using PyPI API tokens in a .pypirc file like the following.

```
[testpypi]
username = __token__
password = pypi-AgEN...
```


tleonhardt:
Recently I've noticed that the `test_proc_reader_send_sigint` unit test is mildly unreliable.  Once in a blue moon it fails.

tleonhardt (approved):

PR #1356, opened 2024-11-01
Integrate rich-argparse

kmvanbrunt:
Added `rich-argparse` to `cmd2`.

One big change is that `Cmd2ArgumentParser` no longer defaults to using a raw help text formatter. Since we can use Rich renderables in descriptions and epilogs, I was able to get the desired paragraph separation with rich tables.

Only the history command needs a raw help text formatter now because of how we print the help for `arg`.
Since an argument's help text can't be a Rich renderable we have to include `\n` characters and only raw help text formatters preserve newlines.

This change is for `cmd2 3.0.0`.

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -220,10 +220,6 @@ class Cmd(cmd.Cmd):
 
     DEFAULT_EDITOR = utils.find_editor()
 
-    INTERNAL_COMMAND_EPILOG = (
-        "Notes:\n" "  This command is for internal use and is not intended to be called from the\n" "  command line."
-    )
-
     # Sorting keys for strings
     ALPHABETICAL_SORT_KEY = utils.norm_fold
     NATURAL_SORT_KEY = utils.natural_keys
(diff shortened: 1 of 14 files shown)


cmd2/cmd2.py
```diff
@@ -4778,15 +4782,16 @@ def _generate_transcript(
...
-            "  set editor (program-name)"
-        )
+        from rich.markdown import Markdown
 
+        edit_description = "Run a text editor and optionally open a file with it."
         edit_parser = argparse_custom.DEFAULT_ARGUMENT_PARSER(description=edit_description)
+
+        edit_parser.epilog = edit_parser.create_text_group(
+            "Note",
+            Markdown("To set a new editor, run: `set editor <program>`"),
```
tleonhardt, line 4792: Being able to use Markdown is really cool.

tleonhardt (approved):

tleonhardt:
I'll have to experiment with how this looks and such, but tentatively I think it is really cool and think we might want to put out an intermediate release before 3.0.0, e.g. 2.6.0, since from a quick look this doesn't really seem to break backwards compatibility in any major way.

PR #1371, opened 2024-11-05
Change CommandSet._cmd to a read-only property which never returns None.

kmvanbrunt:
This addresses issue where `CommandSet` methods which only run after it is registered have to either check if `self._cmd` is `None` or cast it to the CLI class to avoid type checker errors.


cmd2/command_definition.py
```diff
@@ -92,20 +92,42 @@ class CommandSet(object):
     """
 
     def __init__(self) -> None:
-        self._cmd: Optional[cmd2.Cmd] = None
+        # Private reference to the CLI instance in which this CommandSet running.
+        # This will be set when the CommandSet is registered and it should be
+        # accessed by child classes using the self._cmd property.
+        self.__private_cmd: Optional[cmd2.Cmd] = None
```
anselor, line 98: My convention, in python, double-underscore is already private. Python automatically adds the class name to the variable at runtime to prevent easy access externally.

kmvanbrunt: That's true, but I didn't want the property name so close to the member variable (differing only by one underscore). It gets confusing to read since the `CommandSet` class does use variable directly and I want it obvious that it's not using the property.

I could call it 'self.__internal_cmd' or something else without the word private in it.

tleonhardt: I like the idea of something like `self.__internal_cmd`

tleonhardt (approved):

cmd2/command_definition.py
kmvanbrunt: I went with `self.__cmd_internal`.

PR #1395, opened 2024-12-17
Rolled back undocumented changes to printing functions introduced in 2.5.0.

### LINKED ISSUE #1390: cmd2 2.5.0 changed the method chain for output
To describe this issue, I need to define what I mean by "method chain for output". If you are a developer of a `cmd2` based app, you would typically call one of the `cmd2` methods to produce output, i.e, one of `poutput()`, `perror()`, `pwarning()`, `pfeedback()`, or `ppaged()`. The `cmd2.py` implementation of these methods appropriately called each other. For example, `ppaged()` makes a determination whether output is going to a terminal, if so, it opens a pipe and sends the output to your operating system pager (like `less`), but if output is not going to a terminal, `ppaged()` just calls `poutput()`. Similarly `pwarning()` applies a different style and then calls `perror()`. This is what I mean by the "method chain for output".

In `cmd2` version 2.4.3 and prior, the method chain for output was:

* `poutput()` -> `style_aware_write()`
* `perror()` -> `style_aware_write()`
* `pwarning()` -> `perror()`
* `pexcept()` -> `perror()`
* `pfeedback()` -> `poutput()` or `perror()`
* `ppaged()` -> `poutput()` if not on a functional terminal

`cmd2` version 2.5.0 made some material changes to this method chain. It introduces a new method `print_to()`

* `print_to()` -> `ppaged()` or `style_aware_write()`
* `poutput()` -> `print_to()`
* `perror()` -> `print_to()`
* `pwarning()` -> `print_to()`
* `pexcept()` -> `perror()`
* `pfeedback()` -> `print_to()`
* `ppaged()` -> `style_aware_write()`

What is the rationale for this new design? I have some specific questions:
* what's the purpose of the `print_to()` method and what guidance is there for developers for the use/overriding of that method?
* why does `print_to()` call `ppaged()`?
* why doesn't `pexcept()` call `print_to()`?
* why doesn't `ppaged()` call `poutput()` any more?
* why does `pexcept()` still call `perror()` but `pwarning()` doesn't?

Besides the philosophical questions to help me understand this better, this change broke my test suite because I had reimplemented `poutput()` to use `rich.console`. In v2.4.3 when you called `ppaged()` the call chain went to `poutput()` if there was no terminal present. In 2.5.0, `ppaged()` no longer calls `poutput()` so I couldn't intercept the output in the way I was previously doing. There is a relatively simple workaround for me in my app, I can just reimplement `ppaged()`.

One other note, the method signatures for `print_to()` and `style_aware_write()` use a fundamentally different design than `rich.console`. Meaning, if I as a developer of a `cmd2` based app wanted to use `rich.console` for my output, I need to not use any of the `cmd2` output methods or I need to reimplement them all. Even more discouraging, if you are using `rich.console`, there is now no method you can reimplement and get all of the `cmd2` generated output to go to `rich.console`

The `cmd2` approach is based on file descriptors passed a parameters to the methods: you set the file descriptors to be where you want the output to go and then pass the file descriptor around to the various methods which write to that file descriptor. In `rich.console` you create a console object give it the file descriptor to use, but to generate output you must call methods on the console object.

kmvanbrunt:
Fixes #1390.

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -1211,8 +1211,21 @@ class Cmd(cmd.Cmd):
         return ansi.strip_style(self.prompt)
 
     def print_to(
-        self, dest: Union[TextIO, IO[str]], msg: Any, *, end: str = '\n', style: Optional[Callable[[str], str]] = None
+        self,
+        dest: IO[str],
+        msg: Any,
+        *,
+        end: str = '\n',
+        style: Optional[Callable[[str], str]] = None,
     ) -> None:
+        """
+        Print message to a given file object.
+
+        :param dest: the file object being written to
+        :param msg: object to print
+        :param end: string appended after the end of the message, default a newline
+        :param style: optional style function to format msg with (e.g. ansi.style_success)
+        """
         final_msg = style(msg) if style is not None else msg
         try:
             ansi.style_aware_write(dest, f'{final_msg}{end}')
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -1211,124 +1211,55 @@ class Cmd(cmd.Cmd):
         return ansi.strip_style(self.prompt)
 
     def print_to(
-        self,
-        dest: Union[TextIO, IO[str]],
-        msg: Any,
-        *,
-        end: str = '\n',
-        style: Optional[Callable[[str], str]] = None,
-        paged: bool = False,
-        chop: bool = False,
+        self, dest: Union[TextIO, IO[str]], msg: Any, *, end: str = '\n', style: Optional[Callable[[str], str]] = None
     ) -> None:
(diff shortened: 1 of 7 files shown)


cmd2/cmd2.py
```diff
@@ -1211,124 +1211,55 @@ def visible_prompt(self) -> str:
         return ansi.strip_style(self.prompt)
 
     def print_to(
```
tleonhardt, line 1213: This is a lot cleaner here. Now `print_to` is doing one thing and doing it well.  I'd recommend adding a brief docstring to explain the purpose.

tleonhardt (dismissed):

cmd2/cmd2.py
kmvanbrunt: Added docstring.

PR #1444, opened 2025-06-08
This PR forces the use of `from __future__ import annotations` throughout cmd2

tleonhardt:
This PR forces the use of `from __future__ import annotations` throughout `cmd2`.

Advantages of this include:
- No need for `Optional`, can use `| None` instead
- No need for `Union` in type hints, can use stuff like `int | str` instead
  - Still need to use `Union` for things like casts and type vars for versions prior to Python 3.10
- No need for quotes around forward reference types
- Allows moving imports used only for type hints under an `if TYPE_CHECKING:` block
- Can theoretically improve runtime performance by delaying processing of type annotations and imports until it is needed
  - In practice this performance benefit is fairly negligible

The only real disadvantage I can see if we will probably need to leave that import in place until the oldest version we support is Python 3.14 since that is when lazy eval of type annotations is in Python by default. Either that or we might need to restore quotes around types in a couple places where they were used for forward references.

NOTE: Nearly all of the changes in this PR were done automatically by `ruff` after I changed the configuration for `ruff` in `pyproject.toml`.


tleonhardt:
@kmvanbrunt If you get a chance, I'd appreciate some help with manually testing this one. It seems like a good change to me, but it a large and subtle one. I want to make sure it doesn't negatively impact things like type awareness in IDEs and such.

PR #1452, opened 2025-06-26
Simplify Python 3.14 check in argparse_custom.py

tleonhardt:
Simplify Python 3.14 check in `argparse_custom.py`

### DIFF
--- a/cmd2/argparse_custom.py
+++ b/cmd2/argparse_custom.py
@@ -1215,42 +1215,30 @@ class Cmd2ArgumentParser(argparse.ArgumentParser):
                                   behavior on this parser. If this is None or not present, then cmd2 will use
                                   argparse_completer.DEFAULT_AP_COMPLETER when tab completing this parser's arguments
         """
+        kw_args: dict[str, bool] = {}
         if sys.version_info >= (3, 14):
             # Python >= 3.14 so pass new arguments to parent argparse.ArgumentParser class
-            super().__init__(
-                prog=prog,
-                usage=usage,
-                description=description,
-                epilog=epilog,
-                parents=parents if parents else [],
-                formatter_class=formatter_class,  # type: ignore[arg-type]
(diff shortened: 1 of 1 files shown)


cmd2/argparse_custom.py
```diff
@@ -1215,42 +1215,30 @@ def __init__(
                                   behavior on this parser. If this is None or not present, then cmd2 will use
                                   argparse_completer.DEFAULT_AP_COMPLETER when tab completing this parser's arguments
         """
+        kw_args: dict[str, bool] = {}
```
kmvanbrunt: Consider making this `dict[str, Any]` to account for future args.

cmd2/argparse_custom.py
```diff
@@ -1215,42 +1215,30 @@ def __init__(
...
-                formatter_class=formatter_class,  # type: ignore[arg-type]
-                prefix_chars=prefix_chars,
-                fromfile_prefix_chars=fromfile_prefix_chars,
-                argument_default=argument_default,
-                conflict_handler=conflict_handler,
-                add_help=add_help,
-                allow_abbrev=allow_abbrev,
-                exit_on_error=exit_on_error,  # added in Python 3.9
-            )
+            kw_args = {
```
kmvanbrunt: Can we call this `kwargs` to align with the typical Python name?

kmvanbrunt (changes requested):

cmd2/argparse_custom.py
tleonhardt: Sure, renamed as suggested

cmd2/argparse_custom.py
tleonhardt: I want to avoid `Any` where possible because it basically opts out of type checking.  In the future we can make it whatever union type need be; but for now it is an accurate type.

PR #1458, opened 2025-07-05
Parser building functions.

kmvanbrunt:
All built-in commands now use a function to create their argument parser.
This simplifies the process for overriding cmd2's default parser class.


tleonhardt (dismissed):
Everything here looks good to me.

Would it be feasible to move most of the parser building logic to a separate file? If so would it be desirable?

My thought is that `cmd2.py` is a pretty large file. If it is possible to move a good chunk of related logic to a separate file, that might be a good thing. Given that many if not most of the parser builder functions are class methods, I suspect this might be impossible or painful.  But it's worth a thought.

tleonhardt (approved):
LGTM

PR #1462, opened 2025-07-12
No longer setting parser's prog value in with_argparser().

kmvanbrunt:
No longer setting parser's prog value in with_argparser() since it gets set in Cmd._build_parser().

This code had previously been restored to support backward compatibility in cmd2 2.0 family.
Now that we are moving to cmd2 3.0, it can be removed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 - Breaking Changes
 
     - Removed macros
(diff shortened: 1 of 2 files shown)


tleonhardt (changes requested):
Please add a breaking change note to the CHANGELOG

PR #1464, opened 2025-07-17
Restore macros

kmvanbrunt:
Restored macros and added ability to implement custom tab completion for macro arguments.


tleonhardt (approved):
LGTM

PR #1468, opened 2025-07-21
No longer overwriting sys.stderr.

kmvanbrunt:
- In an effort be consistent with the purpose of `self.stdout` and our own documentation, the following changes were made.
    - No longer redirecting `sys.stdout`.
    - No longer capturing pyscript output written to `sys.stdout`.
        - To assist with this change, calling `print()` within a pyscript now writes to `self.stdout`. Calling `self.poutput()` within a pyscript is still preferred, but that may not always be possible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@
       `Cmd._build_parser()`. This code had previously been restored to support backward
       compatibility in `cmd2` 2.0 family.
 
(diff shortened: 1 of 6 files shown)


tleonhardt (approved):
It isn't clear to me if it is better to do this or to eliminate self.stdout.  

I think this is probably fine.

Issue #1351, opened 2024-10-24
cmd2 3.0.0 Discussion


kmvanbrunt:
Now that `cmd2 2.5.0` is out, I'd like to start thinking about `cmd2 3.0.0`. So here are a few features I'd like to see based on feedback we've received over the years.

- Full `rich` and `rich-argparse` integration.
  - I wrote the table and related formatting code in early 2020 to solve a problem that all existing table creators at the time had. None of them preserved text styles across multiple lines of a table cell. That was also right around the time `rich` was developing their table functionality so I wasn't aware of their library. In the years since, `rich` has become the de facto library for formatting text in the terminal and it's widely used and well supported. So I think it's safe at this point to offload all of our text formatting to `rich`. This approach was also discussed in #1251 .
  
  - I have been testing our stuff with `rich-argparse` and I'm finding the switch very easy.

- Make async alerts and prompt changes event driven.
  - Currently developers must obtain `self.terminal_lock` to asynchronously print an alert. In `cmd2 3.0` they will call a function like `add_alert()` which adds to a queue and `cmd2` will handle printing alerts when the prompt is on screen. Async prompt updates will provide a similar method.

- Consider moving some built-in commands to `CommandSets` or mixin classes.
  - What commands are best suited for this?

- Remove macros.
  - They were a neat idea at the time, but as [1350](https://github.com/python-cmd2/cmd2/discussions/1350) points out, the lack of custom tab completion makes them difficult and confusing to use. Removing macros also eliminates the need for `self._input_line_to_statement()`. We could just call `self._complete_statement()` directly.
  - No longer removing macros since people are using them. I added ability to develop custom macro tab completion.

- #1334 
  - Moving the `ext_test` plugin might help here as well. It currently has its own build system and is required to run our unit tests. Can it become a part of the main `cmd2` code or at least not be needed to run our unit tests?

tleonhardt:
I think replacing most of cmd2's custom text formatting code with a combination of [rich](https://github.com/Textualize/rich) and [rich-argparse](https://github.com/hamdanal/rich-argparse) is a fantastic idea. It will reduce the amount of custom code that we need to maintain for `cmd2` and allow us to bring in some additional functionality while presenting our users with the now idiomatic features available within `rich`.

[...]

PR #1471, opened 2025-07-30
Enabled all printing methods to support Rich objects.

kmvanbrunt:


cmd2/rich_utils.py
```diff
@@ -6,13 +6,22 @@
     IO,
```
tleonhardt, line 6: Information on this `rich_utils` module should be added to the documentation, both in the [API Reference](https://cmd2.readthedocs.io/en/stable/api/) and in the section on [Generating Output](https://cmd2.readthedocs.io/en/stable/features/generating_output/). Additionally, something should likely be added to the `CHANGELOG.md` for the 3.0 release about the new public functionality added here.

cmd2/rich_utils.py
```diff
@@ -88,6 +97,28 @@ def set_theme(new_theme: Cmd2Theme) -> None:
         RichHelpFormatter.styles[name] = THEME.styles[name]
 
 
+class RichPrintKwargs(TypedDict, total=False):
```
tleonhardt, line 100: This is a rare case where a using a `TypedDict` actually seems preferable to using a `dataclass`

cmd2/cmd2.py
```diff
@@ -950,11 +948,11 @@ def _register_subcommands(self, cmdset: Union[CommandSet, 'Cmd']) -> None:
                 command_func = self.cmd_func(command_name)
 
             if command_func is None:
-                raise CommandSetRegistrationError(f"Could not find command '{command_name}' needed by subcommand: {method!s}")
+                raise CommandSetRegistrationError(f"Could not find command '{command_name}' needed by subcommand: {method}")
```
tleonhardt, line 951: Is there a reason we had the `!s` before.  `f-string` formatting is peculiar where some most types format using `__str__`; however some types default to formatting using `__repr__`.  Adding the `!s` forces to use `__str__` regardless of type.

This applies in all similar method calls.

An underlying question might reasonably be: "What type is `method` and do we know how an f-string will format that?".

cmd2/cmd2.py
```diff
@@ -1189,170 +1187,364 @@ def visible_prompt(self) -> str:
 
     def print_to(
```
tleonhardt, line 1188: The `CHANGES.md` file needs to be updated with breaking changes for all of these output methods that had their API change in a backwards incompatible way. We also need to massively overhaul the `Generating Output` section of the documentation.

cmd2/cmd2.py
```diff
@@ -1189,170 +1187,364 @@ def visible_prompt(self) -> str:
...
-        msg = ansi.style_success(msg)
-        self.poutput(msg, end=end)
-
-    def pwarning(self, msg: Any = '', *, end: str = '\n') -> None:
-        """Wrap perror, but applies ansi.style_warning by default.
+        self.poutput(
+            *objects,
+            sep=sep,
+            end=end,
+            style="cmd2.success",
```
tleonhardt, line 1336: These style string constants like `cmd2.success`, `cmd2.warning`, and `cmd2.error` feel like they should be constants defined somewhere.

I realize we have  dict mapping these strings to actual rich styles called `DEFAULT_CMD2_STYLES` in `rich_utils.py`.

I'm saying that before declaring `DEFAULT_CMD2_STYLES`, we should define constants for all of the strings used in keys in that dict.  They could be something like `STYLE_ERROR` or we could even implement something like a frozen dataclass called something like `cmd2.Style` and members called stuff like `error` equal to `"cmd2.error"`.  That way you could tab complete import `Style` and use `Style.error`. There are a lot of possible ways to skin this cat, but they should definitely be constants that are tab-completable in an IDE and type-checkable by the IDE as well.

cmd2/cmd2.py
```diff
@@ -1189,170 +1187,364 @@ def visible_prompt(self) -> str:
...
-                if self.broken_pipe_warning:
-                    sys.stderr.write(self.broken_pipe_warning)
-        else:
-            self.poutput(msg, end=end)
+            # Prevent KeyboardInterrupts while in the pager. The pager application will
+            # still receive the SIGINT since it is in the same process group as us.
+            with self.sigint_protection:
+                import subprocess
 
-    def ppretty(self, data: Any, *, indent: int = 2, width: int = 80, depth: Optional[int] = None, end: str = '\n') -> None:
```
tleonhardt, line 1346: Since `ppretty` is gone, we need to mention this in the breaking changes in the `CHANGELOG` and we should also add an example showing how to do pretty printing of objects like dicts using both `rich` and `pprint`

cmd2/cmd2.py
```diff
@@ -2380,13 +2576,17 @@ def get_help_topics(self) -> list[str]:
         # Filter out hidden and disabled commands
         return [topic for topic in all_topics if topic not in self.hidden_commands and topic not in self.disabled_commands]
 
-    def sigint_handler(self, signum: int, _: Optional[FrameType]) -> None:  # noqa: ARG002
+    def sigint_handler(
+        self,
+        signum: int,  # noqa: ARG002,
+        frame: Optional[FrameType],  # noqa: ARG002,
```
tleonhardt, line 2582: `frame` is still unused right?  Or does it have to be called that?

tleonhardt (changes requested):
Overall looks good.  Mainly need to update `CHANGELOG` and documentation. But various comments, at least some of which should likely be addressed.

cmd2/cmd2.py
kmvanbrunt: It bothered me that it wasn't named since it's a part of our public API.

cmd2/cmd2.py
kmvanbrunt: `DEFAULT_CMD2_STYLES` contains all Rich-provided styles as well and Rich doesn't have a list of constants.

```Python
DEFAULT_CMD2_STYLES.update(RichHelpFormatter.styles.copy())
```

Unless we define constants for all their styles and continually update them, a tab-completable list will be incomplete.

cmd2/rich_utils.py
kmvanbrunt: I still go back and forth on this decision, so I'm glad you approve of it.

cmd2/rich_utils.py
kmvanbrunt: I'm saving the documentation updates until all this Rich integration is done. Once I start using the new print methods, I may change them.

cmd2/cmd2.py
kmvanbrunt: I'll add an example at a later date once I learn the Rich pretty print API.

cmd2/cmd2.py
tleonhardt: I'm ok with an incomplete list as long as all custom cmd2 styles have constants. I'd also be ok with a complete list that we need to maintain if it isn't too painful to maintain.

cmd2/cmd2.py
tleonhardt: Do we need it as part of a signint handler?  Can we remove it? Alternatively we can call it `_frame`

cmd2/rich_utils.py
tleonhardt: For now I added a couple items to our running TODO list for 3.0: https://github.com/python-cmd2/cmd2/issues/1351#issuecomment-3037221242

cmd2/cmd2.py
kmvanbrunt: The best info I can find for this use of `{method!s}` is that it's an explicit conversion so other developers know you want to convert to a string.

But I can't find an example where it would matter. If a custom Callable object doesn't define __str__(), the f-string would use __repr__() with or without the `!s`.

Does my line of thinking seem correct?

tleonhardt: Your line of reasoning seems correct.

I just got bit on a project recently where I tried removing the explicit `!s` in some `f-strings` and tests started failing because instead of the simple user-readable `__str__` description for some custom objects, it started printing out the very complex and not user friendly `__repr__`.

cmd2/cmd2.py
kmvanbrunt: I misspoke earlier. `DEFAULT_CMD2_STYLES` contains our styles and rich-argparse's styles.
But our `Cmd2Theme` recognizes everything in `DEFAULT_CMD2_STYLES` and every style provided by Rich.

https://github.com/Textualize/rich/blob/master/rich/default_styles.py

cmd2/cmd2.py
kmvanbrunt: Yes, Python passes both parameters to us.

`signum` isn't used either so we'd have to rename it to `_signum`.

cmd2/cmd2.py
tleonhardt: I added an item to our running `TODO` list to address this later

cmd2/cmd2.py
```diff
@@ -1189,170 +1187,364 @@ def visible_prompt(self) -> str:
...
+                    *objects,
+                    sep=sep,
+                    end=end,
+                    style=style,
+                    soft_wrap=soft_wrap,
+                    rich_print_kwargs=rich_print_kwargs,
+                )
 
-    def ppaged(self, msg: Any, *, end: str = '\n', chop: bool = False) -> None:
+    def ppaged(
```
tleonhardt, line 1457: [Coverage](https://app.codecov.io/gh/python-cmd2/cmd2/pull/1471?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=python-cmd2) of `ppaged` seems a tad less than it could be.  Take a look at uncovered lines and see if you could add a unit test.

cmd2/cmd2.py
tleonhardt: I think we should be safe and leave it in for now unless @anselor can comment regarding the original purpose and whether or not it is needed.

cmd2/cmd2.py
tleonhardt: This will inevitably require mocking using something like the `mocker` fixture from `pytest-mock` or `MagicMock` from `unittest`.

theagilehacker (approved):
I learned a couple of new Python tricks reviewing this change.
I am still not quite versed in the code base enough to grok it enough to make any design/code suggestions, other than logic issues, of which there are non that I could see.

cmd2/cmd2.py
kmvanbrunt: I'm going to leave these changes in for now.

tleonhardt: Sounds good

cmd2/cmd2.py
kmvanbrunt: Done

cmd2/cmd2.py
kmvanbrunt: Change log is updated.

tests/test_cmd2.py
```diff
@@ -2525,36 +2529,38 @@ def test_pexcept_not_exception(base_app, capsys) -> None:
...
+    if not sys.platform.startswith('win'):
+        os.environ.setdefault('TERM', 'simulated')
 
-@with_ansi_style(rich_utils.AllowStyle.ALWAYS)
-def test_ppaged_strips_ansi_when_redirecting_if_always(outsim_app) -> None:
+    # This will force ppaged to call Popen to run a pager
+    popen_mock = mock.MagicMock(name='Popen')
+    monkeypatch.setattr("subprocess.Popen", popen_mock)
+    outsim_app.ppaged("Test", chop=True)
+    popen_mock.assert_called_once()
```
tleonhardt: Can we test something stronger other than `Popen` is being called?

tleonhardt (dismissed):

tests/test_cmd2.py
kmvanbrunt: I updated it to make sure Popen is running either self.pager or self.pager_chop.

PR #1475, opened 2025-08-13
All tables in cmd2 are now rich-based.

kmvanbrunt:


cmd2/rich_utils.py
```diff
@@ -44,34 +50,55 @@ def __repr__(self) -> str:
...
+    EXAMPLE = "cmd2.example"
+    HELP_HEADER = "cmd2.help.header"
+    HELP_TITLE = "cmd2.help.title"
+    RULE_LINE = "cmd2.rule.line"
+    SUCCESS = "cmd2.success"
+    WARNING = "cmd2.warning"
+
+
+# Default styles used by cmd2
 DEFAULT_CMD2_STYLES: dict[str, StyleType] = {
```
tleonhardt, line 72: We should give users the ability to override these default styles for complete customizability of the look-and-feel of their application.

cmd2/rich_utils.py
```diff
@@ -44,34 +50,55 @@ def __repr__(self) -> str:
...
-        :param inherit: Inherit default styles. Defaults to True.
         """
-        cmd2_styles = DEFAULT_CMD2_STYLES.copy() if inherit else {}
+        cmd2_styles = DEFAULT_CMD2_STYLES.copy()
+
+        # Include default styles from rich-argparse
+        cmd2_styles.update(RichHelpFormatter.styles.copy())
+
         if styles is not None:
             cmd2_styles.update(styles)
```
tleonhardt, line 98: We should add a unit test where `styles` is not `None` and verify expected behavior.

cmd2/rich_utils.py
kmvanbrunt: They can do that by setting a theme.

cmd2/rich_utils.py
kmvanbrunt: Once I get to the task to set a theme, I will write unit tests for this.

tleonhardt (approved):

PR #1478, opened 2025-08-19
Refactored the help functions.

kmvanbrunt:


cmd2/cmd2.py
```diff
@@ -4039,6 +4042,37 @@ def complete_help_subcommands(
         completer = argparse_completer.DEFAULT_AP_COMPLETER(argparser, self)
         return completer.complete_subcommand_help(text, line, begidx, endidx, arg_tokens['subcommands'])
 
+    def _build_command_info(self) -> tuple[dict[str, list[str]], list[str], list[str], list[str]]:
```
tleonhardt, line 4045: Would using a dataclass or `NamedTuple` be more readable than a tuple with 4 elements?  Its fine as is, just thinking.

tests/test_cmd2.py
```diff
@@ -1278,6 +1283,18 @@ def help_app():
     return HelpApp()
 
 
+def test_help_headers(capsys) -> None:
+    help_app = HelpApp()
+    help_app.onecmd_plus_hooks('help')
+    out, err = capsys.readouterr()
+
+    assert help_app.doc_leader in out
```
tleonhardt, line 1294: For these asserts to really be testing something useful, I think you need to substantially update the `HelpApp` class to actually set those headers.  Currently it does not.

tleonhardt (changes requested):
Need to improve `HelpApp` class in `test_cmd2.py` to better coverage the stuff that is intended.

cmd2/cmd2.py
kmvanbrunt: I added a detailed docstring to this method.

tests/test_cmd2.py
kmvanbrunt: cmd2 was already setting these members, but I went ahead and explicitly set them in HelpApp just to be extra careful.

cmd2/cmd2.py
```diff
@@ -4111,14 +4170,76 @@ def print_topics(self, header: str, cmds: list[str] | None, cmdlen: int, maxcol:
...
+        self.poutput()
+
+    def _print_documented_command_topics(self, header: str, cmds: list[str], verbose: bool) -> None:
+        """Print topics which are documented commands, switching between verbose or traditional output."""
+        import io
+
+        if not cmds:
+            return
+
+        if not verbose:
```
tleonhardt, line 4191: Nice cleanup.  Consider taking it further and doing:

```Python
if not verbose:
    self.print_topics(header, cmds, 15, 80)
    return
````

And then getting rid of the `else:` block and un-indenting the rest of the code to the main level.

tleonhardt (approved):

cmd2/cmd2.py
kmvanbrunt: Done

tleonhardt (approved):

PR #1479, opened 2025-08-19
Simplied string_to_rich_text().

kmvanbrunt:


theagilehacker:
Changes look good. Nice simplification.

cmd2/rich_utils.py
```diff
@@ -188,36 +188,18 @@ def rich_text_to_string(text: Text) -> str:
...
-        "\x1d",  # Group Separator
-        "\x1e",  # Record Separator
-        "\x85",  # Next Line (NEL)
-        "\u2028",  # Line Separator
-        "\u2029",  # Paragraph Separator
-    }
-    if text and text[-1] in line_break_chars:
-        # We use "\n" because Text.from_ansi() converts all line breaks chars into newlines.
+    # Handle case where Text.from_ansi() removed the trailing newline.
+    if text.endswith("\n"):
```
tleonhardt: Will this really work generically in all the cases the previous solution would?  If so great.

tleonhardt (approved):

tleonhardt (dismissed):

kmvanbrunt:
I went back to the original approach since it's thorough. I just updated comments instead.

cmd2/rich_utils.py
kmvanbrunt: I decided to revert to the old code.

PR #1487, opened 2025-08-24
Added rich_utils.indent() to ensure indented text wraps instead of truncating when soft_wrap is enabled.

kmvanbrunt:


cmd2/rich_utils.py
```diff
@@ -288,6 +293,30 @@ def string_to_rich_text(text: str) -> Text:
     return result
 
 
+def indent(renderable: RenderableType, level: int) -> Padding:
```
tleonhardt, line 296: Strongly recommend adding unit tests specifically for this function, both happy and unhappy path.

tleonhardt (dismissed):

PR #1489, opened 2025-08-25
Styled string conversion

kmvanbrunt:
Fixed a few issues related to converting strings to Text before rendering.
Previously we converted all non-Rich objects to a Text object. Now we only convert strings which contain ANSI styles.

Rich only looks for markup, emojis, and highlighting in strings. Therefore we were losing this capability since they had been converted to Text objects. Additionally, we lost pretty printing on objects like dictionaries because they also had been converted to Text by our code.

Finally, I fixed an issue with incorrect display widths of styled_strings in CompletionItems descriptions.

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -1225,7 +1225,7 @@ class Cmd(cmd.Cmd):
                        method and still call `super()` without encountering unexpected keyword argument errors.
                        These arguments are not passed to Rich's Console.print().
         """
-        prepared_objects = ru.prepare_objects_for_rich_print(*objects)
+        prepared_objects = ru.prepare_objects_for_rich_rendering(*objects)
 
         try:
             Cmd2GeneralConsole(file).print(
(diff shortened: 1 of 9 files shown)


cmd2/cmd2.py
```diff
@@ -2508,7 +2508,11 @@ def _get_settable_completion_items(self) -> list[CompletionItem]:
         results: list[CompletionItem] = []
 
         for cur_key in self.settables:
-            descriptive_data = [self.settables[cur_key].get_value(), self.settables[cur_key].description]
+            settable = self.settables[cur_key]
```
tleonhardt: Is `self.settables` a dict?  If so better to just do:

```Python
for cur_key, settable in self.settables.items():
    descriptive_data = [
        str(settable.get_value()),
        settable.description,
    ]
    results.append(CompletionItem(cur_key, descriptive_data))
```

To prevent having to retrieve settable within the loop.

tleonhardt (dismissed):

cmd2/cmd2.py

tleonhardt (approved):

PR #1498, opened 2025-09-01
Simplify subcommand example within argparse_example.py

tleonhardt:
Simplify subcommand example within `argparse_example.py` to be a simple `calculate` command with `add` and `subtract` subcommands.

The original subcommand example was overly complicated. This replaces it with a much simpler one that is easier to digest and understand. 

Also switched to using the `cmd2.as_subcommand_to` decorator.

### DIFF
--- a/examples/argparse_example.py
+++ b/examples/argparse_example.py
@@ -106,98 +106,45 @@ class ArgparsingApp(cmd2.Cmd):
 
     ## ------ Examples demonstrating how to use argparse subcommands -----
 
-    sport_item_strs = ('Bat', 'Basket', 'Basketball', 'Football', 'Space Ball')
-
     # create the top-level parser for the base command
-    base_parser = cmd2.Cmd2ArgumentParser()
-    base_subparsers = base_parser.add_subparsers(title='subcommands', help='subcommand help')
+    calculate_parser = cmd2.Cmd2ArgumentParser(description="Perform simple mathematical calculations.")
+    calculate_subparsers = calculate_parser.add_subparsers(title='operation', help='Available operations')
 
-    # create the parser for the "foo" subcommand
-    parser_foo = base_subparsers.add_parser('foo', help='foo help')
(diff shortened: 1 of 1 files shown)


examples/argparse_example.py
```diff
@@ -106,98 +106,45 @@ def do_print_unknown(self, args: argparse.Namespace, unknown: list[str]) -> None
...
-        if func is not None:
-            # Call whatever subcommand function was selected
-            func(self, args)
-        else:
-            # No subcommand was provided, so call help
-            self.do_help('base')
-
-    @cmd2.with_argparser(base2_parser)
+    calculate_parser = cmd2.Cmd2ArgumentParser(description="Perform simple mathematical calculations.")
+    calculate_subparsers = calculate_parser.add_subparsers(title='operation', help='Available operations')
```
kmvanbrunt: Set `required=True` in the `add_subparsers()` call since a subcommand is required for `calculate`. This will simplify `do_calculate()`. See below.

examples/argparse_example.py
```diff
@@ -106,98 +106,45 @@ def do_print_unknown(self, args: argparse.Namespace, unknown: list[str]) -> None
...
     @cmd2.with_category(ARGPARSE_SUBCOMMANDS)
-    def do_alternate(self, args: argparse.Namespace) -> None:
-        """Alternate command help."""
-        func = getattr(args, 'func', None)
-        if func is not None:
-            # Call whatever subcommand function was selected
-            func(self, args)
+    def do_calculate(self, ns: argparse.Namespace) -> None:
+        """Calculate a simple mathematical operation on two integers."""
+        handler = ns.cmd2_handler.get()
```
kmvanbrunt: By adding `required=True` to `add_subparsers()`, this function can be simplified to:

```Python
def do_calculate(self, args: argparse.Namespace) -> None:
   """Calculate a simple mathematical operation on two integers."""
    handler = args.cmd2_handler.get()
    handler(args)
```

kmvanbrunt (approved):

PR #1501, opened 2025-09-06
Monkey patched two Rich functions.

kmvanbrunt:
Removed `rich_utils.string_to_rich_text()` in favor of monkey patching Rich's `Text.from_ansi()`. Now a dev isn't required to call a cmd2-specific function to get the bug fix.
    
Monkey patched `Rich's Segment.apply_style()` to fix an issue where background colors incorrectly carry over onto the following line when printing with soft_wrap enabled.


tleonhardt (approved):
I like this elegant solution.

PR #1504, opened 2025-09-07
Optimize performance of running command finalization hooks

tleonhardt:
Optimize performance of the `_run_cmdfinalization_hooks` method by replacing a subprocess call to `stty sane` with a call to `termios.tcsetattr()`. For optimial performance, the initial termios settings are cached when instantiating the `cmd2.Cmd` class and are restored in the `_run_cmdfinalization_hooks` method.

The motivation behind this was a user report of slowness in Discussion #1503. 

I also temporarily tweaked where we we report our elapsed timing when the `timiing` settable is True to include the full command lifecycle timing and did measurements before and after.

I did some testing on a system with the following setup:

- Latest `cmd2` from main branch on GitHub
- Python 3.14.0rc2
- macOS Sequoia 15.6.1 on an older Intel core i9 Macbook pro
- term2 3.5.14

in this testing I modified cmd's built-in timing measurement to capture the full command lifecycle timing. With the code as-is on the `main` branch, the full lifecycle for an empty command on this older computer was about 0.007 seconds (7 thousands of a second). This is not something I would even remotely call "slow" given typical human reaction times.

I then re-ran the same tests using the code in this PR. This did substantially speed things up to the point where an empty command ran with an elapsed time of about 0.00011 seconds (about 64 times faster).

I think we should do some manual evaluation with the code here to verify it is fixing the same things that `stty sane` was fixing. Even though the performance difference should be below what would normally be perceptible for the system I tested on, user reports of slowness in some circumstances suggest this may be a beneficial and noticeable change overall.

TODO:

- [x] Manual testing to verify new code has same beneficial effects as running `stty sane`
- [x] Improve test coverage

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,8 @@ time reading the [rich documentation](https://rich.readthedocs.io/).
         - [argparse_example.py](https://github.com/python-cmd2/cmd2/blob/main/examples/argparse_example.py)
         - [command_sets.py](https://github.com/python-cmd2/cmd2/blob/main/examples/command_sets.py)
         - [getting_started.py](https://github.com/python-cmd2/cmd2/blob/main/examples/getting_started.py)
+    - Optimized performance of terminal fixup during command finalization by replacing `stty sane`
+      with `termios.tcsetattr`
 
 - Bug Fixes
     - Fixed a redirection bug where `cmd2` could unintentionally overwrite an application's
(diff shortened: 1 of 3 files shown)


tleonhardt:
I did some manual testing and verified that the termios approach of restoring terminal sanity works just as well for as `ssty sane` for fixing common terminal corruption issues like broken character echo, broken line editing, etc.

I also verified that neither approach works against some really pathological situations such as changing the character set being used by the terminal.

tleonhardt:
Score one for Gemini 2.5 Pro for one-shotting the 100% test coverage for the new code.

kmvanbrunt (approved):

PR #1528, opened 2025-10-04
Make backwards compatibility slightly easier

tleonhardt:
Revert `CompletionItem.description` field to original name, but expand type from `str` to `str | Sequence[Any]`


tleonhardt:
Based on conversations with @kmvanbrunt I think I'm going to close this PR without merging.

PR #1539, opened 2025-12-07
Remove dependency on Python's built-in cmd module

tleonhardt:
The `cmd2.Cmd` class no longer inherits from `cmd.Cmd` and the `cmd2` module no longer depends on Python's built-in `cmd` module.

To make this happen we had to do the following:

- Add an extra parameter to `cmd2.Cmd.__init__` and several extra initializations within it
- Add implementation for the `completedefault` and `completenames` methods
- Eliminate some redundant casts to fix `mypy` warnings

### DIFF
--- a/tests/test_completion.py
+++ b/tests/test_completion.py
@@ -219,14 +219,6 @@ def cmd2_app():
     return CompletionsExample()
 
 
-def test_cmd2_command_completion_single(cmd2_app) -> None:
-    text = 'he'
-    line = text
-    endidx = len(line)
-    begidx = endidx - len(text)
-    assert cmd2_app.basic_complete(text, line, begidx, endidx, cmd2_app.get_all_commands()) == ['help']
-
-
 def test_complete_command_single(cmd2_app) -> None:
     text = 'he'
     line = text
@@ -322,7 +314,10 @@ def test_cmd2_command_completion_multiple(cmd2_app) -> None:
     line = text
     endidx = len(line)
     begidx = endidx - len(text)
-    assert cmd2_app.basic_complete(text, line, begidx, endidx, cmd2_app.get_all_commands()) == ['help', 'history']
+
+    first_match = complete_tester(text, line, begidx, endidx, cmd2_app)
+    assert first_match is not None
+    assert cmd2_app.completion_matches == ['help', 'history']
 
 
 def test_cmd2_command_completion_nomatch(cmd2_app) -> None:
@@ -330,7 +325,10 @@ def test_cmd2_command_completion_nomatch(cmd2_app) -> None:
     line = text
     endidx = len(line)
     begidx = endidx - len(text)
-    assert cmd2_app.basic_complete(text, line, begidx, endidx, cmd2_app.get_all_commands()) == []
+
+    first_match = complete_tester(text, line, begidx, endidx, cmd2_app)
+    assert first_match is None
+    assert cmd2_app.completion_matches == []
 
 
 def test_cmd2_help_completion_single(cmd2_app) -> None:


tleonhardt:
It's complaining about code coverage and the reason I think is we aren't covering the line for the case where `stdout` is not `None` in the `cmd2.Cmd.__init__`.

cmd2/cmd2.py
```diff
@@ -304,6 +303,9 @@ class Cmd(cmd.Cmd):
     # List for storing transcript test file names
     testfiles: ClassVar[list[str]] = []
 
+    DEFAULT_PROMPT = '(Cmd) '
+    IDENTCHARS = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789_'
```
kmvanbrunt: `cmd2` has not used `identchars` since 0.9.0.

docs/migrating/why.md
```diff
@@ -22,6 +22,11 @@ of [cmd][cmd] will add many features to an application without any further modif
 to `cmd2` will also open many additional doors for making it possible for developers to provide a
 top-notch interactive command-line experience for their users.
 
+!!! warning
+
+    As of version 4.0.0, `cmd2` does not have an actual dependency on `cmd`. It is API compatible, but
```
kmvanbrunt: I wouldn't say we are API compatible since we have documented [incompatibilities](https://github.com/python-cmd2/cmd2/blob/main/docs/migrating/incompatibilities.md).

We also should rewrite this comment found in the cmd2.py header.

> To use, simply import cmd2.Cmd instead of cmd.Cmd; use precisely as though you
were using the standard library's cmd, while enjoying the extra features.

cmd2/cmd2.py
```diff
@@ -326,6 +328,7 @@ def __init__(
         auto_load_commands: bool = False,
         allow_clipboard: bool = True,
         suggest_similar_command: bool = False,
+        intro: str = '',
```
kmvanbrunt: We might consider changing `intro` to a Rich `RenderableType`. It's defined in `rich.console`.

```Python
# A type that may be rendered by Console.
RenderableType = Union[ConsoleRenderable, RichCast, str]
```

If you make this change, default `intro` to `None`.

cmd2/cmd2.py
```diff
@@ -3293,6 +3311,19 @@ def default(self, statement: Statement) -> bool | None:  # type: ignore[override
...
 
+    def completedefault(self, *_ignored: list[str]) -> list[str]:
+        """Call to complete an input line when no command-specific complete_*() method is available.
+
+        By default, it returns an empty list.
+
+        """
+        return []
+
+    def completenames(self, text: str, *_ignored: list[str]) -> list[str]:
```
kmvanbrunt: cmd2 does not use this method. I suggest deleting it.

cmd2/cmd2.py
```diff
@@ -3293,6 +3311,19 @@ def default(self, statement: Statement) -> bool | None:  # type: ignore[override
         self.perror(err_msg, style=None)
         return None
 
+    def completedefault(self, *_ignored: list[str]) -> list[str]:
+        """Call to complete an input line when no command-specific complete_*() method is available.
```
kmvanbrunt, line 3315: Add something like "for non-argparse-based commands" to the end of this docstring.
Also, this really isn't a function that a user calls. Maybe change the first word to "Called".

kmvanbrunt (changes requested):

cmd2/cmd2.py
tleonhardt: Will do.

cmd2/cmd2.py
tleonhardt: Apparently both you and I added a few unit tests in the `test_completeion.py` file about 8 years ago which use `completenames` ;-). Who knew!

tleonhardt: @kmvanbrunt Here is one example of where we used the now otherwise unused `completenames` method in a unit test: https://github.com/python-cmd2/cmd2/blob/main/tests/test_completion.py#L227

What would be the most suitable replacement to use in these tests?

Issue #764, opened 2019-08-22
cmd2 usage within asyncio.get_event_loop().run_until_complete


adriansev:
Hi! I would like to use cmd2 to create a shell for interactivity with a websocket based service.   
For persistence a websocket must be created within asyncio.get_event_loop().run_until_complete()
see [websockets](https://websockets.readthedocs.io/en/stable/intro.html)
How can i adapt cmd2 to such usage? (note that i am interested only in a shell, not sending commands, as that it is covered) for reference my application is [here](https://github.com/adriansev/jalien_py/blob/master/xjalienfs/alien.py)

tleonhardt:
Hi @adriansev, we have info in the `cmd2` documentation about integrating with [Alternate Event Loops](https://cmd2.readthedocs.io/en/latest/examples/alternate_event_loops.html)

Please see if that answers your question about how you can adapt `cmd2` for use with `ayncio`.

adriansev:
yes, i seen the documentation but it is not clear to me what part from example i am supposed to run within the asyncio.get_event_loop().run_until_complete(). the websocket and related operations are supposed to run within the loop, and my goal is to have a shell like interface where commands are sent to websocket server, and i get back the messages from that server.
moreover : 
```
python3
Python 3.7.4 (default, Jul  9 2019, 16:32:37)
[GCC 9.1.1 20190503 (Red Hat 9.1.1-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import cmd2
>>> app = Cmd2EventBased()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
NameError: name 'Cmd2EventBased' is not defined
>>>
```

adriansev:
yes, i am just trying to call async functions from cmd2. the "commands" are websocket send messages (async) that then await for recv answer to the server. Is there any way to accommodate this? 
The alternative is to create a global variable (eventually in a namespace to not clash with anything else) were an send_command function would deposit the awaited answer.

xNinjaKittyx:
The way I do it is by creating a cmd2 object as you normally would according to documentation, then within the `do_*` commands, you call `loop.run_until_complete()` on the functions you want to await.

Since if you have a lot of async stuff, it's a little annoying to write a separate async function for every `do_*` function, so I also had my own wrapper that would allow me to use `async def do_*` functions that be wrapped with `run_until_complete()` at runtime, so that cmd2 would work with async functions.

[...]

PR #1554, opened 2026-01-06
docs: async call example

asokolsky:
demonstrate calling an async function from a cmd2 command.

see https://github.com/python-cmd2/cmd2/issues/764


tleonhardt (approved):
Thanks for the PR Alex.  I'll merge it in as-is and make a minor tweak to handle a ruff warning.

PR #1571, opened 2026-02-05
Fixed incompatibilities with Python 3.14.3.

kmvanbrunt:


tleonhardt (approved):

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 3.1.4 (TBD)
```
tleonhardt: Given all of the changes in cmd2's custom argparse argument and help formatting, I think it might be a good idea to make this the 3.2.0 release and add a note about those changes.

While its not a major backwards incompatibility it will likely break some people's tests if they are strict.

tleonhardt (approved):

tleonhardt (approved):

PR #1575, opened 2026-02-17
Refactor completion API

kmvanbrunt:
Overhauled the completion API to use encapsulation and better reflect how `prompt-toolkit` works.

### DIFF
--- a/cmd2/argparse_completer.py
+++ b/cmd2/argparse_completer.py
@@ -143,7 +143,7 @@ class _NoResultsError(CompletionError):
     def __init__(self, parser: argparse.ArgumentParser, arg_action: argparse.Action) -> None:
         """CompletionError which occurs when there are no results.
 
-        If hinting is allowed, then its message will be a hint about the argument being completed.
+        If hinting is allowed on this argument, then its hint text will display.
 
         :param parser: ArgumentParser instance which owns the action being completed
         :param arg_action: action being completed.
@@ -481,7 +481,6 @@ class ArgparseCompleter:
 
             # If we have results, then return them
             if completions:
-                # Don't overwrite an existing hint
                 if (
                     not completions.completion_hint
                     and not isinstance(pos_arg_state.action, argparse._SubParsersAction)
--- a/cmd2/argparse_completer.py
+++ b/cmd2/argparse_completer.py
@@ -554,8 +554,11 @@ class ArgparseCompleter:
                     match_against.append(flag)
 
         # Build a dictionary linking actions with their matched flag names
-        matched_flags = self._cmd2_app.basic_complete(text, line, begidx, endidx, match_against)
         matched_actions: dict[argparse.Action, list[str]] = {}
+
+        # Keep flags sorted in the order provided by argparse so our completion
+        # suggestions display the same as argparse help text.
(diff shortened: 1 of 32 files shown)


cmd2/cmd2.py
```diff
@@ -2426,191 +2389,150 @@ def _perform_completion(
...
-                    # Figure out what kind of quote to add and save it as the unclosed_quote
-                    completion_token_quote = "'" if any('"' in match for match in self.completion_matches) else '"'
-
-                    self.completion_matches = [completion_token_quote + match for match in self.completion_matches]
-
-            # Check if we need to remove text from the beginning of tab completions
-            elif text_to_remove:
-                self.completion_matches = [match.replace(text_to_remove, '', 1) for match in self.completion_matches]
+        # Check if we need to remove text from the beginning of completions
+        elif text_to_remove:
```
tleonhardt, line 2412: If I read it correctly, the code coverage report shows this `elif text_to_remove` branch as being uncovered. Recommend adding a test for it.

cmd2/argparse_completer.py
```diff
@@ -95,13 +100,13 @@ class _ArgumentState:
 
     def __init__(self, arg_action: argparse.Action) -> None:
         self.action = arg_action
-        self.min: int | str
-        self.max: float | int | str
+        self.min: int
```
tleonhardt, line 103: Why is `self.min` only an `int` when `self.max` is `float | int`? Should they be consistent? Or is it only allowing it max to be an `int` or `INFINITY`?

cmd2/command_definition.py
```diff
@@ -13,16 +14,14 @@
 from .exceptions import (
     CommandSetRegistrationError,
 )
```
tleonhardt: Prefer not using parentheses for single import

cmd2/cmd2.py
```diff
@@ -4473,27 +4404,27 @@ def _build_base_set_parser(cls) -> Cmd2ArgumentParser:
...
             help='parameter to set or view',
-            choices_provider=cls._get_settable_completion_items,
-            descriptive_headers=["Value", "Description"],
+            choices_provider=cls._get_settable_choices,
+            table_header=["Value", "Description"],
         )
 
         return base_set_parser
 
     def complete_set_value(
```
tleonhardt, line 4413: I'm not sure how or where the right place to do it is, but I think we should fix how the `set` completion works in `prompt-toolkit`.  The printed table above looks great.

But if I use the arrow keys to select one of the settables, it shows me the current value.  Instead I'd like to see some combination of the current value and help text.

tleonhardt (dismissed):

cmd2/argparse_completer.py
kmvanbrunt: Yes, it's because of INFINITY.

tleonhardt: Is there a way to make that part of the type more specific so it isn't `float` but is specific to `INFINITY`?

cmd2/command_definition.py
kmvanbrunt: I just updated the imports.

cmd2/cmd2.py
kmvanbrunt: Added test

PR #1582, opened 2026-02-21
Fix async_alert to respect allow_style setting for 3.x branch

### LINKED ISSUE #1581: async_update_prompt and async_alert do not respect the allow_style settable
The `async_update_prompt` and `async_alert` methods within `cmd2.Cmd` do not respect the `allow_style` settable.

If `allow_style` is set to `NEVER`, then ANSI escape codes should be stripped before outputting using these methods as well.

tleonhardt:
This PR fixes the `async_alert` method and other methods which use it such as `async_update_prompt` so that it respects the `allow_style` settable and strips out ANSI escape codes when `allow_style` is `NEVER`.

This closes #1581 

We will need to make equivalent changes on the `main` branch after this merges to the `3.x` branch but it won't be from a simple cherry-pick.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+## 3.2.1 (February TBD, 2026)
+
+- Bug Fixes
+    - The `async_alert` and `async_prompt_update` methods of `cmd2.Cmd` now respect the current
+      value of the `allow_style` settable
+        - If `allow_style` is `NEVER`, all ANSI escape codes will be stripped to ensure plain text
+          output
+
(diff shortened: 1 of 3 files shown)


kmvanbrunt (approved):
Looks good, but don't cherry pick this into main since I'm rewriting the async_alert code for 4.x.

tleonhardt:
@kmvanbrunt I'll merge this into the `3.x` branch and won't cherry-pick it to `main` since you are in the middle of re-writing the `async_alert` code for `4.x`.  Please make sure the new code respects that `allow_style` setting.

PR #1585, opened 2026-02-24
Refactor input

kmvanbrunt:
Moved logic specific to the main command line from `Cmd.read_input()` to `Cmd._read_command_line()`.
    
No longer setting `Cmd.session.input` to a pipe since that causes warnings from `prompt-toolkit`.
For non-interactive sessions (`self.stdin` is a pipe), we use `self.stdin.readline()`.


cmd2/cmd2.py
```diff
@@ -3185,36 +3192,86 @@ def completedefault(self, *_ignored: Sequence[str]) -> Completions:
...
         choices_provider: ChoicesProviderUnbound | None = None,
         completer: CompleterUnbound | None = None,
         parser: argparse.ArgumentParser | None = None,
     ) -> str:
-        """Read input from appropriate stdin value.
-
-        Also supports completion and up-arrow history while input is being entered.
+        """Read a line of input with optional completion and history.
 
-        :param prompt: prompt to display to user
```
tleonhardt, line 3204: I don't think the documentation for either the `prompt` or `history` parameter should be deleted here

cmd2/cmd2.py
```diff
@@ -655,6 +609,63 @@ def _(event: Any) -> None:  # pragma: no cover
...
+            "history": self.history_adapter,
+            "key_bindings": key_bindings,
+            "lexer": self.lexer,
+            "rprompt": self.get_rprompt,
+        }
+
+        if self.stdin.isatty():
+            try:
+                if self.stdin != sys.stdin:
+                    kwargs["input"] = create_input(stdin=self.stdin)
```
tleonhardt, line 655: Recommend adding unit test for this case where `self.stdin != sys.stdin`

cmd2/cmd2.py
```diff
@@ -5248,11 +5215,11 @@ def _alert() -> None:
 
             if hasattr(self, 'session'):
                 # Invalidate to force prompt update
-                self.session.app.invalidate()
+                get_app().invalidate()
```
tleonhardt, line 5242: Recommend adding unit test for this case where `session` exists within `self`

tleonhardt (changes requested):

cmd2/cmd2.py
kmvanbrunt: This function is going to be rewritten.

cmd2/cmd2.py
kmvanbrunt: Fixed

cmd2/cmd2.py
kmvanbrunt: Fixed

cmd2/cmd2.py
```diff
@@ -3185,173 +3198,160 @@ def completedefault(self, *_ignored: Sequence[str]) -> Completions:
...
+        if self.interactive_pipe:
+            self.poutput(prompt_str, end='')
+            self.stdout.flush()
+
+        # Wait for the next line of input
+        line = self.stdin.readline()
+
+        # If the stream is empty, we've reached the end of the input.
+        if not line:
+            raise EOFError
```
tleonhardt, line 3233: Do we not have a specific message because every internal call to this is wrapped in a try/except where we catch this and handle appropriately?

cmd2/cmd2.py
```diff
@@ -5248,11 +5245,11 @@ def _alert() -> None:
 
             if hasattr(self, 'session'):
                 # Invalidate to force prompt update
-                self.session.app.invalidate()
+                get_app().invalidate()
```
tleonhardt, line 5242: Should this be wrapped in a try/except Exception with a pass? I think it could raise an exception if the prompt exited concurrently

cmd2/cmd2.py
```diff
@@ -5248,11 +5245,11 @@ def _alert() -> None:
...
 
         # Schedule the alert to run on the main thread's event loop
         try:
-            self.session.app.loop.call_soon_threadsafe(_alert)  # type: ignore[union-attr]
+            get_app().loop.call_soon_threadsafe(_alert)  # type: ignore[union-attr]
         except AttributeError:
```
tleonhardt, line 5247: Are there other types of exceptions that could happen here other than `AttributeError`?

tests/test_cmd2.py
```diff
@@ -3609,68 +3577,6 @@ def test_path_complete_users_windows(monkeypatch, base_app):
     assert expected in matches
 
 
-def test_async_alert_success(base_app):
```
tleonhardt, line 3612: Why did these async alert tests get deleted?  I can see them getting modified, but it feels like they should be here.

CHANGELOG.md
```diff
@@ -56,6 +56,7 @@ prompt is displayed.
       `Statement.redirect_to`.
     - Changed `StatementParser.parse_command_only()` to return a `PartialStatement` object.
     - Renamed `Macro.arg_list` to `Macro.args`.
+    - Removed `terminal_utils.py` since `prompt-toolkit` provides this functionality.
```
tleonhardt, line 59: We should also update the `upgrades.md` docuemntation for 4.x with a section on deleted modules and both `rl_utils` and `terminal_utils` should be called out

cmd2/cmd2.py
kmvanbrunt: This function is being rewritten on the next PR.

cmd2/cmd2.py
kmvanbrunt: This function is being rewritten on the next PR.

tests/test_cmd2.py
kmvanbrunt: I am rewriting the async stuff on the next PR.

cmd2/cmd2.py
```diff
@@ -3373,6 +3365,8 @@ def _cmdloop(self) -> None:
                 except KeyboardInterrupt:
                     self.poutput('^C')
                     line = ''
+                except EOFError:
+                    line = "_eof"
```
tleonhardt, line 3369: I think we might need a new test to cover this case

kmvanbrunt: Fixed

CHANGELOG.md
kmvanbrunt: Done

tleonhardt (approved):

PR #1586, opened 2026-02-25
Refactored async alerts.

kmvanbrunt:
Replaced `async_alert()` and `async_update_prompt()` with a single function called `add_alert()`. This new function is thread-safe and does not require you to acquire a mutex before calling it like the previous functions did.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,9 @@ prompt is displayed.
     - Changed `StatementParser.parse_command_only()` to return a `PartialStatement` object.
     - Renamed `Macro.arg_list` to `Macro.args`.
     - Removed `terminal_utils.py` since `prompt-toolkit` provides this functionality.
+    - Replaced `async_alert()` and `async_update_prompt()` with a single function called
+      `add_alert()`. This new function is thread-safe and does not require you to acquire a mutex
+      before calling it like the previous functions did.
 - Enhancements
     - New `cmd2.Cmd` parameters
         - **auto_suggest**: (boolean) if `True`, provide fish shell style auto-suggestions. These
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,8 +69,6 @@ prompt is displayed.
           displaying realtime status information while the prompt is displayed, see the
           `cmd2.Cmd2.get_bottom_toolbar` method that can be overridden as well as the updated
           `getting_started.py` example
-    - Added `cmd2.Cmd._in_prompt` flag that is set to `True` when the prompt is displayed and the
-      application is waiting for user input
     - New `cmd2.Cmd` methods
         - **get_bottom_toolbar**: populates bottom toolbar if `bottom_toolbar` is `True`
         - **get_rprompt**: override to populate right prompt
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -286,7 +286,11 @@ class AsyncAlert:
     """Contents of an asynchonous alert which display while user is at prompt.
 
     :param msg: an optional message to be printed above the prompt.
-    :param prompt: an optional string to dynamically replace the current prompt.
+    :param prompt: an optional string to dynamically replace the active prompt.
+
+    :ivar timestamp: monotonic creation time of the alert. If an alert was created
+                     before the active prompt started, the prompt update is ignored
+                     to avoid a stale display but the msg will still be displayed.
     """
 
     msg: str | None = None
@@ -3367,7 +3371,7 @@ class Cmd:
                 # Get the next alert while still holding the condition lock.
                 alert = self._alert_queue.get()
 
-            # Only apply prompt changes generated after the current prompt started.
+            # Only apply prompt changes generated after the active prompt started.
             prompt_updated = False
             if (alert.prompt is not None and
                 alert.prompt != self.prompt and
@@ -3376,7 +3380,7 @@ class Cmd:
                 prompt_updated = True
 
             if alert.msg:
-                # Print the message above the current prompt.
+                # Print the message above the active prompt.
                 with patch_stdout():
                     print_formatted_text(pt_filter_style(alert.msg))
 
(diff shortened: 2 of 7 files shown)


cmd2/cmd2.py
```diff
@@ -273,6 +281,19 @@ def remove(self, command_method: CommandFunc) -> None:
             del self._parsers[full_method_name]
 
 
+@dataclass(kw_only=True)
+class AsyncAlert:
+    """Contents of an asynchonous alert which display while user is at prompt.
+
+    :param msg: an optional message to be printed above the prompt.
+    :param prompt: an optional string to dynamically replace the current prompt.
```
tleonhardt, line 291: Recommend documenting `timestamp` parameter as well in the docstring. I think it is particularly important to document that prompt updates older than the current prompt will be dropped without warning but older alert messages will still be shown (assuming I read the code correctly).

tleonhardt (changes requested):

cmd2/cmd2.py
```diff
@@ -370,7 +391,7 @@ def __init__(
         self._initialize_plugin_system()
 
         # Configure a few defaults
-        self.prompt = Cmd.DEFAULT_PROMPT
+        self.prompt: str = Cmd.DEFAULT_PROMPT
```
tleonhardt, line 400: Is `self.prompt` always a `str` and never anything like a `rich.Text`?

cmd2/cmd2.py
```diff
@@ -5207,66 +5304,25 @@ def do__relative_run_script(self, args: argparse.Namespace) -> bool | None:
         # self.last_result will be set by do_run_script()
         return self.do_run_script(su.quote(relative_path))
 
-    def async_alert(self, alert_msg: str, new_prompt: str | None = None) -> None:
-        """Display an important message to the user while they are at a command line prompt.
-
-        To the user it appears as if an alert message is printed above the prompt and their
-        current input text and cursor location is left alone.
+    def add_alert(self, *, msg: str | None = None, prompt: str | None = None) -> None:
```
tleonhardt, line 5336: Can either `msg` or `prompt` be something other than a `str` or `None` such as a `rich.Text` or `pt.ANSI`?

docs/features/prompt.md
```diff
@@ -28,14 +28,11 @@ for an example of dynamically updating the prompt.
...
-with the command line. This means the feedback is provided to the user when they are still entering
-text at the prompt. To use this functionality, the application must be running in a terminal that
-supports [VT100](https://en.wikipedia.org/wiki/VT100) control characters. Linux, Mac, and Windows 10
-and greater all support these.
-
-- [cmd2.Cmd.async_alert][]
-- [cmd2.Cmd.async_update_prompt][]
+`cmd2` provides a function to deliver asynchronous feedback to the user without interfering with the
+command line. This means the feedback is provided to the user when they are still entering text at
+the prompt.
```
tleonhardt: I feel like we should elaborate a bit more here and explain a little more in depth about how this works.

CHANGELOG.md
tleonhardt: Need to remove this line about `_in_prompt` now that you got rid of that

tleonhardt (changes requested):

cmd2/cmd2.py
kmvanbrunt: Done

cmd2/cmd2.py
kmvanbrunt: We don't currently support rich.Text prompts. We should consider that for a future PR.

cmd2/cmd2.py
kmvanbrunt: Not currently.

docs/features/prompt.md
kmvanbrunt: Done

docs/features/prompt.md
```diff
@@ -28,18 +28,27 @@ for an example of dynamically updating the prompt.
...
+
+### Asynchronous Feedback Mechanisms
+
+Alerts can interact with the CLI in two ways:
+
+1. **Message Printing**: It can print a message directly above the current prompt line.
+1. **Prompt Updates**: It can dynamically replace the text of the active prompt to reflect changing
+   state.
+
+> **Note**: To ensure the user interface remains accurate, a prompt update is ignored if the alert
```
tleonhardt: `Zensical` supports the same [Admonitions](https://zensical.org/docs/authoring/admonitions/) that `MkDocs with the Material` theme does.

You can replace this with an official **Note** admonition like so:

```Markdown
!!! note
    
    <Text goes here>
```

tleonhardt (dismissed):

docs/features/prompt.md
kmvanbrunt: I like that better and made the change.

tleonhardt (dismissed):

cmd2/cmd2.py
```diff
@@ -3321,6 +3358,60 @@ def read_input(
...
+                        and alert.timestamp > self._alert_prompt_timestamp
+                    ):
+                        latest_prompt = alert.prompt
+                        self._alert_prompt_timestamp = alert.timestamp
+                        break
+
+                # Clear the alerts
+                self._alert_queue.clear()
+
+                if alert_text:
```
tleonhardt, line 3395: Since in all cases in this if/elif, `self.prompt` gets set to `latest_prompt` if `latest` prompt is not `None`, I recommend refactoring and simplifying this code to look something like the following:

```Python
if latest_prompt is not None:
    # Update prompt so patch_stdout() or get_app().invalidate() can redraw it
    self.prompt = latest_prompt

if alert_text:
    # Print the alert messages above the prompt
    with patch_stdout():
        print_formatted_text(pt_filter_style(alert_text))

elif latest_prompt is not None and not self._at_continuation_prompt:
    # Refresh UI immediately to show the new prompt
    get_app().invalidate()
```

Perhaps I'm missing something?  But this seems simpler.

kmvanbrunt: That code allows for the `patch_stdout` print to overwrite a continuation prompt.

tleonhardt (dismissed):

PR #1587, opened 2026-02-25
Fixed custom types and moved common ones into types.py.

kmvanbrunt:


cmd2/types.py
```diff
@@ -0,0 +1,66 @@
...
+# Types used in choices_providers and completers
+##################################################
+
+# Represents the parsed tokens from argparse during completion
+ArgTokens: TypeAlias = Mapping[str, Sequence[str]]
+
+# Represents a type that can be matched against when completing.
+# Strings are matched directly while CompletionItems are matched
+# against their 'text' member.
+Matchable: TypeAlias = Union[str, "CompletionItem"]
```
tleonhardt: I am cool with types being used in multiple places being defined in a single `types.py` file.  

But `Matchable` only appears to be used in `cmd2.py`.  I think it might be better to just define it there.  What do you think?

cmd2/types.py
```diff
@@ -0,0 +1,66 @@
...
+    Callable[[CmdOrSet, str, str, int, int], "Completions"]
+    |
+    # Context-aware: (self, text, line, begidx, endidx, arg_tokens) -> Completions
+    Callable[[CmdOrSet, str, str, int, int, ArgTokens], "Completions"]
+)
+
+# A bound completer used internally by cmd2 for basic completion logic.
+# The 'self' argument is already tied to an instance and is omitted.
+# Format: (text, line, begidx, endidx) -> Completions
+CompleterBound: TypeAlias = Callable[[str, str, int, int], "Completions"]
```
tleonhardt, line 61: Similarly, `CompleterBound` is only used in `cmd2.py` and might be better defined there?

tleonhardt (dismissed):

cmd2/argparse_completer.py
```diff
@@ -3,6 +3,8 @@
 See the header of argparse_custom.py for instructions on how to use these features.
 """
 
+from __future__ import annotations
```
tleonhardt: As a library (as opposed to an application), I strongly feel that we should **NOT** add `from __future__ import annotations` to any of our code except possibly for test or example code.

The main danger of putting `from __future__ import annotations` everywhere is that it changes the semantics of type hint evaluation, which can break tools and libraries that rely on runtime access to evaluated type annotations, such as Pydantic, SQLAlchemy, or functools.singledispatch.

Adding this could break application code in very subtle and hard to debug ways. I strongly feel we should avoid it. Moreover, it is deprecated and will be removed in future Python versions. PEP 649/749 introduces better deferred evaluation which was adopted starting with Python 3.14. But the way that is handled isn't fully compatible with the way `from __future__ import annotations` handles it.

tleonhardt (changes requested):

cmd2/types.py
kmvanbrunt: I removed this type since it only applies to one function (basic_completer) and not normal completers.

cmd2/types.py
kmvanbrunt: I prefer keeping CompleterBound in types.h. cmd2.py is pretty big and I can see that type being useful for other devs.

tleonhardt (approved):

PR #1589, opened 2026-02-27
Switch to using prompt-toolkit's multiline filtering.

kmvanbrunt:
Only add completed multiline commands to prompt-toolkit's history.

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -69,7 +69,10 @@ from typing import (
 )
 
 import rich.box
-from prompt_toolkit import print_formatted_text
+from prompt_toolkit import (
+    filters,
+    print_formatted_text,
+)
 from prompt_toolkit.application import get_app
 from prompt_toolkit.auto_suggest import AutoSuggestFromHistory
 from prompt_toolkit.completion import Completer, DummyCompleter
@@ -136,6 +139,8 @@ from .exceptions import (
     CompletionError,
     EmbeddedConsoleExit,
     EmptyStatement,
+    IncompleteStatement,
+    MacroError,
     PassThroughException,
     RedirectionError,
     SkipPostcommandHooks,
@@ -200,6 +205,7 @@ from .utils import (
 if TYPE_CHECKING:  # pragma: no cover
     StaticArgParseBuilder = staticmethod[[], argparse.ArgumentParser]
     ClassArgParseBuilder = classmethod['Cmd' | CommandSet, [], argparse.ArgumentParser]
+    from prompt_toolkit.buffer import Buffer
 else:
     StaticArgParseBuilder = staticmethod
     ClassArgParseBuilder = classmethod
@@ -510,12 +516,6 @@ class Cmd:
         # Used to keep track of whether we are redirecting or piping output
         self._redirecting = False
 
-        # Used to keep track of whether a continuation prompt is being displayed
-        self._at_continuation_prompt = False
-
-        # The multiline command currently being typed which is used to complete multiline commands.
-        self._multiline_in_progress = ''
-
         # Characters used to draw a horizontal rule. Should not be blank.
         self.ruler = "─"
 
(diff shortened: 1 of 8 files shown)


cmd2/cmd2.py
```diff
@@ -672,7 +672,7 @@ def _(event: Any) -> None:  # pragma: no cover
             "complete_in_thread": True,
             "complete_while_typing": False,
             "completer": Cmd2Completer(self),
-            "history": Cmd2History(self),
+            "history": Cmd2History(item.raw for item in self.history),
```
tleonhardt, line 707: Interesting so we are now passing in a generator expression of raw history items instead of the cmd2.Cmd instance.  This is probably better.

cmd2/cmd2.py
```diff
@@ -2902,6 +2902,9 @@ def _complete_statement(self, line: str) -> Statement:
         if not statement.command:
             raise EmptyStatement
 
+        # Add the complete command to prompt-toolkit's history.
+        cast(Cmd2History, self.main_session.history).add_command(statement.raw)
```
tleonhardt: So we are adding to Prompt Toolkit (up-arrow) history at a different time than cmd2 history command (potentially persistent) history.  I guess that is ok.

cmd2/pt_utils.py
```diff
@@ -152,41 +152,46 @@ def get_completions(self, document: Document, _complete_event: object) -> Iterab
...
-        However, PromptSession needs to know about it for the *current* session history navigation.
-        If we don't store it here, UP arrow might not work for the just entered command
-        unless cmd2 re-initializes the session or history object.
+    def add_command(self, string: str) -> None:
+        """Manually add a finalized command string to the UI history stack.
 
-        This method is intentionally empty.
+        Ensures consecutive duplicates are not stored.
         """
+        if string and (not self._loaded_strings or self._loaded_strings[0] != string):
```
tleonhardt, line 176: Does prompt-toolkit's history append strings at element 0?

cmd2/cmd2.py
kmvanbrunt: We've always added it earlier because `readline` and `prompt-toolkit` add history the moment you hit enter.

cmd2/pt_utils.py
kmvanbrunt: Yes, because `self._loaded_strings` is ordered newest to oldest.

cmd2/cmd2.py
kmvanbrunt: Yep, there's no reason to pass the whole `Cmd` object.

tleonhardt: Nice change.  I should have thought of that ;-)

cmd2/cmd2.py
kmvanbrunt: Plus this is meant to be the history of what the user typed. By the time we add to `cmd2's` history in `onecmd()`, the `Statement` has passed through multiple hooks which could have changed it.

tleonhardt: I see several side effects of doing this here in `_complete_statement` instead of elsewhere. Some of these are arguably desirable, but many are IMO undesirable.

One is that the UI / PT (up-arrow) history will not respect the `exclude_from_history` list of commands to exclude. This is likely desirable.

Other effects I'm concerned that might be **undesirable** are as follows:
- I do not think we will respect the `add_to_history` flag from scripts which might typically have this set to `False` and therefore commands from scripts might "clutter" the interactive sessions history with commands the user never actually ran
- I think we are adding commands to history before we do macro expansion. I haven't tested the impact, but I'm worried we might have a bad bug here where we do something like add both the macro and its expansion to the history.
- The casting feels awkward and like maybe we should add a helper method along the lines of `_add_to_pt_history` to add stuff without awkward casting

tleonhardt: I'm also worried if `self.default_to_shell` is true, we may not be adding the desired thing to history.

tleonhardt: I did some more testing.  I'm happy with the macro and default_to_shell behavior.

I do think we need to fix it so script commands don't end up in the prompt-toolkit history.

kmvanbrunt: The new approach of using prompt-toolkit's multiline filtering fixes all of these.

cmd2/cmd2.py
```diff
@@ -643,6 +643,39 @@ def __init__(
         # the current command being executed
         self.current_command: Statement | None = None
 
+    def _should_continue_multiline(self) -> bool:
```
tleonhardt, line 646: Both local testing and code coverage analysis integrated with our CI/CD show that we have zero coverage of this helper method.

We should add some unit and/or integration tests for it.

tleonhardt (changes requested):

cmd2/cmd2.py
kmvanbrunt: It's been added.

tests/test_cmd2.py
tleonhardt, line 1252: I'm concerned that because of the way `prompt-toolkit` evaluates and resolves input under the new multiline setup that this test might be flaky where the background alerter thread might occasionally be starved of enough runtime to emit its queued alerts before the `cmdloop` exits.

i think we might want to purposely make the quit delayed a little to give the background thread sufficient time to clear the message queue.  

I'm not sure if we should add something to `cmd2` for this in general or if we should do it just in this test.

tleonhardt (changes requested):

tests/test_cmd2.py
tleonhardt: I'm thinking of something like a 0.1 second delay here.  Not something that would annoy users.  But enough to get background threads time to do their thing.

tleonhardt (approved):

tests/test_cmd2.py
kmvanbrunt: Our background alerter thread is properly synced against anything like that. No need to sleep or poll.

Issue #1578, opened 2026-02-19
Rich Console Status Visual Bug


ShiftPoster:
### Description
Using the `Cmd.p*` output methods from while a rich status spinner is running causes the messages to merge on the terminal.

### Environment
- Windows 11
- Windows Terminal Version: 1.23.20211.0
- Python 3.12.2
- cmd2-3.2.0

### Steps to Reproduce
```python
import time
from argparse import Namespace

from cmd2 import Cmd, with_category
from rich.console import Console

TEST_CATAGORY: str = "Test Commands"
console = Console()


class MyCmd(Cmd):
    @with_category(TEST_CATAGORY)
    def do_exemplar(self, args: Namespace):
        """This is what the status is expected to look like."""
        with console.status("Working!"):
            for _ in range(10):
                console.log("console.log: Hello from cmd!")
                time.sleep(0.1)

    @with_category(TEST_CATAGORY)
    def do_problem(self, args: Namespace):
        """Demonstrate the problem."""
        with console.status("Working!"):
            for _ in range(10):
                self.pwarning("pwarning: Hello from cmd!")
                time.sleep(0.1)


if __name__ == "__main__":
    MyCmd().cmdloop()
```

<img width="390" height="456" alt="Image" src="https://github.com/user-attachments/assets/a7141af4-b665-4709-a7f1-8e3b341425a0" />

### Solutions

I did some digging into `rich` and `cmd2` and could not find a way to insert my own console instance into the `cmd2.Cmd` for use with the print methods (`cmd2.Cmd.p*`). If there is an obvious solution, please let me know. I have been a long time user of this package and the new changes integrating `rich` are very exciting.

One possible fix would be to not create a new console object for each call to the print methods. I am not really sure how easy that will be to change or maintain. There could be a console attribute for the `cmd2.Cmd` class or `cmd2` could manage a global console object like how `rich` [does](https://github.com/Textualize/rich/blob/master/rich/__init__.py#L23).

ShiftPoster:
I think a `print_to()` console argument would work perfectly, its not really a major issue.

I went down a rabbit hole with the console attribute scenario and had an idea for a context manager to handle the state. If you guys decide to go down the attribute route, hopefully it will be helpful. The test cases are pretty bare, I had trouble wrapping head around existing tests and trying to think of better test cases. Source is here: https://github.com/ShiftPoster/cmd2-poc

AlwaysTwilight:
[...]

PR #1592, opened 2026-02-28
Added ability to pass a console object to Cmd.print_to().

kmvanbrunt:
See #1578

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,9 @@
 # 3.2.3 (TBD)
 
-  - Enhancements
-    - Added ability to pass a console object to `Cmd.print_to()`. This provides support for
-      things like wrapping a `print_to()` call in a `console.status()` or `console.capture()`
-      context manager.
+- Enhancements
+    - Added ability to pass a console object to `Cmd.print_to()`. This provides support for things
+      like wrapping a `print_to()` call in a `console.status()` or `console.capture()` context
+      manager.
 
 ## 3.2.2 (February 21, 2026)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,15 @@
-# 3.2.3 (TBD)
+# 3.3.0 (TBD)
 
 - Enhancements
     - Added ability to pass a console object to `Cmd.print_to()`. This provides support for things
       like wrapping a `print_to()` call in a `console.status()` or `console.capture()` context
       manager.
+
+- Breaking Changes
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+# 3.2.3 (TBD)
+
+- Enhancements
+    - Added ability to pass a console object to `Cmd.print_to()`. This provides support for things
+      like wrapping a `print_to()` call in a `console.status()` or `console.capture()` context
+      manager.
```
tleonhardt, line 6: We should document that the `file` parameter of `cmd2.Cmd.print_to` has been renamed to `destination` and that this is a potentially breaking change.

As such we should probably make it a 3.3.0 release instead of 3.2.3.

cmd2/cmd2.py
```diff
@@ -1259,13 +1264,30 @@ def print_to(
...
+        """Print objects to a given destination (file stream or cmd2 console).
 
         This method is configured for general-purpose printing. By default, it enables
         soft wrap and disables Rich's automatic detection for markup, emoji, and highlighting.
         These defaults can be overridden by passing explicit keyword arguments.
 
-        :param file: file stream being written to
+        See the Rich documentation for more details on emoji codes, markup tags, and highlighting.
+
+        !!! note
```
tleonhardt, line 1293: Oh wow, I didn't realize it would render things like notes correctly within docstrings.  This is great. I tested it locally and verified it works as intended.

tleonhardt (changes requested):

cmd2/cmd2.py
```diff
@@ -1247,30 +1252,66 @@ def visible_prompt(self) -> str:
...
+        destination: IO[str] | Cmd2BaseConsole,
         *objects: Any,
         sep: str = " ",
         end: str = "\n",
         style: StyleType | None = None,
-        soft_wrap: bool = True,
-        emoji: bool = False,
-        markup: bool = False,
-        highlight: bool = False,
+        soft_wrap: bool | None = None,
```
tleonhardt, line 1260: Should we add more unit tests for output methods now that these parameters can also be `None` instead of a `bool`?

CHANGELOG.md
kmvanbrunt: Done

cmd2/cmd2.py
kmvanbrunt: Since `Cmd2GeneralConsole` already sets these values to `False`, changing the arg defaults to `None` doesn't affect existing behavior.

tleonhardt (approved):

PR #1600, opened 2026-03-07
Update cmd2.Cmd.select to use prompt-toolkit choice

tleonhardt:
The [choice](https://python-prompt-toolkit.readthedocs.io/en/3.0.52/pages/asking_for_a_choice.html) function from `prompt-toolkit` is now used for `cmd2.Cmd.select` method when both stdin and stdout are TTYs.

Key Changes:
   - `prompt_toolkit.shortcuts.choice` integration: The `select` method now utilizes the modern, interactive choice shortcut when both stdin and stdout are TTYs. This provides a more user-friendly selection menu (usually supports arrow keys and searching).
   - Backward Compatibility: Maintained the original numbered-list implementation as a fallback for non-TTY environments. This ensures that existing scripts, pipes, and tests (which mock read_input) continue to function correctly.
   - Robust Argument Handling: Standardized the conversion of various input formats (strings, lists of strings, lists of tuples) to the (value, label) format required by choice.
   - Error Handling: Wrapped the choice call in a loop and a try-except block to correctly handle KeyboardInterrupt (Ctrl-C) by printing ^C and re-raising, and to handle cancellations by reprompting, maintaining consistency with original select behavior.

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -73,7 +73,7 @@ from prompt_toolkit import (
     filters,
     print_formatted_text,
 )
-from prompt_toolkit.application import get_app
+from prompt_toolkit.application import create_app_session, get_app
 from prompt_toolkit.auto_suggest import AutoSuggestFromHistory
 from prompt_toolkit.completion import Completer, DummyCompleter
 from prompt_toolkit.formatted_text import ANSI, FormattedText
@@ -4399,10 +4399,11 @@ class Cmd:
                 except (IndexError, TypeError):
                     fulloptions.append((opt[0], str(opt[0])))
 
-        if self.stdin.isatty() and self.stdout.isatty():
+        if self._is_tty_session(self.main_session):
             try:
                 while True:
-                    result = choice(message=prompt, options=fulloptions)
+                    with create_app_session(input=self.main_session.input, output=self.main_session.output):
+                        result = choice(message=prompt, options=fulloptions)
                     if result is not None:
                         return result
             except KeyboardInterrupt:
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -82,7 +82,7 @@ from prompt_toolkit.input import DummyInput, create_input
 from prompt_toolkit.key_binding import KeyBindings
 from prompt_toolkit.output import DummyOutput, create_output
 from prompt_toolkit.patch_stdout import patch_stdout
-from prompt_toolkit.shortcuts import CompleteStyle, PromptSession, set_title
+from prompt_toolkit.shortcuts import CompleteStyle, PromptSession, choice, set_title
 from rich.console import (
     Group,
     RenderableType,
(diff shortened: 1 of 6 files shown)


examples/remove_settable.py
```diff
@@ -9,6 +12,12 @@ def __init__(self) -> None:
         super().__init__()
         self.remove_settable('debug')
 
+    def do_eat(self, arg):
```
kmvanbrunt: Why put an example for `select()` in the example about removing `Settables`?

cmd2/cmd2.py
```diff
@@ -4387,15 +4387,29 @@ def select(self, opts: str | Iterable[str] | Iterable[tuple[Any, str | None]], p
...
-                    fulloptions.append((opt[0], opt[1]))
-                except IndexError:
-                    fulloptions.append((opt[0], opt[0]))
+                    val = opt[0]
+                    text = str(opt[1]) if len(opt) > 1 and opt[1] is not None else str(val)
+                    fulloptions.append((val, text))
+                except (IndexError, TypeError):
+                    fulloptions.append((opt[0], str(opt[0])))
+
+        if self.stdin.isatty() and self.stdout.isatty():
```
kmvanbrunt: I suggest changing this to `if self._is_tty_session(self.main_session):`

Otherwise the `choice` function won't run when in a pyscript or  when redirecting the output of a command.

tests/test_cmd2.py
```diff
@@ -1745,6 +1745,83 @@ def test_select_ctrl_c(outsim_app, monkeypatch) -> None:
     assert out.rstrip().endswith('^C')
 
 
+def test_select_choice_tty(outsim_app, monkeypatch) -> None:
+    # Mock choice to return the first option
+    choice_mock = mock.MagicMock(name='choice', return_value='sweet')
+    monkeypatch.setattr("cmd2.cmd2.choice", choice_mock)
+
+    # Mock isatty to be True for both stdin and stdout
+    monkeypatch.setattr(outsim_app.stdin, "isatty", lambda: True)
```
kmvanbrunt: If you make the logic change I suggested in `select()`, replace these `isatty` monkeypatches with something like this:

```Python
with create_pipe_input() as pipe_input:
        base_app.main_session = PromptSession(
            input=pipe_input,
            output=DummyOutput(),
    )

    prompt = 'Sauce? '
    options = ['sweet', 'salty']
    ...
```

Do the same for the next test.

kmvanbrunt (changes requested):

cmd2/cmd2.py
kmvanbrunt: Also do this on the `choice()` call so `prompt-toolkit` uses our IO streams. Otherwise it will default to `sys.stdin` and `sys.stdout`, which may actually be different if a dev passed in custom `self.stdin` and `self.stdout` values to `cmd2.Cmd()`.

```Python
with create_app_session(input=self.main_session.input, output=self.main_session.output):
    result = choice(message=prompt, options=fulloptions)
```

examples/remove_settable.py
tleonhardt: I wanted to put an example for using select somewhere.  We had an example in our documentation but now anywhere in our examples.  This was just a random example that had almost no code in it.  I'm happy to move this somewhere else if you think it would feel more appropriate.

tleonhardt: In retrospect this did feel out of place here, so I moved it to the `read_input.py` example instead.

kmvanbrunt (approved):

PR #1609, opened 2026-03-17
Removed need for Rich patches in Cmd2BaseConsole.print() and Cmd2BaseConsole.log().

kmvanbrunt:


cmd2/rich_utils.py
```diff
@@ -221,52 +201,32 @@ def print(
         soft_wrap: bool | None = None,
         new_line_start: bool = False,
     ) -> None:
-        """Override to support ANSI sequences and address a bug in Rich.
+        """Override to support ANSI sequences.
```
tleonhardt, line 204: If I understand the changes in this PR correctly, the `Cmd2BaseConsole` class itself (or specifically the `print` and `log` method within it) still have the Rich bug if used directly.

I recommend updating their docstrings to add a warning and recommend using the constructor wrapper added here.

Perhaps something like:
```Python
"""
Note: Due to a bug in Rich where complex renderables (like Table and Rule)
may not receive formatting settings passed directly to this method, it is
recommended to pass them to the _create_base_printing_console console constructor instead.
See https://github.com/Textualize/rich/issues/4028
"""
```

This same comment applies to the `log` method.

cmd2/cmd2.py
```diff
@@ -1315,6 +1315,27 @@ def visible_prompt(self) -> str:
         """
         return su.strip_style(self.prompt)
 
+    def _create_base_printing_console(
```
tleonhardt, line 1318: This could be a little more flexible. While currently sufficient for `print_to`, adding `soft_wrap` to this helper would make it more consistent with the other formatting defaults being moved to the constructor and could provide a central place to manage all console defaults.

Suggest adding a `soft_wrap: bool = True,` parameter.

kmvanbrunt: This is just a temporary function until the Rich bug is fixed.
I only added it so I didn't have to repeat the code and comment in two places.

tleonhardt (approved):

cmd2/rich_utils.py
kmvanbrunt: All Rich console classes currently have the bug, so it's nothing specific to `Cmd2BaseConsole`.

Also, `_create_base_printing_console()` is not intended to be a public method so I don't want to point users to it.

PR #1610, opened 2026-03-17
Added 'justify' to the print methods.

kmvanbrunt:


cmd2/cmd2.py
```diff
@@ -1344,6 +1345,7 @@ def print_to(
         end: str = "\n",
         style: StyleType | None = None,
         soft_wrap: bool = True,
+        justify: JustifyMethod | None = None,
```
tleonhardt, line 1348: Just curious - is there a reason we added it in the middle instead of at the end of the parameters?

More importantly, if I understand `JustifyMethod` correctly, it is a literal string type with certain fixed possible values, one of which is `default`.  I think we should not allow the none type and make the default uh duh `"default"` ;-) 

This 2nd comment applies everywhere you use `JustifyMethod`.

cmd2/cmd2.py
```diff
@@ -1700,6 +1716,7 @@ def ppaged(
                     sep=sep,
                     end=end,
                     style=style,
+                    justify=justify,
                     soft_wrap=soft_wrap,
                     **(rich_print_kwargs if rich_print_kwargs is not None else {}),
                 )
```
tleonhardt, line 1722: The `else:` case below for where we are not outputing to a pager and just use `self.poutput` directly doesn't pass through the `justify` argument. It doesn't feel right to lose justification in this case?

tleonhardt (changes requested):

cmd2/cmd2.py
tleonhardt: Alternatively, it might be good to keep keyword-only arguments in alphabetical order for maintainability

tleonhardt (dismissed):

cmd2/cmd2.py
kmvanbrunt: Fixed

tleonhardt (dismissed):

PR #1612, opened 2026-03-21
feat: Add Typer/Click support as alternative to argparse

KelvinChung2000:
Summary                                                                                                   
                                                                                                            
  - Add `@cmd2.with_typer` decorator for defining commands using type annotations and Typer/Click instead of argparse                               
  - Support two patterns: auto-built from method signature (simple commands) and explicit `typer.Typer()` app (subcommands)
  - Integrate with cmd2 help, tab completion, error handling, and CommandSet                                                                      
                                                                                                                                                  
  What's included                                                                                                                                 
                                                                                                                                                  
  - cmd2/typer_custom.py — Typer-specific logic for building Click commands, completion, and subcommand resolution                                
  - cmd2/decorators.py — `@with_typer` decorator                                                              
  - cmd2/cmd2.py — Extended _CommandParsers and help/completion paths to handle both argparse and Typer commands                                  
  - tests/test_typer.py — 35 tests covering execution, help, completion, CommandSet, and subcommands                                              
  - docs/features/typer.md — Documentation with usage examples                                                                                    
  - examples/typer_example.py — Working example app                                                                                               
  - Optional dependency via `pip install cmd2[typer]`                                                                                               
                                                                                                                                                  
  Test plan                                                                                                                                       
                                                                                                                                                  
  - make check passes (ruff, mypy)                                                                                                                
  - All existing tests pass (no regressions)                                                                
  - python examples/typer_example.py runs, and tab completion works

The following is a screenshot of the typer_example.py

<img width="1251" height="870" alt="image" src="https://github.com/user-attachments/assets/13806c58-e1b6-4a07-ae4c-4aaadbd235a4" />

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -204,6 +204,14 @@ from .utils import (
     suggest_similar,
 )
 
+try:
+    from .typer_custom import TyperParser
+except ImportError:
+
+    class TyperParser:  # type: ignore[no-redef]
+        """Sentinel: isinstance checks always return False when typer is not installed."""
+
+
 if TYPE_CHECKING:  # pragma: no cover
     StaticArgParseBuilder = staticmethod[[], argparse.ArgumentParser]
     ClassArgParseBuilder = classmethod['Cmd' | CommandSet, [], argparse.ArgumentParser]
@@ -229,6 +237,7 @@ class _CommandParsers:
     """Create and store all command method argument parsers for a given Cmd instance.
 
     Parser creation and retrieval are accomplished through the get() method.
+    Supports both argparse-based and Typer/Click-based commands.
     """
 
     def __init__(self, cmd: 'Cmd') -> None:
@@ -236,7 +245,8 @@ class _CommandParsers:
 
         # Keyed by the fully qualified method names. This is more reliable than
         # the methods themselves, since wrapping a method will change its address.
-        self._parsers: dict[str, argparse.ArgumentParser] = {}
+        # Values are argparse.ArgumentParser for argparse commands or TyperParser (click.Command) for Typer commands.
+        self._parsers: dict[str, argparse.ArgumentParser | TyperParser] = {}
 
     @staticmethod
     def _fully_qualified_name(command_method: CommandFunc) -> str:
@@ -250,15 +260,16 @@ class _CommandParsers:
         """Return whether a given method's parser is in self.
 
         If the parser does not yet exist, it will be created if applicable.
-        This is basically for checking if a method is argarse-based.
+        This is basically for checking if a method uses argparse or Typer.
         """
         parser = self.get(command_method)
         return bool(parser)
 
-    def get(self, command_method: CommandFunc) -> argparse.ArgumentParser | None:
-        """Return a given method's parser or None if the method is not argparse-based.
+    def get(self, command_method: CommandFunc) -> argparse.ArgumentParser | TyperParser | None:
+        """Return a given method's parser or None if the method is not parser-based.
 
         If the parser does not yet exist, it will be created.
+        Handles both argparse and Typer/Click commands.
         """
         full_method_name = self._fully_qualified_name(command_method)
         if not full_method_name:
@@ -268,6 +279,14 @@ class _CommandParsers:
             if not command_method.__name__.startswith(COMMAND_FUNC_PREFIX):
                 return None
             command = command_method.__name__[len(COMMAND_FUNC_PREFIX) :]
+
+            # Check for Typer command
+            if getattr(command_method, constants.CMD_ATTR_TYPER_FUNC, None) is not None:
+                from .typer_custom import build_typer_command
+
+                parent = self._cmd.find_commandset_for_command(command) or self._cmd
+                self._parsers[full_method_name] = build_typer_command(parent, command_method)
+                return self._parsers[full_method_name]
 
             parser_builder = getattr(command_method, constants.CMD_ATTR_ARGPARSER, None)
             if parser_builder is None:
@@ -1030,7 +1049,7 @@ class Cmd:
             # is the actual command since command synonyms don't own it.
             if cmd_func_name == command_method.__name__:
                 command_parser = self._command_parsers.get(command_method)
-                if command_parser is not None:
+                if isinstance(command_parser, argparse.ArgumentParser):
                     check_parser_uninstallable(command_parser)
 
     def _register_subcommands(self, cmdset: Union[CommandSet, 'Cmd']) -> None:
@@ -1075,7 +1094,7 @@ class Cmd:
             if command_func is None:
                 raise CommandSetRegistrationError(f"Could not find command '{command_name}' needed by subcommand: {method}")
             command_parser = self._command_parsers.get(command_func)
-            if command_parser is None:
+            if not isinstance(command_parser, argparse.ArgumentParser):
                 raise CommandSetRegistrationError(
                     f"Could not find argparser for command '{command_name}' needed by subcommand: {method}"
                 )
@@ -1161,7 +1180,7 @@ class Cmd:
                 # but keeping in case it does for some strange reason
                 raise CommandSetRegistrationError(f"Could not find command '{command_name}' needed by subcommand: {method}")
             command_parser = self._command_parsers.get(command_func)
-            if command_parser is None:  # pragma: no cover
+            if not isinstance(command_parser, argparse.ArgumentParser):  # pragma: no cover
                 # This really shouldn't be possible since _register_subcommands would prevent this from happening
                 # but keeping in case it does for some strange reason
                 raise CommandSetRegistrationError(
@@ -2333,11 +2352,11 @@ class Cmd:
                 if func_attr is not None:
                     completer_func = func_attr
                 else:
-                    # There's no completer function, next see if the command uses argparse
+                    # There's no completer function, next see if the command uses a parser
                     func = self.cmd_func(command)
                     argparser = None if func is None else self._command_parsers.get(func)
 
-                    if func is not None and argparser is not None:
+                    if func is not None and isinstance(argparser, argparse.ArgumentParser):
                         # Get arguments for complete()
                         preserve_quotes = getattr(func, constants.CMD_ATTR_PRESERVE_QUOTES)
                         cmd_set = self.find_commandset_for_command(command)
@@ -2348,6 +2367,16 @@ class Cmd:
 
                         completer_func = functools.partial(
                             completer.complete, tokens=raw_tokens[1:] if preserve_quotes else tokens[1:], cmd_set=cmd_set
+                        )
+                    elif func is not None and isinstance(argparser, TyperParser):
+                        from .typer_custom import typer_complete
+
+                        preserve_quotes = getattr(func, constants.CMD_ATTR_PRESERVE_QUOTES)
+                        completer_func = functools.partial(
+                            typer_complete,
+                            self,
+                            command_func=func,
+                            args=raw_tokens[1:] if preserve_quotes else tokens[1:],
                         )
                     else:
                         completer_func = self.completedefault  # type: ignore[assignment]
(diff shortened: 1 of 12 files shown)


docs/features/index.md
```diff
@@ -3,6 +3,7 @@
 <div class="grid cards" markdown>
 <!--intro-start-->
 - [Argument Processing](argument_processing.md)
+- [Typer Support](typer.md)
```
tleonhardt, line 6: Please keep this in alphabetical order

docs/features/typer.md
```diff
@@ -0,0 +1,188 @@
...
+  `with_unknown_args`
+- your application already has substantial argparse parser customizations
+- you want `cmd2`'s rich-argparse help formatting (Typer uses Click's own help formatter)
+
+## Notes
+
+- `help <command>` uses Click's help output for Typer-based commands, not argparse help.
+- Parse errors use Click's error formatting and are caught so they do not exit the REPL.
+- Completion for Typer-based commands is provided by Click's `shell_complete` API.
+- Typer support is optional and requires installing the `typer` extra.
```
tleonhardt, line 185: I think we need to put this in a `!!! warning` or something similar. The fact that `Typer`-based commands use different help and completion mechanisms is something people should be aware of.

The help formatting isn't bad I don't mind it, but it may feel odd when an app contains inconsistent help formatting for built-in commands versus Typer-based ones. One thing that would help is if all the Typer-based commands automatically supported a `-h`/`--help` argument. The fact that doesn't work feels badly inconsistent.  Perhaps if we put all `Typer`-based commands in a different default category if none is specified that could at least give users a reasonable expectation of different behavior?

The completion also mostly works fine, but there are a few little annoying differences. One thing that would help is if when a user hits `<TAB>` after a command it would show a hint similar to what `cmd2` does for `argparse`-based commands. - e.g. if you type `history <TAB>` it shows a hint for the required argument. But when you are using the `typer_example.py` example, if you type `assign <TAB>` nothing happens at all.

examples/typer_example.py
```diff
@@ -0,0 +1,194 @@
...
+
+
+class TyperExampleApp(cmd2.Cmd):
+    """A cmd2 application showcasing every Typer integration feature."""
+
+    intro = "Welcome! Type 'help' to list commands. Try tab-completion!\n"
+    prompt = "typer-demo> "
+
+    # 1. Simple positional argument + option with default
+    @cmd2.with_typer
```
tleonhardt, line 68: I'd recommend putting all of the Typer-based commands in a different category to make it easier to see from `help` or `help -v` which the custom commands are.

cmd2/cmd2.py
```diff
@@ -229,14 +237,16 @@ class _CommandParsers:
...
+    Supports both argparse-based and Typer/Click-based commands.
     """
 
     def __init__(self, cmd: 'Cmd') -> None:
         self._cmd = cmd
 
         # Keyed by the fully qualified method names. This is more reliable than
         # the methods themselves, since wrapping a method will change its address.
-        self._parsers: dict[str, argparse.ArgumentParser] = {}
+        # Values are argparse.ArgumentParser for argparse commands or TyperParser (click.Command) for Typer commands.
```
tleonhardt, line 248: Don't need this extra comment - the updated type hint should suffice

cmd2/cmd2.py
```diff
@@ -204,6 +204,14 @@ def __init__(self, msg: str = '') -> None:
     suggest_similar,
 )
 
+try:
+    from .typer_custom import TyperParser
```
tleonhardt, line 208: I see what this is attempting to do, but it won't work as-is.

Even if `typer` isn't present, you can still import `TyperParser` from `.typer_customer` without an import error so long as `click` is present since it is a `TypeAlias` for `click.Command`. A great many things in the Python ecosystem depend upon `click`, so it wouldn't be uncommon for that to be present when `typer` isn't - for example `zensical` which `cmd2` uses for documentation depends upon `click`.

Even if both `click` and `typer` are not present, it will raise a `ModuleNotFoundError` on all non-EOL modern Python versions and not an `ImportError`.

One option might be to directly import `typer` in `typer_custom.py` and change this except to catch a `ModuleNotFoundError`.  Another might be able to define a module-level boolean variable called something like `has_typer` based on whether or not `typer` is importable.

However we fix it, I think we should also configure a `TypeAlias` here called something like `Cmd2Parser` that is just aliased to `argparse.ArgumentParser` when `typer` isn't present and `argparse.ArgumentParser | TyperParser` when it is.  Then that type alias should be used in other type hints in this file as appropriate.

Something along these lines:

```Python
try:
    import typer
    from .typer_custom import TyperParser
    Cmd2ArgParser: TypeAlias = argparse.ArgumentParser | TyperParser
except ModuleNotFoundError:
    Cmd2ArgParser: TypeAlias = argparse.ArgumentParser 
    class TyperParser:  # type: ignore[no-redef]
        """Sentinel: isinstance checks always return False when typer is not installed."""
```

Then down below we can use something like:
```Python
self._parsers: dict[str, Cmd2ArgParser] = {}
```

cmd2/cmd2.py
```diff
@@ -250,15 +260,16 @@ def __contains__(self, command_method: CommandFunc) -> bool:
...
         If the parser does not yet exist, it will be created if applicable.
-        This is basically for checking if a method is argarse-based.
+        This is basically for checking if a method uses argparse or Typer.
         """
         parser = self.get(command_method)
         return bool(parser)
 
-    def get(self, command_method: CommandFunc) -> argparse.ArgumentParser | None:
-        """Return a given method's parser or None if the method is not argparse-based.
+    def get(self, command_method: CommandFunc) -> argparse.ArgumentParser | TyperParser | None:
```
tleonhardt, line 268: If a `Cmd2Parser` type alias is defined above as suggested, can replace type hint with it here:

```Python
def get(self, command_method: CommandFunc) -> Cmd2ArgParser | None:
```

cmd2/typer_custom.py
```diff
@@ -0,0 +1,319 @@
...
+        # Merge in kwargs passed programmatically (e.g. do_method(..., key=val))
+        # so they take effect even when Click doesn't see them on the command line.
+        extra_kwargs = getattr(command_method, constants.CMD_ATTR_TYPER_KWARGS, {})
+        merged_kwargs = dict(kwargs)
+        for key, value in extra_kwargs.items():
+            if key not in merged_kwargs or merged_kwargs[key] is None:
+                merged_kwargs[key] = value
+        if has_self:
+            return target_func(parent, *args, **merged_kwargs)
+        return target_func(*args, **merged_kwargs)
```
tleonhardt, line 80: Add unit test to cover this `else` case for when `has_self` is `False`.

cmd2/typer_custom.py
```diff
@@ -0,0 +1,319 @@
...
+    :param app: the Typer app to bind
+    :param parent: object which owns the command (Cmd or CommandSet instance)
+    :param command_method: the decorated command method
+    :return: deep copy of the app with bound callbacks
+    """
+    bound_app = copy.deepcopy(app)
+
+    def _bind_group(group: 'typer.Typer') -> None:
+        if group.registered_callback is not None and callable(group.registered_callback.callback):
+            group.registered_callback.callback = build_typer_callback(
```
tleonhardt, line 106: Add unit test to cover code inside this `if`

cmd2/typer_custom.py
```diff
@@ -0,0 +1,319 @@
...
+                command_method,
+            )
+
+        for command_info in group.registered_commands:
+            if command_info.callback is not None:
+                command_info.callback = build_typer_callback(command_info.callback, parent, command_method)
+
+        for group_info in group.registered_groups:
+            if callable(group_info.callback):
+                group_info.callback = build_typer_callback(group_info.callback, parent, command_method)
```
tleonhardt, line 118: Add unit test to cover code inside this if

cmd2/typer_custom.py
```diff
@@ -0,0 +1,319 @@
...
+
+    :param parent: object which owns the command (Cmd or CommandSet instance)
+    :param command_method: the decorated command method
+    :return: Click Command object
+    :raises ImportError: if typer is not installed
+    """
+    try:
+        import typer
+    except ModuleNotFoundError as exc:
+        raise ImportError("Typer support requires 'cmd2[typer]' to be installed") from exc
```
tleonhardt, line 171: Add unit test to cover this except case

cmd2/typer_custom.py
```diff
@@ -0,0 +1,319 @@
...
+    :param args: parsed argument tokens so far
+    :return: a Completions object
+    """
+    from click.shell_completion import CompletionItem as ClickCompletionItem
+    from click.shell_completion import ShellComplete
+
+    try:
+        parser = cmd._command_parsers.get(command_func)
+        if not isinstance(parser, click.Command):
+            return Completions()
```
tleonhardt, line 230: Add unit test to cover this line

cmd2/typer_custom.py
```diff
@@ -0,0 +1,319 @@
...
+                display=item.value,
+                display_meta=item.help or "",
+            )
+            if isinstance(item, ClickCompletionItem)
+            else CompletionItem(value=item)
+            for item in results
+        ]
+        return Completions(items=items)
+    except Exception as exc:  # noqa: BLE001
+        return Completions(error=cmd.format_exception(exc))
```
tleonhardt, line 247: Add unit test to cover this line

cmd2/typer_custom.py
```diff
@@ -0,0 +1,319 @@
...
+    for subcommand in subcommands:
+        commands = getattr(current, 'commands', None)
+        if commands is None or subcommand not in commands:
+            raise KeyError(subcommand)
+        current = commands[subcommand]
+        resolved_names.append(current.name or subcommand)
+    return current, resolved_names
+
+
+def typer_complete_subcommand_help(
```
tleonhardt, line 269: This commend has no test coverage. Add tests to cover it.

cmd2/decorators.py
```diff
@@ -344,6 +346,120 @@ def cmd_wrapper(*args: Any, **kwargs: dict[str, Any]) -> bool | None:
...
+            self.poutput(str(a + b))
+    ```
+
+    """
+    import functools
+
+    try:
+        import typer
+    except ModuleNotFoundError as exc:
+        raise ImportError("Typer support requires the 'typer' package. Install it with: pip install cmd2[typer]") from exc
```
tleonhardt, line 381: Add test to cover this case

cmd2/decorators.py
```diff
@@ -344,6 +346,120 @@ def cmd_wrapper(*args: Any, **kwargs: dict[str, Any]) -> bool | None:
...
+                    args=parsed_arglist,
+                    prog_name=command_name,
+                    standalone_mode=False,
+                )
+            except Exception as exc:
+                if isinstance(exc, click.ClickException):
+                    exc.show(file=sys.stderr)
+                    raise Cmd2ArgparseError from exc
+                if isinstance(exc, (click.exceptions.Exit, SystemExit)):
+                    raise Cmd2ArgparseError from exc
```
tleonhardt, line 429: Add tests to cover both this `raise` and the empty one below

tleonhardt (changes requested):

cmd2/cmd2.py
```diff
@@ -4097,11 +4126,30 @@ def complete_help_subcommands(
...
-        completer = argparse_completer.DEFAULT_AP_COMPLETER(argparser, self)
-        return completer.complete_subcommand_help(text, line, begidx, endidx, arg_tokens['subcommands'])
+        argparser = self._command_parsers.get(func)
+        if isinstance(argparser, argparse.ArgumentParser):
+            completer = argparse_completer.DEFAULT_AP_COMPLETER(argparser, self)
+            return completer.complete_subcommand_help(text, line, begidx, endidx, arg_tokens['subcommands'])
+
+        # Typer/Click subcommand help completion
+        if isinstance(argparser, TyperParser):
+            from .typer_custom import typer_complete_subcommand_help
```
tleonhardt, line 4140: Add tests to cover this `if` case and the generic return for the default `else` case after it

docs/features/typer.md
KelvinChung2000: After you pointed it out, I personally really dislike the mismatch. I am considering building an argparser alongside, so all text rendering will use the rich-argparse pipeline. Then add an extra flag to allow the user to opt out of rich-argparse and back to using click. 

The main problem is having two sets of parsing systems, and things cannot be reused between them. The best solution I came up with is to use a flag to choose which parser format the built-in command uses. But this means will need to update/rewrite to decouple the built-in command from a specific parser to allow using either parser format

PR #1614, opened 2026-03-23
feat: Annotated based argparse, and auto completer inference

KelvinChung2000:
This is a full rework of #1612. Instead of wrapping Typer. We now extract the types and build the `argparse` parser. 

I want to mark it as a draft for now, as some of the stuff will likely need a bit more cleanup. Please have a look at the documentation and example, and let me know if I missed anything obvious.


cmd2/annotated.py
```diff
@@ -0,0 +1,457 @@
...
+        # Fallback to name lookup
+        try:
+            return enum_class[value]
+        except KeyError as err:
+            valid = ', '.join(_value_map)
+            raise argparse.ArgumentTypeError(f"invalid choice: {value!r} (choose from {valid})") from err
+
+    _convert.__name__ = enum_class.__name__
+    # Preserve the enum class for downstream consumers like tab completion.
+    _convert._cmd2_enum_class = enum_class
```
tleonhardt: Need to fix fundamental type error causing all tests and `mypy` to fail:

```sh
cmd2/annotated.py:234:5: error: "Callable[[str], Enum]" has no attribute "_cmd2_enum_class"  [attr-defined]
```

tleonhardt (changes requested):

cmd2/annotated.py
KelvinChung2000: I am aware of this. I am working on the code's style right now. I will push once it is at a good enough state. 

This page of code contains all the logic required to generate the argparser. While it is not too difficult to understand, I don't like the long if-else chain because it is not intuitive.

tleonhardt: I haven't look at this particular case, but since `cmd2` now only supports Python 3.10 or newer, using the `match` statement can be a nice option for structural pattern matching with match/case to replace some long if/else chains.

KelvinChung2000: I originally considered something like a dict dispatch pattern:

```python
def dis(type_to_resolve):
    type_dict = {Enum: _reslove_enum}
    return type_dict[type_to_resolve]
```

But the things that return for the `_resolve_*` are too different, and are slowly moving toward more like filling in an "`argparse` specification," then having a function that takes in this specification and turns it into a parser.  

I like this dict dispatch pattern because it is very easy to read and extend, but it might not be the right pattern here.

cmd2/annotated.py
```diff
@@ -0,0 +1,609 @@
...
+    sig = inspect.signature(func)
+    try:
+        hints = get_type_hints(func, include_extras=True)
+    except (NameError, AttributeError, TypeError) as exc:
+        raise TypeError(
+            f"Failed to resolve type hints for {func.__qualname__}. Ensure all annotations use valid, importable types."
+        ) from exc
+
+    for name, param in sig.parameters.items():
+        if name == 'self':
```
tleonhardt: I think there is the potential for info on internal cmd2 framework parameters to leak into the CLI arguments here.

Parameters injected internally by the `cmd2` decorator (`cmd2_statement`, `cmd2_handler`, and `_unknown`) are currently not skipped during parser generation. If a user includes these in their function signature to receive them, they will be erroneously exposed in `--help` and parsed by argparse (potentially as required positional arguments), before being immediately overwritten by the decorator.

Suggested change:
```Python
if name in ('self', 'cmd2_statement', 'cmd2_handler', '_unknown'):
    continue
```

cmd2/annotated.py
```diff
@@ -0,0 +1,609 @@
+"""Build argparse parsers from type-annotated function signatures.
+
+.. warning:: Experimental
+
+   This module is experimental and its behaviour may change in future releases.
```
tleonhardt: Please prefer American English spellings - i.e. `behavior` without the `u`

cmd2/annotated.py
```diff
@@ -0,0 +1,609 @@
...
+        kwargs, metadata, positional, _is_bool_flag = _resolve_annotation(
+            annotation,
+            has_default=has_default or is_kw_only,
+            default=default,
+        )
+
+        # Keyword-only without default becomes a required option
+        if is_kw_only and not has_default:
+            positional = False
+            kwargs['required'] = True
```
tleonhardt: The file header documentation states:

> Keyword-only parameters (after ``*``) always become options; without a default they are required. 

Recommend changing this logic and comment to match that, e.g.:

```Python
# Keyword-only parameters always become options; if no default, they are required
if is_kw_only:
    positional = False
    if not has default:
        kwargs['required'] = True
```

tleonhardt (changes requested):
Overall I like the user experience of this better - there is one essentially consistent behavior throughout for help and completion all based on argparse.

I'm a little concerned about how much code this adds and if that might be a maintenance burden.

I left a few comments where I think there are a couple minor edge case bugs. I'll need to experiment some more with this once you address the comments here.

cmd2/annotated.py
```diff
@@ -0,0 +1,453 @@
...
+    parser = DEFAULT_ARGUMENT_PARSER()
+
+    sig = inspect.signature(func)
+    try:
+        hints = get_type_hints(func, include_extras=True)
+    except (NameError, AttributeError, TypeError):
+        hints = {}
+
+    for name, param in sig.parameters.items():
+        if name == 'self':
```
tleonhardt: If a user follows a different naming convention for the method's first parameter than `self` (e.g., `this` or `cls`), it will be incorrectly treated as a command-line argument. It is safer to skip the first parameter by position, as `cmd2` commands are intended to be methods.

Perhaps something like this?
```Python
parameters = list(sig.parameters.values())
if parameters:
    parameters.pop(0)

for param in parameters:
    name = param.name
```

cmd2/argparse_completer.py
```diff
@@ -729,7 +731,26 @@ def _get_raw_choices(self, arg_state: _ArgumentState) -> list[CompletionItem] |
...
+        action_type = arg_state.action.type
+        if action_type is not None:
+            if action_type is pathlib.Path or (isinstance(action_type, type) and issubclass(action_type, pathlib.Path)):
+                from .cmd2 import Cmd
+
+                return ChoicesCallable(is_completer=True, to_call=Cmd.path_complete)
+
+            if isinstance(action_type, type) and issubclass(action_type, enum.Enum):
+                return [CompletionItem(str(m.value), display_meta=m.name) for m in action_type]
+
```
tleonhardt, line 724: If I understand correctly, positional boolean arguments lack tab completion. While positional boolean args are pretty uncommon, positional arguments using `_parse_bool` (which is assigned by the decorator) don't have explicit choices, so the completer will only show a generic hint. Providing standard boolean strings as completion items would improve the UX.

Recommend adding something like this here:

```Python
# Check for boolean converter assigned by @with_annotated
if action_type.__name__ == '_parse_bool':
    return [CompletionItem(v) for v in ['true', 'false', 'yes', 'no', 'on', 'off', '1', '0']]
```

cmd2/decorators.py
```diff
@@ -344,6 +346,85 @@ def cmd_wrapper(*args: Any, **kwargs: Any) -> bool | None:
...
+            unknown_param = inspect.signature(fn).parameters.get('_unknown')
+            if unknown_param is None:
+                raise TypeError('with_annotated(with_unknown_args=True) requires a parameter named _unknown')
+            if unknown_param.kind is inspect.Parameter.POSITIONAL_ONLY:
+                raise TypeError('Parameter _unknown must be keyword-compatible when with_unknown_args=True')
+
+        command_name = fn.__name__[len(constants.COMMAND_FUNC_PREFIX) :]
+
+        @functools.wraps(fn)
+        def cmd_wrapper(*args: Any, **_kwargs: Any) -> bool | None:
```
tleonhardt: The `cmd_wrapper` captures `**_kwargs` but they are never used or passed to the wrapped function `fn`. If `with_annotated` is used in a context where other decorators pass keyword arguments, they will be lost.

Recommend something like:

```Python
def cmd_wrapper(*args: Any, **kwargs: Any) -> bool | None:
...
func_kwargs.update(kwargs)
result: bool | None = fn(owner, **func_kwargs)
return result
```

tleonhardt (changes requested):

cmd2/annotated.py
```diff
@@ -0,0 +1,875 @@
...
+
+    # Skip the first parameter by position (self/cls for methods)
+    params = list(sig.parameters.items())
+    if params:
+        params = params[1:]
+
+    for name, param in params:
+        if name in skip_params:
+            continue
+
```
tleonhardt, line 670: Positional-only parameters are not rejected during parser construction but will crash at runtime.

The `@with_annotated` decorator's wrapper ultimately calls the decorated function using `fn(owner, **func_kwargs)`. Because it unpacks all parsed arguments as keyword arguments, any positional-only parameter (indicated by a `/` in the function signature) will raise a `TypeError` when the command is executed. It is safer to explicitly reject them early during the parser construction phase.

This can be done by adding a check like the following here:
```Python
if param.kind == inspect.Parameter.POSITIONAL_ONLY:
    raise TypeError(
        f"Parameter {name!r} in {func.__qualname__} is positional-only, "
        f"which is not supported by @with_annotated because parameters are passed as keyword arguments."
    )
```

tleonhardt (changes requested):

cmd2/argparse_completer.py
```diff
@@ -715,6 +717,16 @@ def print_help(self, tokens: Sequence[str], file: IO[str] | None = None) -> None
 
     def _choices_to_items(self, arg_state: _ArgumentState) -> list[CompletionItem]:
```
tleonhardt, line 712: The `_resolve_enum` function correctly adds the `choices` kwarg for Enum types, which means `arg_state.action.choices` will be populated (not `None`). Because this `if` block requires `arg_state.action.choices is None`, it will be skipped entirely for Enums defined via `@with_annotated`, falling back to the standard choices logic which does not populate `display_meta`. 

You should check the action type regardless, and use `choices` to filter the returned members if it is populated.

Recommend something along these lines:

```Python
    def _choices_to_items(self, arg_state: _ArgumentState) -> list[CompletionItem]:
        """Convert choices from action to list of CompletionItems."""
        action_type = arg_state.action.type
        if action_type is not None:
            enum_class = None
            if isinstance(action_type, type) and issubclass(action_type, enum.Enum):
                enum_class = action_type
            else:
                enum_class = getattr(action_type, '_cmd2_enum_class', None)

            if isinstance(enum_class, type) and issubclass(enum_class, enum.Enum):
                if arg_state.action.choices is None:
                    return [CompletionItem(str(m.value), display_meta=m.name) for m in enum_class]
                return [
                    CompletionItem(str(m.value), display_meta=m.name)
                    for m in enum_class
                    if m.value in arg_state.action.choices or m.name in arg_state.action.choices or m in arg_state.action.choices
                ]

            if arg_state.action.choices is None and getattr(action_type, '__name__', None) == '_parse_bool':
                return [CompletionItem(v) for v in ['true', 'false', 'yes', 'no', 'on', 'off', '1', '0']]

        if arg_state.action.choices is None:
            return []

```

cmd2/annotated.py
```diff
@@ -0,0 +1,881 @@
...
+        kwargs = resolver(tp, args, **ctx)
+        base_type = kwargs.pop('base_type', tp)
+    else:
+        base_type = tp
+        kwargs = {}
+
+    if metadata:
+        kwargs.update(metadata.to_kwargs())
+        if metadata.nargs is not None:
+            kwargs['nargs'] = metadata.nargs
```
tleonhardt: The `action` parameter of the `Option` class is completely dropped for non-boolean types.

The `Option` metadata explicitly accepts an `action` argument, but it is missing from `_BaseArgMetadata._KWARGS_MAP` and never injected back into the argparse `kwargs` during `_resolve_type` for non-boolean types. This means that a user providing `Option(action='count')` for an integer will have their action silently ignored by argparse.

Recommend adding a block after this, e.g.:

```Python
if metadata:
    kwargs.update(metadata.to_kwargs())
    if metadata.nargs is not None:
        kwargs['nargs'] = metadata.nargs
    if isinstance(metadata, Option) and getattr(metadata, 'action', None) is not None:
        if 'action' not in kwargs:
            kwargs['action'] = metadata.action
```

tleonhardt (changes requested):

cmd2/argparse_completer.py
KelvinChung2000: Should I actually inline all of them to the `annotated.py`,  such that the completer is only automatically inferred if you are using the annotation method? Depends on how one views this sort of automation: either too much magic or a very nice-to-have. 

Another option would be removing the choice setting logic from the `annotated.py` and concentrating all of them here?

tleonhardt: I think the more we can keep the logic for this PR purely within `annotated.py`, the better.

KelvinChung2000:
Are you thinking of including this in the 4.0, or will you keep it after the 4.0? If you are planning this to happen after 4.0, we can leave this PR for now, and I will do a single rebase once 4.0 is out.

tleonhardt:
@KelvinChung2000 Once the state of 4.0 stabilizes (probably soon), then @kmvanbrunt will need to discuss whether we want to try to get this into a Release Candidate or address it after the release of 4.0.

I do like the developer experience of what is here.

Also, in case you weren't aware, you'll need to get the latest in and there are conflicts.

KelvinChung2000:
Sure. Just let me know once things are stabilised, and I will do a rebase afterwards.

tleonhardt:
@KelvinChung2000 I invited you to be a member of our organization. This will enable you to create a branch in the main repo which will in turn make collaboration with other developers on the project easier.

Could you please close this PR in its current form as a fork and re-open it from a branch on the main repo?

We would want to release this as an `Experimental Feature` that is in beta testing with no promises that it would survive long-term within `cmd2` proper until we get user feedback. As such, it it might be best if all code for the feature live within a single new file - i.e. it might be best to move code out of `decorators.py` into `annotated.py`.

We are getting close to a 4.0 release and the hectic pace of changes should be slowing down, so I anticipate painful merge conflicts should be fewer.

PR #1618, opened 2026-03-28
Improve subcommand manipulation.

kmvanbrunt:


cmd2/argparse_custom.py
```diff
@@ -977,6 +984,44 @@ def _SubParsersAction_remove_parser(self: argparse._SubParsersAction, name: str)
...
+# Patch argparse._SubParsersAction to add add_existing_parser function
+############################################################################################################
+
+
+def _SubParsersAction_add_existing_parser(  # noqa: N802
+    self: argparse._SubParsersAction,  # type: ignore[type-arg]
+    name: str,
+    subcmd_parser: argparse.ArgumentParser,
+    **add_parser_kwargs: Any,
+) -> None:
```
tleonhardt, line 957: The standard `argparse.add_parser` method returns the newly created parser. To maintain a similar API and allow for potential chaining or immediate use of the return value, `add_existing_parser` should probably return `subcmd_parser`.

cmd2/decorators.py
```diff
@@ -373,13 +375,13 @@ def arg_decorator(func: ArgparseCommandFunc[CmdOrSet]) -> ArgparseCommandFunc[Cm
...
 
         # Keyword arguments for subparsers.add_parser()
-        add_parser_kwargs: dict[str, Any] = {}
+        final_kwargs: dict[str, Any] = dict(add_parser_kwargs)
         if help is not None:
-            add_parser_kwargs['help'] = help
+            final_kwargs['help'] = help
         if aliases:
-            add_parser_kwargs['aliases'] = aliases[:]
+            final_kwargs['aliases'] = tuple(aliases)
```
tleonhardt, line 383: Is the conversion to a `tuple` necessary here?

The change from `aliases[:]` (which creates a shallow copy of a list) to tuple(aliases) is safe, but since `argparse` accepts any sequence for aliases, a simple copy or even passing the original sequence is usually sufficient. 

Is the intent to ensure immutability?

cmd2/argparse_custom.py
```diff
@@ -977,6 +984,44 @@ def _SubParsersAction_remove_parser(self: argparse._SubParsersAction, name: str)
 
 setattr(argparse._SubParsersAction, 'remove_parser', _SubParsersAction_remove_parser)
 
+############################################################################################################
+# Patch argparse._SubParsersAction to add add_existing_parser function
+############################################################################################################
+
+
+def _SubParsersAction_add_existing_parser(  # noqa: N802
```
tleonhardt: Should we document this new method in the CHANGELOG? If so, should it be public and not have the leading underscore?

tleonhardt (dismissed):

cmd2/decorators.py
kmvanbrunt: Since I'm keeping a copy of what they passed in, I want it immutable.

cmd2/argparse_custom.py
kmvanbrunt: This method isn't public, but the monkey-patched version is (`action.add_existing_parser()`).
We never documented the other monkey-patched additions to `argparse` in CHANGELOG, but they are in the [docs](https://cmd2.readthedocs.io/en/stable/api/argparse_custom/).

tleonhardt (dismissed):

cmd2/argparse_custom.py
kmvanbrunt: I feel returning a value is misleading about the intention of this function. It's not a factory, it's a side-effect function.

kmvanbrunt: To clarify, the `subcmd_parser` variable isn't being edited or copied by `add_existing_parser()`, so the caller can edit it before and after passing it. The intention is really for that parser to be done before passing it anyway.

tleonhardt (approved):

PR #1621, opened 2026-04-01
Refactor argparse custom

kmvanbrunt:
- Require devs to use `Cmd2ArgumentParser-based` parsers.
- Moved previously-patched functions into `Cmd2ArgumentParser`.
- Simplified creation of cmd2-specific `argparse.Action` attributes.
- Made common prefix for all private cmd2 attributes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,8 @@ prompt is displayed.
     - Removed `Cmd.ruler` since `cmd2` no longer uses it.
     - All parsers used with `cmd2` commands much be an instance of `Cmd2ArgumentParser` or a child
       class of it.
+    - Removed `set_ap_completer_type()` and `get_ap_completer_type()` since `ap_completer_type` is
+      now a public member of `Cmd2ArgumentParser`.
 - Enhancements
     - New `cmd2.Cmd` parameters
         - **auto_suggest**: (boolean) if `True`, provide fish shell style auto-suggestions. These
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,7 +64,7 @@ prompt is displayed.
       before calling it like the previous functions did.
     - Removed `Cmd.default_to_shell`.
     - Removed `Cmd.ruler` since `cmd2` no longer uses it.
-    - All parsers used with `cmd2` commands much be an instance of `Cmd2ArgumentParser` or a child
+    - All parsers used with `cmd2` commands must be an instance of `Cmd2ArgumentParser` or a child
       class of it.
     - Removed `set_ap_completer_type()` and `get_ap_completer_type()` since `ap_completer_type` is
       now a public member of `Cmd2ArgumentParser`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,8 @@ prompt is displayed.
       before calling it like the previous functions did.
     - Removed `Cmd.default_to_shell`.
     - Removed `Cmd.ruler` since `cmd2` no longer uses it.
+    - All parsers used with `cmd2` commands much be an instance of `Cmd2ArgumentParser` or a child
+      class of it.
 - Enhancements
     - New `cmd2.Cmd` parameters
         - **auto_suggest**: (boolean) if `True`, provide fish shell style auto-suggestions. These
--- a/cmd2/argparse_custom.py
+++ b/cmd2/argparse_custom.py
@@ -2,16 +2,9 @@
 
 It also defines a parser class called Cmd2ArgumentParser which improves error
 and help output over normal argparse. All cmd2 code uses this parser and it is
-recommended that developers of cmd2-based apps either use it or write their own
-parser that inherits from it. This will give a consistent look-and-feel between
-the help/error output of built-in cmd2 commands and the app-specific commands.
-If you wish to override the parser used by cmd2's built-in commands, see
-custom_parser.py example.
-
-Since the new capabilities are added by patching at the argparse API level,
-they are available whether or not Cmd2ArgumentParser is used. However, the help
-and error output of Cmd2ArgumentParser is customized to notate nargs ranges
-whereas any other parser class won't be as explicit in their output.
+required that developers of cmd2-based apps either use it or write their own
+parser that inherits from it. If you wish to override the parser used by cmd2's
+built-in commands, see custom_parser.py example.
 
 
 **Added capabilities**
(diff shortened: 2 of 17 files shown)


cmd2/cmd2.py
```diff
@@ -911,7 +907,7 @@ def _build_parser(
             parser = parser_builder.__func__(parent.__class__)
         elif callable(parser_builder):
             parser = parser_builder()
```
tleonhardt: If `parser_builder` is a callable that returns a standard `argparse.ArgumentParser` instead of a `Cmd2ArgumentParser`, the current code will bypass the runtime type check. This will later result in an `AttributeError` when `cmd2` attempts to access `parser.ap_completer_type` on the parser.

Recommend adding a check immediately after this for the type, e.g.:

```Python
if not isinstance(parser, Cmd2ArgumentParser):
    raise TypeError(f"parser_builder must return a Cmd2ArgumentParser, not {type(parser)}")
```

cmd2/argparse_custom.py
```diff
@@ -376,322 +352,83 @@ def set_parser_prog(parser: argparse.ArgumentParser, prog: str) -> None:
...
+    :param param_name: Name of the parameter. This must be a valid Python identifier.
+    :param validator: Optional function to validate and/or transform the parameter value.
+                      It accepts the Action instance and the value as arguments.
+    """
+    if not param_name.isidentifier():
+        raise KeyError(f"Invalid parameter name '{param_name}' - cannot be used as a python identifier")
 
-    This function is added by cmd2 as a method called ``get_choices_callable()`` to ``argparse.Action`` class.
+    attr_name = constants.cmd2_attr_name(param_name)
+    if param_name in CUSTOM_ACTION_ATTRIBS or hasattr(argparse.Action, attr_name):
```
tleonhardt: I don't think this is the best way of checking for a duplicate property.


 The check `hasattr(argparse.Action, attr_name)` will sometimes evaluate to `False` because the underlying custom parameter attributes (e.g., `_cmd2_test`) are stored as instance variables, not class variables. To accurately prevent overwriting existing accessor methods, we should probably check for the presence of the getter method on the class instead, e.g.:

```Python
if param_name in CUSTOM_ACTION_ATTRIBS or hasattr(argparse.Action, f'get_{param_name}'):
```

CHANGELOG.md
```diff
@@ -64,6 +64,10 @@ prompt is displayed.
       before calling it like the previous functions did.
     - Removed `Cmd.default_to_shell`.
     - Removed `Cmd.ruler` since `cmd2` no longer uses it.
+    - All parsers used with `cmd2` commands much be an instance of `Cmd2ArgumentParser` or a child
```
tleonhardt: type: should be **must**, not much

docs/migrating/next_steps.md
```diff
@@ -9,13 +9,13 @@ leveraging other `cmd2` features. The three ideas here will get you started. Bro
 For all but the simplest of commands, it's probably easier to use
 [argparse](https://docs.python.org/3/library/argparse.html) to parse user input than to do it
 manually yourself for each command. `cmd2` provides a `@with_argparser()` decorator which associates
-an `ArgumentParser` object with one of your commands. Using this method will:
+an `Cmd2ArgumentParser` object with one of your commands. Using this method will:
```
tleonhardt: Grammar: should be `a` not `an`

cmd2/decorators.py
```diff
@@ -213,17 +216,17 @@ def cmd_wrapper(*args: Any, **kwargs: Any) -> bool | None:
...
     *,
     ns_provider: Callable[..., argparse.Namespace] | None = None,
     preserve_quotes: bool = False,
     with_unknown_args: bool = False,
 ) -> Callable[[ArgparseCommandFunc[CmdOrSet]], RawCommandFuncOptionalBoolReturn[CmdOrSet]]:
-    """Decorate a ``do_*`` method to populate its ``args`` argument with the given instance of argparse.ArgumentParser.
+    """Decorate a ``do_*`` method to populate its ``args`` argument with the given instance of Cmd2ArgumentParser.
 
-    :param parser: instance of ArgumentParser or a callable that returns an ArgumentParser for this command
+    :param parser: instance of Cmd2ArgumentParser or a callable that returns an Cmd2ArgumentParser for this command
```
tleonhardt: Grammar: should be a not an

cmd2/decorators.py
```diff
@@ -359,7 +362,7 @@ def as_subcommand_to(
 
     :param command: Command Name. Space-delimited subcommands may optionally be specified
     :param subcommand: Subcommand name
-    :param parser: instance of ArgumentParser or a callable that returns an ArgumentParser for this subcommand
+    :param parser: instance of Cmd2ArgumentParser or a callable that returns an Cmd2ArgumentParser for this subcommand
```
tleonhardt: Grammar: should be a not an

tleonhardt (changes requested):

CHANGELOG.md
kmvanbrunt: Done

cmd2/decorators.py
kmvanbrunt: Done

cmd2/decorators.py
kmvanbrunt: Done

docs/migrating/next_steps.md
kmvanbrunt: Done

cmd2/cmd2.py
kmvanbrunt: That's true of the `staticmethod` and `classmethod` return values as well.

cmd2/argparse_custom.py
kmvanbrunt: Checking `CUSTOM_ACTION_ATTRIBS` is meant to look for the presence of our attributes.
The second check is to prevent us from overwriting an existing attribute of `argparse.Action`.

kmvanbrunt: I'll separate these checks with different error messages to make it more clear.

kmvanbrunt: I updated the error handling to address these concerns.

tleonhardt (dismissed):

cmd2/cmd2.py
kmvanbrunt: It now checks for a `Cmd2ArgumentParser` return value from each of those factories.

tleonhardt (approved):
LGTM

PR #1623, opened 2026-04-03
Added public API for attaching/detaching a subcommand with an existing parser.

kmvanbrunt:


cmd2/cmd2.py
```diff
@@ -1026,18 +1025,22 @@ def unregister_command_set(self, cmdset: CommandSet) -> None:
...
-                            )
-                        check_parser_uninstallable(subparser)
-                    break
+            try:
+                subparsers_action = parser._get_subparsers_action()
+            except ValueError:
+                # No subcommands to check
+                return
+
+            for subparser in subparsers_action.choices.values():
```
tleonhardt, line 1040: `subparsers_action.choices.values()` yields the exact same parser instance multiple times if a subcommand has aliases. This causes `check_parser_uninstallable` to redundantly check the same parser and its entire subcommand tree multiple times, potentially leading to an explosion of checks for deeply nested command sets.

Suggest eliminating this redundant traversal using something like the following:

```Python
# Prevent redundant traversal of parser aliases
checked_parsers: set[Cmd2ArgumentParser] = set()
for raw_subparser in subparsers_action.choices.values():
    subparser = cast(Cmd2ArgumentParser, raw_subparser)
    if subparser in checked_parsers:
        continue
    checked_parsers.add(subparser)
    attached_cmdset_id = getattr(subparser, constants.PARSER_ATTR_COMMANDSET_ID, None)
    if attached_cmdset_id is not None and attached_cmdset_id != cmdset_id:
        raise CommandSetRegistrationError(
            f"Cannot uninstall CommandSet: '{subparser.prog}' is required by another CommandSet"
        )
    check_parser_uninstallable(subparser)
```

tleonhardt (changes requested):

cmd2/cmd2.py
kmvanbrunt: Fixed in two places.

CHANGELOG.md
```diff
@@ -68,6 +68,7 @@ prompt is displayed.
       class of it.
     - Removed `set_ap_completer_type()` and `get_ap_completer_type()` since `ap_completer_type` is
       now a public member of `Cmd2ArgumentParser`.
+    - Moved `set_parser_prog()` to `Cmd2ArgumentParser.update_prog()`.
```
tleonhardt, line 71: Should we update anything in the documentation in relation to this PR? Perhaps in the migration guide for 4.x?

tests/test_commandset.py
```diff
@@ -98,7 +98,8 @@ def do_main(self, args: argparse.Namespace) -> None:
     # main -> sub
     subcmd_parser = cmd2.Cmd2ArgumentParser(description="Sub Command")
 
-    @cmd2.as_subcommand_to('main', 'sub', subcmd_parser, help="sub command")
+    # Include aliases to cover the alias check in cmd2's check_parser_uninstallable().
+    @cmd2.as_subcommand_to('main', 'sub', subcmd_parser, help="sub command", aliases="sub_alias")
```
tleonhardt: Passing a string to aliases iterates over its characters.

The `aliases` argument expects a `Sequence[str]`. Passing a raw string "sub_alias" will cause `argparse` to iterate over each character, inadvertently registering 9 single-character aliases ('s', 'u', 'b',
  etc.) instead of the full intended word.

Change it to `aliases=["sub_alias"]`.

tleonhardt (changes requested):

tests/test_commandset.py
kmvanbrunt: Oops. It's fixed now.

tleonhardt (approved):

PR #1625, opened 2026-04-05
Removed Cmd2AttributeWrapper class.

kmvanbrunt:


cmd2/constants.py
```diff
@@ -32,35 +32,56 @@
...
+    """Build a private attribute name with the _cmd2_ prefix.
 
     :param name: the name of the attribute
     :return: the prefixed attribute name
     """
-    return f'{CMD2_ATTR_PREFIX}{name}'
+    return f'{PRIVATE_ATTR_PREFIX}{name}'
+
 
+def cmd2_public_attr_name(name: str) -> str:
```
tleonhardt, line 50: Would this be better just be left as `cmd2_attr_name`?  If we are renaming it like this here, should we mention that in the `CHANGELOG`?

tleonhardt (dismissed):

cmd2/constants.py
kmvanbrunt: None of these naming functions exist in an official version of cmd2 yet, so no need to mention in `CHANGELOG`.
Also, besides the two new `Namespace` attributes, `constants.py` is not part of the public API.

tleonhardt (dismissed):

tleonhardt (approved):
LGTM

PR #1628, opened 2026-04-12
Fix ansi color dumb terminal

Nilsonfsilva:
Hi

I am the Debian package maintainer for cmd2. While updating the package to
version 3.4.0 and running the test suite in the Debian build infrastructure
(sbuild/autopkgtest), I encountered 16 test failures related to ANSI color
output.

The Debian build infrastructure always sets TERM=dumb to prevent any
interactive terminal usage during builds. This exposed a bug in cmd2: even
with ALLOW_STYLE=ALWAYS explicitly set, all ANSI sequences were silently
stripped and no colors were emitted.

The root cause is in Rich's Console initialization. When color_system is not
explicitly specified, Rich calls _detect_color_system() to auto-detect the
terminal capabilities. With TERM=dumb, is_dumb_terminal returns True and
_detect_color_system() returns None — meaning no color support — regardless
of force_terminal=True. So force_terminal=True alone is not sufficient to
guarantee color output when the terminal is detected as dumb.

This affects two places in rich_utils.py:
- Cmd2BaseConsole.__init__() sets force_terminal=True when ALLOW_STYLE=ALWAYS
  but does not set color_system, so Rich silently discards all colors.
- rich_text_to_string() has the same issue.

The fix passes color_system="truecolor" explicitly in both places when the
intent is to always emit ANSI styles, bypassing the broken auto-detection.
The default mode (TERMINAL) is unaffected and continues to respect TERM=dumb
as expected.

Additionally, test_ansi_terminal_tty was implicitly relying on the developer's
TERM environment variable. The fix adds a mock for TERM=xterm-256color to make
the test self-contained and environment-independent.

Nilson Silva

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+## 3.2.0 (TBD)
+
+- Bug Fixes
+    - Fixed incompatibilities with Python 3.14.3.
+
+- Potentially Breaking Changes
+    - To avoid future incompatibilities with argparse, we removed most of our overridden help
+      functions. This should not break an application, but it could affect unit tests which parse
+      help text.
+
 ## 3.1.3 (February 3, 2026)
 
 - Bug Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 3.2.2 (February 21, 2026)
+
+- Bug Fixes
+    - Updated `rich_utils.ANSI_STYLE_SEQUENCE_RE` to only match ANSI SGR (Select Graphic Rendition)
+      sequences for text styling.
+
 ## 3.2.1 (February 21, 2026)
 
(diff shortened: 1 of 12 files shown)


tleonhardt:
@Nilsonfsilva Hey there, thanks for the PR and the detailed description.  

It appears your branch is not up to date with the latest `main` branch and their are conflicts that need to be resolved so I can't auto-update it.  Would you please be so kind as to update your fork with the latest code from `main` and then I can get the GitHub Actions CI to run and review this.

Nilsonfsilva:
Hi!
You can close this PR, as I rebuilt it from the updated branch: https://github.com/python-cmd2/cmd2/pull/1629

Nilson Silva

tleonhardt:
Closing at PR author's request since he has another verison.

PR #1633, opened 2026-04-13
Force truecolor support to avoid automatic color detection. (3.x)

kmvanbrunt:
Explicitly set the color system to "truecolor" in `Cmd2BaseConsole` and `rich_text_to_string()` when styling is allowed. This avoids Rich's automatic color detection, which can strip colors in test environments where TERM=dumb is set.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 3.5.0 (TBD)
+
+- Bug Fixes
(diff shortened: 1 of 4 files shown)


tleonhardt (approved):

tleonhardt:
@kmvanbrunt What is the plan for making similar changes on the `main` branch?

PR #1635, opened 2026-04-16
Overhauled custom types.

kmvanbrunt:


cmd2/argparse_completer.py
```diff
@@ -35,9 +35,9 @@
 from .exceptions import CompletionError
 from .rich_utils import Cmd2SimpleTable
 from .types import (
-    ChoicesProviderUnbound,
-    CmdOrSet,
-    CompleterUnbound,
+    CmdOrSetT,
```
tleonhardt, line 38: Why change `CmdOrSet` to `CmdOrSetT`?

cmd2/command_set.py
```diff
@@ -39,37 +36,30 @@ def __init__(self) -> None:
...
 
+            class MyCommandSet(CommandSet[MyCustomApp]):
+                ...
 
         :raises CommandSetRegistrationError: if CommandSet is not registered.
         """
-        if self._cmd_internal is None:
-            raise CommandSetRegistrationError('This CommandSet is not registered')
-        return self._cmd_internal
+        if (cmd := self._cmd_internal) is not None:
```
tleonhardt: You can still use the walrus operator to do assignment in the conditional but check for the anomalous case in the if and return `cmd` afterwards.  I think leaving it like that as a guard clause would be more readable.

cmd2/command_set.py
```diff
@@ -119,13 +109,13 @@ def add_settable(self, settable: Settable) -> None:
 
         :param settable: Settable object being added
         """
-        if self._cmd_internal is not None:
-            if not self._cmd.always_prefix_settables:
-                if settable.name in self._cmd.settables and settable.name not in self._settables:
+        if (cmd := self._cmd_internal) is not None:
```
tleonhardt, line 111: This logic is now subtly different than it was previously.  Previously inside the `if` methods were being called on `self._cmd`.  But now they are being called on `self._cmd_internal`.

cmd2/cmd2.py
```diff
@@ -164,10 +160,12 @@
 )
 from .styles import Cmd2Style
 from .types import (
-    ChoicesProviderUnbound,
+    BoundCommandFunc,
+    BoundCompleter,
     CmdOrSet,
-    CompleterBound,
-    CompleterUnbound,
+    CmdOrSetT,
```
tleonhardt, line 166: I am so confused.  Why do we have both `CmdOrSet` and `CmdOrSetT`.  Can we find a better name for one of them?

I intuitively get what `CmdOrSet` is. 

I don't think I properly understand what `CmdOrSetT` is. I read `CmdOrSetT` as `CmdOrSet Type`, and `CmdOrSet` is literally the type alias for that. So the fact that we have something else that is `CmdOrSetT` confuses me.  Perhaps we can give that later one a more self-describing name?

cmd2/cmd2.py
```diff
@@ -164,10 +160,12 @@
 )
 from .styles import Cmd2Style
 from .types import (
-    ChoicesProviderUnbound,
+    BoundCommandFunc,
```
tleonhardt, line 163: I like putting the `Bound` and `Unbound` first in these type names as it helps it stand out to make it more clear what is expected.

cmd2/command_set.py
kmvanbrunt: `self._cmd` is a property which returns `self._cmd_internal`, but it raises a `CommandSetRegistrationError` if `self._cmd_internal` is still `None`.

cmd2/argparse_completer.py
kmvanbrunt: `CmdOrSet` previously was a `TypeVar`.
It's now a `TypeAlias` and the `TypeVar` is called `CmdOrSetT`.

cmd2/cmd2.py
kmvanbrunt: I agree. Our custom class names use prefixes as well. (e.g. `Cmd2ArgumentParser`, `RawDescriptionCmd2HelpFormatter`).

cmd2/decorators.py
```diff
@@ -363,10 +365,27 @@ def as_subcommand_to(
...
+    base_parser.add_subparsers(metavar='SUBCOMMAND', required=True)
+    sub_parser = cmd2.Cmd2ArgumentParser()
+
+    class MyApp(cmd2.Cmd):
+        @cmd2.with_argparser(base_parser)
+        def do_base(self, args: argparse.Namespace) -> None:
+            args.cmd2_subcmd_handler(args)
+
+        @cmd2.as_subcommand_to('base', 'sub', sub_parser, help="the subcommand")
+        def sub_handler(self, args: argparse.Namespace) -> None:
```
tleonhardt, line 382: Gemini complained about the type hint of `None` being incorrect in this example code in the docstring.

kmvanbrunt: An earlier commit mistakenly had the `-> None` on the `as_subcommand_to` decorator.
Is that what Gemini is referring to?

cmd2/command_set.py
kmvanbrunt: I simplified this code.

cmd2/decorators.py
tleonhardt: Yes, I think so

cmd2/cmd2.py
kmvanbrunt: This is about the difference between types (including `TypeAlias`) and `TypeVar`.
```Python
# A Cmd or CommandSet instance
CmdOrSet: TypeAlias = Union["Cmd", "CommandSet[Any]"]

# Tracks the specific subclass instance (either a Cmd or CommandSet)
CmdOrSetT = TypeVar("CmdOrSetT", bound=CmdOrSet)
```

Adding `T` to a type communicates to other devs that it's a `TypeVar` bound to that type. It's still restricted to the same family of that type, but it adds the ability to track the specific one used.

This is why we also have a `TypeVar` for `Cmd` called `CmdT`. It allows us to make `CommandSet` a generic class which knows what type of `Cmd` it is loaded into.

Basically, `TypeVars` are important for preserving the type of something.

For instance:
```Python
# Using the TypeAlias: The IDE loses the specific class info
def get_my_app(app: CmdOrSet) -> CmdOrSet:
    return app

# Using the TypeVar: The IDE remembers the specific class info
def get_my_app(app: CmdOrSetT) -> CmdOrSetT:
    return app
    
# Result:
app = MyCustomApp()
my_app = get_my_app(app) 
# With TypeAlias: 'my_app' is seen as just a 'Cmd'. You lose autocomplete for custom methods.
# With TypeVar: 'my_app' is recognized as 'MyCustomApp'. Autocomplete works.
```

tleonhardt: Ok, I think I grok it now. `TypeVar`s are place holders for use in generic functions. Whereas type aliases are just simply attaching a name to a type.

tleonhardt (approved):

PR #1641, opened 2026-04-20
Fixed CompletionItem.text re-populating from value after being cleared.

kmvanbrunt:


cmd2/completion.py
tleonhardt, line 106: Unless I'm missing something, direct instantiation with `text=""` results in a blank `display`, contradicting this comment.

Since `self.text` can now be explicitly set to an empty string `""`, the logic immediately below will result in display being assigned `""` if the class is instantiated directly with `text=""` and no explicit `display`. This contradicts the comment "Ensure display is never blank" and behaves differently than modifying text via `dataclasses.replace` (which correctly retains the original `display` value).

One option might be replacing the snippet with something like:

```Python
 # Ensure display is never blank.
if not self.display:
    object.__setattr__(self, "display", self.text or str(self.value))
```

kmvanbrunt: This refers to code in the draft and not the final version.

tleonhardt (approved):

PR #1645, opened 2026-04-22
Raise ValueError in attach_subcommand() when the subcommand already exists.

kmvanbrunt:


tests/test_argparse_utils.py
```diff
@@ -467,6 +467,12 @@ def test_subcommand_attachment_errors() -> None:
     with pytest.raises(TypeError, match=r"must be an instance of 'Cmd2ArgumentParser' \(or a subclass\)"):
         root_parser.attach_subcommand([], "sub", ap_parser)  # type: ignore[arg-type]
 
+    # Verify ValueError when subcommand name already exists
+    sub_parser = Cmd2ArgumentParser(prog="sub")
+    root_parser.attach_subcommand([], "sub", sub_parser)
+    with pytest.raises(ValueError, match="conflicting subparser: sub"):
+        root_parser.attach_subcommand([], "sub", sub_parser)
```
tleonhardt, line 474: For Python 3.10, this line is not raising an exception at all.

cmd2/argparse_utils.py
```diff
@@ -752,7 +753,11 @@ def attach_subcommand(
             )
 
         # Use add_parser to register the subcommand name and any aliases
-        placeholder_parser = subparsers_action.add_parser(subcommand, **add_parser_kwargs)
+        try:
+            placeholder_parser = subparsers_action.add_parser(subcommand, **add_parser_kwargs)
+        except ArgumentError as ex:
```
tleonhardt: For Python 3.10, it appears either no exception is happening or something different than an `ArgumentError` that is somehow being swallowed.

tleonhardt (changes requested):

tleonhardt (dismissed):

PR #1647, opened 2026-04-23
#1627: Removed feedback_to_output settable and made pfeedback consistent

### LINKED ISSUE #1627: Settable output is overly verbose and lacks a silent mode
Updating a parameter via `set` currently produces a three-line confirmation. This is unnecessarily verbose and creates visual clutter, especially since the "was" and "now" values are unaligned in separate lines.

Most CLI tools (like gdb) are silent on success or use a single line. I propose:

* Consolidating the output to a single, aligned line (e.g., variable: old -> new).
* Adding an option to silence these notifications entirely.
* Making it silent by default to follow standard CLI conventions.

### Example of current output:
```
(CLI)> set display_theme light
display_theme - was: 'dark'
now: 'light' 

(CLI)> 
```

### Example of desired output:
```
(CLI)> set display_theme light
(CLI)> 
(CLI)> set display_theme purple
Error setting display_theme: invalid choice: 'purple' (choose from 'light', 'dark')
(CLI)>
```

bambu:
### Description

This PR removes the `feedback_to_output` settable.  It also does the following:

- `pfeedback` always outputs to `self.stdout` when the `quiet` settable is `Falsae`
- `cmd2` now makes more consistent use of `peedback` for non-essential output that is silenceable
- Status into that we don't want to be re-directable, like elapsed timing data, now outputs to `stderr`, but without any color or style
- Improve user experience of the set command by consolidating output into 1 line and allowing for the output to be hidden when  `quiet` is set. The previous output was overly verbose (three lines) and could not be silenced.

Closes #1627


kmvanbrunt:
@bambu I need to consider this. I don't think the previous output was too verbose. I'm OK if we reformat it, but the amount of information displayed was, in my view, reasonable.

I'm also concerned about the new output being on one line. A `settable`  could be anything, including an entire command line to run a tool. If we are still going to display both the previous and new value, I think they need to be on separate lines.

Help me understand the reason behind this change. Are you running `set` a lot in scripts and seeing a lot of output or something?

cmd2/cmd2.py
```diff
@@ -4721,7 +4721,16 @@ def do_set(self, args: argparse.Namespace) -> None:
...
+                    # Create the feedback message using Rich Text for color
+                    feedback_msg = Text.assemble(
+                        args.param,
+                        ": ",
+                        (f"{orig_value!r}", "red"),
+                        " -> ",
+                        (f"{settable.value!r}", "green"),
+                    )
+                    self.pfeedback(feedback_msg)
+
```
tleonhardt, line 4731: I see 3 fundamental distinct changes here:
1. Output is more abbreviated - 1 line instead of 2
2. Previous and new values are colorized to enhance readability as a diff
3. `pfeedback` is used instead of `poutput`

In general I think the later two are unquestionably welcome changes.

For most use cases I like the more succinct single-line output format. However, I can envision a theoretical edge case where both the old and new values are many characters such that fitting on a single line might be problematic. I'd be fine with a 2-line solution or with an intelligent splitting based on terminal width.

tleonhardt:
@bambu @kmvanbrunt 
I pushed a change which is a compromise which attempts to preserve the intent of the change in this PR but address the concerns brought up. Below is an image of what the new format looks like.

Screenshot 2026-04-27 at 11 38 18 AM

tleonhardt:
@bambu 
@kmvanbrunt and I chatted about how to best address the intent of this PR. We ended up making more changes because we felt like the overall issue was bigger than just the `set` command output. Now `pfeedback` has a single purpose and timing output isn't gated by two separate settables. Please let us know what you think about the changes.

bambu:
LGTM. It is one line and the output can easily be turned off. Thanks!

kmvanbrunt (approved):

PR #1653, opened 2026-04-24
Add styles to theme for prompt-toolkit completion styles

tleonhardt:
Add styles to theme for completion item and metadata foreground and background colors.

This is an attempt to provide higher-contrast color options and to make them configurable as part of a theme.

Added `rich_to_pt_style` and `rich_to_pt_color` conversion helper functions to convert from `rich` color/style objects to `prompt-toolkit` color/style strings.

- For `ColorSystem.STANDARD` colors (0-15), it maps directly to `prompt-toolkit`'s `ansicolor` names (e.g. `ansired`, `ansibrightred`). This ensures that `prompt-toolkit` delegates to the terminal's theme for the basic 16 colors.
- For `ColorSystem.EIGHT_BIT` and `ColorSystem.TRUECOLOR`, it falls back to parsing the RGB hex code (e.g., `#ff0000`), which `prompt-toolkit` supports natively.

Also added the existing `Cmd2Lexer` colors to the styles and theme.


cmd2/pt_utils.py
```diff
@@ -50,6 +74,46 @@ def pt_filter_style(text: str | ANSI) -> str | ANSI:
...
+
+def rich_to_pt_style(rich_style: StyleType) -> str:
+    """Convert a rich Style object to a prompt_toolkit style string."""
+    if not rich_style:
+        return ""
+
+    if isinstance(rich_style, str):
+        rich_style = Style.parse(rich_style)
+
+    parts = ["noreverse"]
```
kmvanbrunt: What if the Rich style has enabled `reverse`?

cmd2/pt_utils.py
```diff
@@ -50,6 +74,46 @@ def pt_filter_style(text: str | ANSI) -> str | ANSI:
...
+
+    fg_color = rich_to_pt_color(rich_style.color)
+    parts.append(f"fg:{fg_color}")
+
+    bg_color = rich_to_pt_color(rich_style.bgcolor)
+    parts.append(f"bg:{bg_color}")
+
+    if rich_style.bold is not None:
```
kmvanbrunt, line 109: prompt-toolkit has these style attributes in common with Rich. I believe the Rich equivalent to `hidden` is `conceal`.
- bold
- italic
- underline
- blink
- reverse
- hidden

cmd2/pt_utils.py
```diff
@@ -50,6 +74,46 @@ def pt_filter_style(text: str | ANSI) -> str | ANSI:
...
+        return "default"
+
+    # Use prompt_toolkit's 16 standard ansi color names if applicable.
+    # This prevents overriding terminal themes with absolute RGB values.
+    if color.number is not None and 0 <= color.number <= 15:
+        return ANSI_NAMES[color.number]
+
+    # For 8-bit and truecolor, we fallback to hex RGB strings which prompt-toolkit supports natively
+    c = color.get_truecolor()
+    return f"#{c.red:02x}{c.green:02x}{c.blue:02x}"
```
kmvanbrunt, line 90: Make sure to enable true color support on the main session in cmd2.py. It's the `color_depth` parameter to `PromptSession()`.

You'll need to set it on `temp_session` as well in `read_input()`.

cmd2/pt_utils.py
```diff
@@ -50,6 +74,46 @@ def pt_filter_style(text: str | ANSI) -> str | ANSI:
...
 
+def rich_to_pt_color(color: "Color | None") -> str:
+    """Convert a rich Color object to a prompt_toolkit color string."""
+    if not color or color.is_default:
+        return "default"
+
+    # Use prompt_toolkit's 16 standard ansi color names if applicable.
+    # This prevents overriding terminal themes with absolute RGB values.
+    if color.number is not None and 0 <= color.number <= 15:
+        return ANSI_NAMES[color.number]
```
kmvanbrunt, line 86: Prompt-toolkit and Rich use similar color names. The only difference is that prompt-toolkit includes the `ansi` prefix while Rich includes underscores (e.g. ansibrightred = bright_red).

Therefore you could eliminate `ANSI_NAMES` and convert the Rich color names with something like the following.

```Python
pt_color_name = color.name.replace('_', '')
return f"ansi{pt_color_name}"
```

cmd2/cmd2.py
```diff
@@ -716,6 +723,30 @@ def _should_continue_multiline(self) -> bool:
                 # No macro found or already processed. The statement is complete.
                 return False
 
+    def _get_pt_style(self) -> "PtStyle":
+        """Return the prompt_toolkit style for the completion menu."""
```
kmvanbrunt: Should this docstring be more broad? It may not always apply to just the completion menu.

cmd2/cmd2.py
```diff
@@ -716,6 +723,30 @@ def _should_continue_multiline(self) -> bool:
...
+        if self._cached_pt_style is not None and self._cached_pt_style_params == current_params:
+            return self._cached_pt_style
+
+        item_style = rich_to_pt_style(rich_item_style)
+        meta_style = rich_to_pt_style(rich_meta_style)
+
+        self._cached_pt_style_params = current_params
+        self._cached_pt_style = PtStyle.from_dict(
+            {
+                "completion-menu.completion.current": item_style,
```
kmvanbrunt: Are there other completion styles to change (e.g. the color of the completion menu itself)?

cmd2/pt_utils.py
```diff
@@ -196,28 +260,21 @@ class Cmd2Lexer(Lexer):
...
         self.cmd_app = cmd_app
-        self.command_color = command_color
-        self.alias_color = alias_color
-        self.macro_color = macro_color
-        self.flag_color = flag_color
-        self.argument_color = argument_color
+
+        # Retrieve styles dynamically from the current theme
+        theme = ru.get_theme()
+        self.command_color = rich_to_pt_style(theme.styles.get(Cmd2Style.LEXER_COMMAND, ""))
```
kmvanbrunt, line 299: Will these colors change if a user changes the theme during runtime?

kmvanbrunt (changes requested):

cmd2/pt_utils.py
kmvanbrunt: On second thought, keep the color list. The lookup is faster.

cmd2/pt_utils.py
tleonhardt: I fixed this in the latest commit. For efficiency we don't want to retrieve them on every call to `lex_document` since `prompt-toolkit` calls that quite frequently.  I made it so every time `set_theme` is called, it resets these colors for any existing Cmd2Lexer instance.

cmd2/pt_utils.py
tleonhardt: Fixed in latest commit

cmd2/pt_utils.py
tleonhardt: `rich_to_pt_style` now supports all of the style attributes you called out here

cmd2/pt_utils.py
kmvanbrunt: But why are we setting `noreverse` before checking the Rich style? Seems like this line can go.

cmd2/rich_utils.py
```diff
@@ -351,6 +351,15 @@ def set_theme(styles: Mapping[str, StyleType] | None = None) -> None:
     for name in Cmd2HelpFormatter.styles.keys() & theme.styles.keys():
         Cmd2HelpFormatter.styles[name] = theme.styles[name]
 
+    # Update colors in active prompt-toolkit lexers
+    try:
+        from . import pt_utils
+
+        for lexer in pt_utils._lexers:
+            lexer.set_colors()
+    except ImportError:
```
kmvanbrunt: Why would this import ever fail?

cmd2/pt_utils.py
tleonhardt: We aren't - I changed that line to initialize an empty list.

cmd2/rich_utils.py
tleonhardt: It wouldn't I removed this in the latest commit

cmd2/cmd2.py
```diff
@@ -3569,6 +3601,7 @@ def read_input(
             key_bindings=self.main_session.key_bindings,
             input=self.main_session.input,
             output=self.main_session.output,
+            style=DynamicStyle(self._get_pt_style),
```
kmvanbrunt: This should be `self.main_session.style`.
The same applies to `temp_session` in `read_secret()`.

cmd2/pt_utils.py
tleonhardt: Done

cmd2/cmd2.py
tleonhardt: Made it more broad

cmd2/cmd2.py
tleonhardt: Yes, I added the other completion styles.

cmd2/cmd2.py
tleonhardt: Fixed

kmvanbrunt (approved):

PR #1654, opened 2026-04-29
Overhauled redirection and output capturing

kmvanbrunt:
`cmd2` redirection now only captures output directed to `self.stdout` (e.g., via `self.poutput()`). Standard `print()` calls write directly to `sys.stdout` and are not captured. However, `print()` calls within `pyscripts` and the interactive Python shell are treated as command output and sent to `self.stdout`, allowing them to be captured.

This means we no longer alter the system-wide `sys.stdout` stream.

To facilitate this change,  our custom `argparse` code now writes all output to `self.stdout` instead of `sys.stdout`.

To support `pyscript` developers who don't have access to `self.stdout`, the `print()` function in that environment has been replaced with one that wraps `self.print_to()`. It writes to `self.stdout` by default and respects the `allow_style` setting. It also supports printing `Rich` objects. This means that `run_pyscript script.py > out.txt` will capture any `print()` calls in the script.

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -4395,8 +4395,6 @@ class Cmd:
 
     def _print_documented_command_topics(self, header: str, commands: Sequence[str], verbose: bool) -> None:
         """Print topics which are documented commands, switching between verbose or traditional output."""
-        import io
-
         if not commands:
             return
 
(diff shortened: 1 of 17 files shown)


cmd2/cmd2.py
```diff
@@ -15,8 +15,10 @@
 - Redirection to file or paste buffer (clipboard) with > or >>
 - Bash-style ``select`` available
 
-Note, if self.stdout is different than sys.stdout, then redirection with > and |
-will only work if `self.poutput()` is used in place of `print`.
+Note: cmd2 redirection only captures output directed to self.stdout (e.g., via self.poutput()).
+Standard print() calls write directly to sys.stdout and are not captured. However, print() calls
+within pyscripts and the interactive Python shell are treated as command output and sent to
+self.stdout, allowing them to be captured.
```
tleonhardt, line 21: This is important and subtly different than before. I think we should probably make sure to update some more of our documentation to make how this works slightly more clear / discoverable.

Considering updating `generating_output.md` and `redirection.md`.

tleonhardt (changes requested):

tleonhardt (dismissed):

tleonhardt (dismissed):

tleonhardt (dismissed):

tleonhardt (approved):

PR #1657, opened 2026-05-09
Added LRU caching of converted prompt-toolkit styles and colors.

kmvanbrunt:


cmd2/pt_utils.py
```diff
@@ -90,6 +91,7 @@ def rich_to_pt_color(color: "Color | None") -> str:
     return f"#{c.red:02x}{c.green:02x}{c.blue:02x}"
 
 
+@lru_cache(maxsize=1024)
```
tleonhardt, line 95: Adding an `lru_cache` for these two functions is a very good idea. I should have thought of that.

cmd2/cmd2.py
```diff
@@ -724,35 +727,39 @@ def _should_continue_multiline(self) -> bool:
                 # No macro found or already processed. The statement is complete.
                 return False
 
-    def update_pt_style(self) -> None:
-        """Update the cached prompt_toolkit style."""
+    def _get_pt_style(self) -> PtStyle:
```
tleonhardt: I still prefer the callback method over this approach as it never wastes any time calling all the stuff at the top to do the check.  Thus the callback method is more efficient.

cmd2/pt_utils.py
```diff
@@ -309,6 +308,8 @@ def lex_document(self, document: Document) -> Callable[[int], Any]:
         exclude_tokens.update(self.cmd_app.statement_parser.terminators)
         arg_pattern = re.compile(r'(\s+)|(--?[^\s\'"]+)|("[^"]*"?|\'[^\']*\'?)|([^\s\'"]+)')
 
+        self.set_colors()
```
tleonhardt: Again, I like the callback method far better than this as it is much more efficient.

tleonhardt (changes requested):
In general I love the couple places you added an `@lru_cache` and think that is a very smart idea.

However, I strongly prefer the callback approach in the couple places it was used versus the approach here. The callback approach never wastes any time calling any code to check if things changed when it doesn't need to.  This approach here will still waste significant time checking when nothing has changed.

tleonhardt (approved):
Nice, I wasn't aware of `Sytle.null()`

PR #1661, opened 2026-05-11
Added prompt-toolkit version of the app theme which remains synchronized when theme is updated.

kmvanbrunt:


CHANGELOG.md
```diff
@@ -155,8 +155,6 @@ prompt is displayed.
     - Added `Cmd2ArgumentParser.output_to()` context manager to temporarily set the output stream
       during `argparse` operations. This is helpful for directing output for functions like
       `parse_args()`, which default to `sys.stdout` and lack a `file` argument.
-    - Added `cmd2.rich_utils.register_theme_update_callback` function to register callback functions
```
tleonhardt, line 158: I'm pretty sure there has to be something else in the `CHANGELOG` and the documentation that needs to be updated based on these changes.

cmd2/__init__.py
```diff
@@ -51,11 +51,16 @@
...
-    set_theme,
 )
 from .string_utils import stylize
 from .styles import Cmd2Style
+from .theme import (
+    get_pt_theme,
+    get_theme,
+    register_pt_mapping,
+    register_synchronized_prefix,
+    set_theme,
```
tleonhardt, line 59: Any of the examples using anything related to this need to be updated.  This definitely includes `rich_theme.py`.  I'm not sure about others.

tleonhardt (changes requested):

cmd2/__init__.py
```diff
@@ -51,11 +51,16 @@
     RawDescriptionCmd2HelpFormatter,
     RawTextCmd2HelpFormatter,
     TextGroup,
-    get_theme,
-    set_theme,
```
tleonhardt, line 55: Some documentation will need to be updated based on this moving.  Definitely `upgrades.md`.  Maybe others.

tleonhardt (changes requested):

cmd2/theme.py
```diff
@@ -0,0 +1,188 @@
...
+
+    :param style_name: The style name used in the Rich theme.
+    :param pt_ui_names: One or more prompt-toolkit UI component names (e.g., 'completion-menu').
+    """
+    if isinstance(pt_ui_names, str):
+        pt_ui_names = [pt_ui_names]
+
+    # Filter out UI names identical to the style name to avoid redundant registration.
+    unique_names = [n for n in pt_ui_names if n != style_name]
+    _PT_UI_MAP[style_name] = unique_names
```
tleonhardt: Calling this function multiple times with the same style overwrites previous mappings.

Is this the desired behavior? Would it be better to append to the existing mappings instead?

kmvanbrunt: If I append, then there's no way to overwrite a mapping.

If a dev does need to append, they can always get the existing mapping and pass that along with their new mapping.

kmvanbrunt: What if I changed `_PT_UI_MAP` to be a `dict[str, set[str]]`. That would be a safe way to support appending to an existing mapping.

But then I would probably need to add a `clear_pt_mapping()` function.

tleonhardt: I'm flexible.  I was just trying to think of edge cases that might bite someone.

cmd2/theme.py
```diff
@@ -0,0 +1,240 @@
...
+    if pt_ui_names is None:
+        del _PT_UI_MAP[style_name]
+        changed = True
+    else:
+        if isinstance(pt_ui_names, str):
+            pt_ui_names = [pt_ui_names]
+
+        for name in pt_ui_names:
+            _PT_UI_MAP[style_name].discard(name)
+            changed = True
```
tleonhardt: `changed` is unconditionally set to `True` in this loop, regardless whether or not any of the names are actually present (i.e. regardless of whether or not `discard` actually removes anything.

Should we verify the item exists before assuming the map was changed? e.g.:

```Python
for name in pt_ui_names:
    if name in _PT_UI_MAP[style_name]:
            _PT_UI_MAP[style_name].discard(name)
            changed = True
```

tleonhardt (changes requested):

cmd2/theme.py
kmvanbrunt: I fixed this.

CHANGELOG.md
kmvanbrunt: These are not changes to anything that existed in 3.0, so no `CHANGELOG` entry is needed.

I will add documentation in other places though.

kmvanbrunt: There was a reference in `CHANGELOG` that I missed. It's fixed now.

tleonhardt (dismissed):

tleonhardt (approved):

PR #1663, opened 2026-05-14
Expanded subcommand management API.

kmvanbrunt:
1. Detaching a subcommand returns metadata for easily reattaching later
2. Added ability to remove all subcommands of a given command

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -110,7 +110,12 @@ from . import (
 )
 from . import rich_utils as ru
 from . import string_utils as su
-from .argparse_utils import Cmd2ArgumentParser
+from .argparse_utils import (
+    Cmd2ArgumentParser,
+    ParserSource,
+    SubcommandRecord,
+    SubcommandSpec,
+)
 from .clipboard import (
     get_paste_buffer,
     write_to_paste_buffer,
@@ -205,12 +210,7 @@ from .utils import (
 )
 
 if TYPE_CHECKING:  # pragma: no cover
-    StaticArgParseBuilder = staticmethod[[], Cmd2ArgumentParser]
-    ClassArgParseBuilder = classmethod[CmdOrSet, [], Cmd2ArgumentParser]
     from prompt_toolkit.buffer import Buffer
-else:
-    StaticArgParseBuilder = staticmethod
-    ClassArgParseBuilder = classmethod
 
 
 class _SavedCmd2Env:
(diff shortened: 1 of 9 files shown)


cmd2/cmd2.py
```diff
@@ -916,37 +916,48 @@ def register_command_set(self, cmdset: CommandSet[Any]) -> None:
...
-            if isinstance(parser_builder, staticmethod):
-                parser = parser_builder.__func__()
-            elif isinstance(parser_builder, classmethod):
-                parser = parser_builder.__func__(parent.__class__)
-            elif callable(parser_builder):
-                parser = parser_builder()
+            # Inspect the signature to determine if this factory expects a class argument.
+            builder_sig = inspect.signature(parser_source)
+
+            if builder_sig.parameters:
```
tleonhardt, line 951: Here is a comment from Gemini CLI code review, do with it what you will:

```
File: cmd2/cmd2.py
  L944: [MEDIUM] Broad parameter check for callable factories can pass unintended arguments.
  Using if builder_sig.parameters: to determine whether to pass owner.__class__ assumes that any factory function with parameters expects the owner class as its first argument. This will inadvertently pass owner.__class__ to factories that have
  optional parameters (e.g., def factory(prog="default")) or use *args without intending to receive the class, which represents a change from previous behavior. Consider checking if the first parameter specifically accepts a positional argument.

  Suggested change:

   1 -            if builder_sig.parameters:
   2 +            first_param = next(iter(builder_sig.parameters.values()), None)
   3 +            if first_param and first_param.kind in (inspect.Parameter.POSITIONAL_ONLY, inspect.Parameter.POSITIONAL_OR_KEYWORD):
   4                  parser = cast(Callable[[Any], Cmd2ArgumentParser], parser_source)(owner.__class__)
   5              else:
   6                  parser = cast(Callable[[], Cmd2ArgumentParser], parser_source)()
```

tleonhardt (dismissed):
The `test_proc_reader_terminate` test is failing due to a timeout specifically for Window and Python 3.10. It may be a one-off fluke or it might not be.

I don't think I'd worry about it for this PR unless you have insight into what might be causing it. 

We talked earlier today about dropping support for Python 3.10 in the upcoming 4.0.0 release and I am fully in support of that. I think we should do that in a separate PR.

cmd2/cmd2.py
kmvanbrunt: I updated the comment to explain my approach.

tleonhardt (dismissed):

PR #1666, opened 2026-05-15
feat: Experimental annotated argparse

KelvinChung2000:
Adds `@with_annotated`, a type-hint-driven alternative to `@with_argparser` that builds the parser automatically from a command's signature (positional/option inference, enum/literal/path/collection handling, subcommands, groups, mutex). Marked experimental.

  - New module `cmd2/annotated.py` plus `Argument` / `Option` metadata classes exported from `cmd2`
  - Underscored param names auto-dasherize in generated flags (`dry_run` → `--dry-run`); opt out via `Option("--my_flag")`
  - Dedicated docs page `docs/features/annotated.md` with an experimental admonition; `argument_processing.md` keeps a short pointer
  - Example app `examples/annotated_example.py` and test suite in `tests/test_annotated.py`


tleonhardt:
I plan to review this PR this weekend.  @kmvanbrunt @bambu I'd very much appreciate your feedback as well.

cmd2/annotated.py
```diff
@@ -0,0 +1,1095 @@
...
+def with_annotated(
+    func: Callable[..., Any] | None = None,
+    *,
+    ns_provider: Callable[..., argparse.Namespace] | None = None,
+    preserve_quotes: bool = False,
+    with_unknown_args: bool = False,
+    base_command: bool = False,
+    subcommand_to: str | None = None,
+    help: str | None = None,  # noqa: A002
+    aliases: Sequence[str] | None = None,
```
kmvanbrunt: Change this to `aliases: Sequence[str] = ()` to match `as_subcommand_to()'s` signature and since it can't be `None` in a `SubcommandSpec`.

cmd2/annotated.py
```diff
@@ -0,0 +1,1095 @@
...
+            def do_greet(self, name: str, count: int = 1): ...
+
+            @with_annotated(base_command=True)
+            def do_team(self, *, cmd2_handler): ...
+
+            @with_annotated(subcommand_to='team', help='create a team')
+            def team_create(self, name: str): ...
+
+    """
+    if (help is not None or aliases is not None) and subcommand_to is None:
```
kmvanbrunt: Update this check to reflect that `aliases` can no longer be `None`.

cmd2/annotated.py
```diff
@@ -0,0 +1,1095 @@
...
+                subcommand_to,
+                base_command=base_command,
+                groups=groups,
+                mutually_exclusive_groups=mutually_exclusive_groups,
+            )
+            spec = SubcommandSpec(
+                name=subcmd_name,
+                command=subcommand_to,
+                help=help,
+                aliases=tuple(aliases) if aliases else (),
```
kmvanbrunt: Change this to `aliases=tuple(aliases)` since `aliases` can no longer be `None`.

examples/annotated_example.py
```diff
@@ -0,0 +1,314 @@
+#!/usr/bin/env python3
+"""Annotated decorator example -- type-hint-driven argument parsing.
```
kmvanbrunt, line 2: Please fix all Mypy errors in this file.

cmd2/annotated.py
```diff
@@ -0,0 +1,1095 @@
...
+    ns_provider: Callable[..., argparse.Namespace] | None = None,
+    preserve_quotes: bool = False,
+    with_unknown_args: bool = False,
+    base_command: bool = False,
+    subcommand_to: str | None = None,
+    help: str | None = None,  # noqa: A002
+    aliases: Sequence[str] | None = None,
+    groups: tuple[tuple[str, ...], ...] | None = None,
+    mutually_exclusive_groups: tuple[tuple[str, ...], ...] | None = None,
+) -> Any:
```
kmvanbrunt: Please provide the proper return types so this isn't an untyped decorator.

kmvanbrunt (changes requested):

kmvanbrunt:
@KelvinChung2000 

How can I do the following?
1. Set the title and description for an argument group?
2. Set the description and epilog for a parser?
3. Set a custom help formatter class for a parser?
4. Use a custom parser class?

cmd2/__init__.py
```diff
@@ -11,6 +11,11 @@
     rich_utils,
     string_utils,
 )
+from .annotated import (
```
kmvanbrunt: I would prefer not having these in `__init__.py` while annotations is still considered an experimental feature.

cmd2/annotated.py
```diff
@@ -0,0 +1,1194 @@
...
+        if value in value_map:
+            return value_map[value]
+        if value.lower() in _BOOL_TRUE_VALUES:
+            bool_value = True
+        elif value.lower() in _BOOL_FALSE_VALUES:
+            bool_value = False
+        else:
+            bool_value = None
+
+        if bool_value is not None and bool_value in literal_values:
```
tleonhardt: There is a very subtle edge case bug here when integer literal values are used that can incorrectly return booleans due to the fact that in Python `False == 0` and `True == 1` both evaluate to `True`.

Because Python evaluates `False == 0` and `True == 1` as `True`, if a user provides an input like "yes" or "no" to an integer literal such as `Literal[0, 1, 2]`, the expression `bool_value` in `literal_values` will evaluate to `True`. This causes the converter to return a boolean instead of the literal integer, violating the type hint and potentially causing downstream type errors.

A potential fix might be

```Python
if bool_value is not None:
    for v in literal_values:
        if type(v) is bool and v == bool_value:
            return bool_value
```

I added a new `timeout` parameter to the `do_deploy` method in the added example to demonstrate this bug.

Running that results in the incorrect behavior:

```sh
(Cmd) deploy prod --timeout yes
Deploying prod in safe mode with budget 1.50 and timeout True
```

I'll go ahead and push a fix for this.

tleonhardt (changes requested):

cmd2/annotated.py
```diff
@@ -0,0 +1,1196 @@
...
+
+    def __call__(
+        self,
+        _parser: argparse.ArgumentParser,
+        namespace: argparse.Namespace,
+        values: Any,
+        _option_string: str | None = None,
+    ) -> None:
+        result = values
+        if self._container_factory is not None and isinstance(values, list):
```
tleonhardt, line 533: I'm not sure this code is going to work correctly in all circumstances.

If a user applies `Argument(nargs="?")` to a collection type (like `tuple[int, ...]`), `argparse` returns a single value instead of a list. This skips the `isinstance(values, list)` check, leaving the value uncast and passing a single integer to the command function instead of the expected tuple.

tleonhardt (changes requested):

cmd2/annotated.py
KelvinChung2000: When you supply `?`, this means you want either 0 or 1 answer, which I assume would be more natural to write `a: Annotated[int | None, Argument()]` or `a: int | None`. 

But anyway, this is something I have missed. I have covered `int | None = None` but not `int | None`.

cmd2/annotated.py
```diff
@@ -0,0 +1,1242 @@
...
+    # Exact-match entries (order among these doesn't affect subclass lookup).
+    bool: _resolve_bool,
+    int: _make_simple_resolver(int),
+    float: _make_simple_resolver(float),
+    decimal.Decimal: _make_simple_resolver(decimal.Decimal),
+    list: _make_collection_resolver(list),
+    set: _make_collection_resolver(set),
+    tuple: _resolve_tuple,
+    Literal: _resolve_literal,
+}
```
tleonhardt, line 670: `Collection` is missing from `_TYPE_RESOLVES` despite being explicitly documented in `features/annotated.md`.

The documentation (docs/features/annotated.md) explicitly lists `Collection[T]` as a supported type that behaves identically to `list[T]`. However, because `collections.abc.Collection` is omitted from `_TYPE_RESOLVERS`, it fails to resolve as a collection type and is incorrectly processed as a standard scalar argument, leading to unexpected behavior or a `TypeError` when combined with `nargs='+'`. You should also remember to add `Collection` to your `collections.abc` imports at line 102.

Suggested change here:

```Python
    # issubclass fallback. enum.Enum must precede int (IntEnum <: int).
    enum.Enum: _resolve_enum,
    Path: _resolve_path,
    # Exact-match entries (order among these doesn't affect subclass lookup).
    bool: _resolve_bool,
    int: _make_simple_resolver(int),
    float: _make_simple_resolver(float),
    decimal.Decimal: _make_simple_resolver(decimal.Decimal),
    collections.abc.Collection: _make_collection_resolver(list),
    list: _make_collection_resolver(list),
    set: _make_collection_resolver(set),
    tuple: _resolve_tuple,
    Literal: _resolve_literal,
}
```

(though it would probably be better to sort all of the above in alphabetical order for maintainability).

tleonhardt (changes requested):

cmd2/annotated.py
KelvinChung2000: Collection type is something I originally had, but later I removed it since inferring a `Collection` does not really make sense, you likely want something more explicit, like `list` and `set`. I will update the documentation.

cmd2/annotated.py
```diff
@@ -0,0 +1,1242 @@
...
+        kwargs.pop("type", None)
+
+    if has_default:
+        kwargs["default"] = default
+
+    if is_kw_only and not has_default:
+        kwargs["required"] = True
+
+    # An optional positional scalar (``T | None`` without a default) takes 0-or-1 tokens.
+    if is_optional and is_positional and "nargs" not in kwargs and not kwargs.get("is_collection"):
```
tleonhardt: I think there is a subtle edge case here where positional arguments with default values will fail to parse.

If a user explicitly creates a positional argument with a default value (e.g., `arg: Annotated[str, Argument()] = "foo"`), `has_default` is `True` and `is_positional` is `True`. However, because `is_optional` is `False`, `nargs="?"` is not set. As a result, `argparse` will still strictly require the argument to be provided, breaking the intended optional nature of the parameter.

Suggested change:

```Python
if (is_optional or has_default) and is_positional and "nargs" not in kwargs and not kwargs.get("is_collection"):
```

cmd2/annotated.py
```diff
@@ -0,0 +1,1242 @@
...
+        kwargs["nargs"] = "?"
+
+    if kwargs.get("choices_provider") or kwargs.get("completer"):
+        kwargs.pop("choices", None)
+        converter = kwargs.get("type")
+        if getattr(converter, "_cmd2_strict_choice_converter", False):
```
tleonhardt: Providing a custom completer on `Literal` or `Enum` types removes type coercion, returning a raw string.

When a `completer` or `choices_provider` is added to an `Enum` or `Literal`, the strict type converter is stripped. For types like `Literal[1, 2]` or `Color.blue`, this means the command function will silently receive a `str` instead of the expected `int` or `Enum` instance. This will likely cause runtime exceptions (e.g., `AttributeError` on `.value`). Consider retaining base type coercion (like converting to `int` for `Literals`) or raising a `TypeError` / warning when attempting to combine fixed strict types with dynamic completions.

tleonhardt (changes requested):

cmd2/annotated.py
```diff
@@ -0,0 +1,1248 @@
...
+    """Best-effort type name for diagnostic messages."""
+    return tp.__name__ if hasattr(tp, "__name__") else str(tp)
+
+
+def _nargs_yields_list(nargs: Any) -> bool:
+    """Return ``True`` when an argparse ``nargs`` value produces a list at parse time.
+
+    ``nargs=1`` is included: argparse returns ``[value]``, not the bare value.
+    """
+    return nargs in ("*", "+") or (isinstance(nargs, int) and nargs >= 1)
```
tleonhardt: This function incorrectly yields `False` for `argparse.REMAINDER`.

`argparse.REMAINDER` (the string `'...'`) produces a list of remaining arguments at parse time. Because it is not included in the allowed values, a user could annotate a parameter with a scalar type (e.g., `str`) while explicitly passing `Argument(nargs=argparse.REMAINDER)`, bypassing the collection-type validation and resulting in an unexpected `list` at runtime.

Suggested fix:

```Python
return nargs in ("*", "+", argparse.REMAINDER) or (isinstance(nargs, int) and nargs >= 1)
```

cmd2/annotated.py
```diff
@@ -0,0 +1,1248 @@
...
+
+    for name, param in params:
+        if name in skip_params:
+            continue
+
+        if param.kind == inspect.Parameter.POSITIONAL_ONLY:
+            raise TypeError(
+                f"Parameter {name!r} in {func.__qualname__} is positional-only, "
+                "which is not supported by @with_annotated because parameters are passed as keyword arguments."
+            )
```
tleonhardt: `*args` and `**kwargs` parameters are mistakenly processed as positional arguments, causing runtime crashes.

Because `inspect.Parameter.VAR_POSITIONAL` (`*args`) and `inspect.Parameter.VAR_KEYWORD` (`**kwargs`) are not explicitly skipped or rejected, `_resolve_parameters` will create standard argparse positional arguments for them. When the parsed namespace is unpacked as keyword arguments (e.g., `fn(owner, **func_kwargs)`), passing a parsed argument to a function expecting `*args` will raise a `TypeError: func() got an unexpected keyword argument 'args'`.

Here is a minimal code example that demonstrates the crash:

```Python
import cmd2
from cmd2.annotated import with_annotated

class CrashApp(cmd2.Cmd):
    # The developer intents for *args to collect multiple positional arguments
    @with_annotated
    def do_crash(self, *args: str):
        self.poutput(f"Collected args: {args}")

if __name__ == '__main__':
    app = CrashApp()
    
    # Simulate the user executing the command
    print("Running: 'crash hello'")
    app.onecmd("crash hello")
```

Suggested fix (to add after above code):

```Python
if param.kind in (inspect.Parameter.VAR_POSITIONAL, inspect.Parameter.VAR_KEYWORD):
    raise TypeError(
        f"Parameter {name!r} in {func.__qualname__} uses *args or **kwargs, "
        "which is not supported by @with_annotated. "
        "Use explicitly named parameters with list/tuple types instead."
    )
```

tleonhardt (changes requested):

cmd2/annotated.py
KelvinChung2000: We technically can also infer `*args: str` as `args: tuple[str, ...]`, which I think also matches the expectation. 

For `**kwargs`, I will raise a type error, since this just does not make sense for argument parsing.

tleonhardt: Sounds like a good plan Kelvin. As long as we have documented that we infer `*args` that way it sounds like a good fit.

cmd2/annotated.py
```diff
@@ -0,0 +1,1393 @@
...
+    # Some argparse actions (e.g. count/store_true) do not accept a type converter.
+    action_name = kwargs.get("action")
+    if isinstance(action_name, str) and action_name in _ACTIONS_DISALLOW_TYPE:
+        kwargs.pop("type", None)
+
+    if has_default:
+        kwargs["default"] = default
+
+    if is_kw_only and not has_default:
+        kwargs["required"] = True
```
tleonhardt: Options created via `Option()` metadata without a default bypass type safety by defaulting to `None`.

When a required parameter (no default) is converted to an option via `Annotated[str, Option("-c")]`, `is_kw_only` is False, so `kwargs["required"] = True` is not set. Argparse makes the flag optional and passes `None` if it is omitted on the CLI. This violates the `str` type hint (since it's not `str | None`) and circumvents Python's standard requirement that parameters without defaults must be provided. Expanding the check to include any non-positional parameter without a default will enforce type safety while preserving the developer's intent.

Possible fix:

```Python
if not is_positional and not has_default:
    kwargs["required"] = True
```

tleonhardt (changes requested):

cmd2/annotated.py
KelvinChung2000: I will spend a bit more time and do a larger refactor to catch this type of leaking/edge case combination. I don't think keeping patching various if statements to reach the desired behaviour is ideal.

tleonhardt: Due to being busy with work I typically find little chunks of time of around 15 minutes or so where I can review open-source code and this is a large PR so I've reviewed bits and pieces at a time. I think we are getting close, but `argparse` is a large and complex library with a lot of edge cases.

If you are going to spend the time on a larger refactor anyways, here are a few things that are likely worth considering:

- In some cases I think your argument type hints don't match the signature for `Cmd2ArgumentParser` or our `add_argument()` patch.
    - Because of the signature mismatch, a type checker like `mypy` would give errors if you pass `rich` objects for description or epilog
    - I think you might have an overly restrictive type for `choices` as well - e.g. it is better in most cases to use `Iterable` instead of `list` 
- Support is missing for some important `argparse` features such as unless I missed it you can't set the required, title, or description fields for a subparsers group

I really appreciate the work you are putting into this! I think it is coming along nicely and will be a great addition that many of our users will love.

KelvinChung2000:
I have updated the implementation to be as declarative as possible. So, in terms of maintenance or edge cases, you can add/ban them by just adding a new rule. 

Hopefully, I have now covered most of the missing features.

KelvinChung2000:
Actually, do you want me to split the mega file and store it under something like `cmd2/annotated`? I can move out some of the "noise" like changes to potentially make things easier to review. This sort of large diff commit, GitHub is just not that well designed for it.

tleonhardt:
@KelvinChung2000 I'm flexible.  I can see an argument for keeping it all self-contained within one file. 

But I could also see the argument for splitting it into multiple files underneath the `cmd2.annotated` namespace - basically into its own sub-package.

@kmvanbrunt Would you have a preference here?

docs/features/argument_processing.md
```diff
@@ -16,18 +16,21 @@ following for you:
 1. Adds the usage message from the argument parser to your command's help.
 1. Checks if the `-h/--help` option is present, and if so, displays the help message for the command
 
-These features are all provided by the [@with_argparser][cmd2.with_argparser] decorator which is
-imported from `cmd2`.
+These features are provided by two decorators:
+
+- [@with_argparser][cmd2.with_argparser] -- build parsers manually with `add_argument()` calls
+- [@with_annotated][cmd2.annotated.with_annotated] -- build parsers automatically from type hints
```
tleonhardt: I think we should allow the decorator into the top-level namespace even though it is experimental. I think it would be a better user experience and feel more aligned with how the `with_argparser` dectorator is used.

This would involve importing it and adding it within the `__init__.py`.

Then it can just be referred to as `cmd2.with_annotated`.

cmd2/annotated.py
```diff
@@ -0,0 +1,2471 @@
...
+    self_arg: Any,
+    func_kwargs: dict[str, Any],
+    *,
+    leading_names: list[str],
+    var_positional_name: str | None,
+) -> Any:
+    """Call *func* from parsed kwargs, unpacking ``*args`` positionally when present."""
+    if var_positional_name is None:
+        return func(self_arg, **func_kwargs)
+    positional = [func_kwargs.pop(name) for name in leading_names if name in func_kwargs]
```
tleonhardt: I think there is a subtle edge case but here where this line could possibly shift positional argugments.

If a parameter listed in `leading_names` is ever missing from `func_kwargs` (e.g., dropped by a custom `ns_provider` or a bug), the `if name in func_kwargs` check silently skips it. Because Python maps the resulting `*positional` list by index, any missing element will cause `*var_values` to shift left into the wrong function parameters, leading to subtle and dangerous logic errors. 

It is safer to remove the `if` condition and fail explicitly with a `KeyError` if an expected positional parameter is missing.

Suggested change:

```Python
positional = [func_kwargs.pop(name) for name in leading_names]
```

tleonhardt (changes requested):

tleonhardt:
@KelvinChung2000 I chatted with @kmvanbrunt - for now please leave everything in the one big file `annotated.py`.  Later if things grow, we can consider breaking up into multiple files.

tleonhardt:
@KelvinChung2000 I pushed a change to add `@with_annotated` to the top level namespace after chatting with @kmvanbrunt about it. 

I also pushed a fix for the minor edge case bug I mentioned last time.

From my perspective, this is good to merge.  If anyone has any further comments or recommendations, put them forth sooner rather than later. I'll merge this a day or two from now if there are no further suggestions.

tleonhardt (dismissed):

tleonhardt (dismissed):

tleonhardt (approved):

tleonhardt:
@KelvinChung2000 Thank you for bearing with us on the long road to get here and congrats on getting your first `cmd2` feature merged in!

PR #1668, opened 2026-05-23
Added monkey patch for Rich AnsiDecoder.decode() bug.

### LINKED ISSUE #1667: Add cmd2 monkeypatch for bug introduced in rich 15.0.0
When `rich` fixed a bug in `Text.from_ansi()` that we reported, they introduced [another bug](https://github.com/Textualize/rich/issues/4090). We should add a monkeypatch within cmd2 to work around this issue until rich fixes it on their end.

People have submitted PRs for this rich bug twice and the sole maintainer has immediately closed them. For one he explicitly said it was for violating AI policy but he didn't give a reason for the other one.

kmvanbrunt:
Fixes #1667

### DIFF
--- a/tests/test_rich_utils.py
+++ b/tests/test_rich_utils.py
@@ -317,3 +317,28 @@ def test_formatter_set_color(mocker: MockerFixture) -> None:
     assert mock_set_color.call_count == 2
     mock_set_color.assert_any_call(True, file=sys.stdout)
     mock_set_color.assert_any_call(True)
+
+
+@pytest.mark.parametrize("line_break", ["\n", "\r\n"])
+def test_ansi_decoder_patch(line_break: str) -> None:
+    # Check if we are still patching AnsiDecoder.decode(). If this check fails, then Rich
+    # has fixed the bug. Therefore, we can remove this test function and ru._AnsiDecoder_decode.
+    from rich.ansi import AnsiDecoder
+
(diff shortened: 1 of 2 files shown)


tests/test_rich_utils.py
```diff
@@ -317,3 +317,28 @@ def side_effect(color: bool, **kwargs: Any) -> None:
...
     mock_set_color.assert_any_call(True)
+
+
+@pytest.mark.parametrize("line_break", ["\n", "\r\n"])
+def test_ansi_decoder_patch(line_break: str) -> None:
+    # Check if we are still patching AnsiDecoder.decode(). If this check fails, then Rich
+    # has fixed the bug. Therefore, we can remove this test function and ru._AnsiDecoder_decode.
+    from rich.ansi import AnsiDecoder
+
+    assert AnsiDecoder.decode is ru._AnsiDecoder_decode
```
tleonhardt, line 328: This will break downstream consumers that run our test suite such as Linux distros once `rich` fixes the bug if they run tests before we remove the monkeypatch.  I guess in theory that might be ok as long as we are quick.  

But would something like this be better to avoid creating any unnecessary problems for Linux distros that package `cmd2`?

```Python
if AnsiDecoder.decode is not ru._AnsiDecoder_decode:
    import warnings
    warnings.warn("Rich has fixed the CRLF bug! The monkey patch can now be removed.")
```

tleonhardt (approved):

PR #1678, opened 2026-06-06
tests: make tests compatible with rich 14.3.3

mcepl:
It seems that with the upgrade to rich 14.3.3 (what we have in openSUSE/Tumbleweed) `stylize` function eats EOLs. I have modified tests to acommodate that, and I hope it should be universal enough so that even with older versions of rich nothing should break.

### DIFF
--- a/tests/test_cmd2.py
+++ b/tests/test_cmd2.py
@@ -3692,7 +3692,7 @@ def test_ansi_pouterr_always_tty(mocker, capsys) -> None:
     mocker.patch.object(sys.stderr, "isatty", return_value=True)
 
     expected_plain = "oopsie\n"
-    expected_styled = su.stylize("oopsie\n", Cmd2Style.ERROR)
+    expected_styled = su.stylize('oopsie', Cmd2Style.ERROR) + '\n'
 
     app.onecmd_plus_hooks("echo_error oopsie")
     out, err = capsys.readouterr()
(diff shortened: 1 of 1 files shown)


tleonhardt:
@mcepl What version of `cmd2` are you packaging in openSUSE/Tumbleweed? I ask because there have been a lot of little changes and bug fixes in recent versions of `rich` and cmd2 4.0.0 requires `rich >= 15.0.0`. Even cmd2 3.5.1, the latest version in the 3.x line, also requires `rich >= 15.0.0`.

@kmvanbrunt is intimately familiar with the recent `rich` releases and what bug fixes were in each, so he is the one that should review this as a potential change.

Regardless, you could need to run `ruff format` first to comply with our auto-formatting rules. We have a `make check` convenience command built into our repo which runs all of our auto-formatting and linting checks using `prek`.

kmvanbrunt:
The issue isn't with `cmd2`. Rich fixed a [bug](https://github.com/Textualize/rich/pull/4076) in version 15.0.0 where `Text.from_ansi()` didn't preserve all trailing newlines. Prior to their fix, `cmd2` included a monkey patch that resolved this bug.

You are using a setup with a new version of `cmd2` (no monkey patch) and an old version of Rich (14.3.3) which does not include the fix. It's an incompatible configuration. As @tleonhardt mentioned, `cmd2` requires `rich >=15.0.0`.

mcepl:
On Sat Jun 6, 2026 at 9:06 PM CEST, Kevin Van Brunt wrote:
> You are using a setup with a new version of `cmd2` (no monkey patch) and an old version of Rich (14.3.3) which does not include the fix. It's an incompatible configuration. ***@***.*** mentioned, `cmd2` requires `rich >=15.0.0`. 

Yes, you are right, with rich 15.0.0 the test suite passes. You can close this ticket.

PR #1692, opened 2026-06-10
fix: fix annotated typing

### LINKED ISSUE #1684: Mypy complains about method decorated with `with_annotated` decorator
# Problem

Mypy complains about methods decorated with `with_annotated` when using strict settings.

## How to reproduce

```python
class App(Cmd):

    @with_annotated
    def do_something(...) -> None:  # <- Mypy complains here
        ...
```

## Expected behavior

Mypy does not complain.

## Actual behavior

Mypy reports:

```
...: error: Type of decorated function contains type "Any" ("Callable[..., Any]")  [misc]
        def do_something(...)
```

I think this can be fixed by using `ParamSpec` instead of the ellipsis notation (`Callable[..., Any]`). I have not verified it yet.

KelvinChung2000:
close #1684

### DIFF
--- a/cmd2/annotated.py
+++ b/cmd2/annotated.py
@@ -190,6 +190,8 @@ from typing import (
     Any,
     ClassVar,
     Literal,
+    ParamSpec,
+    Protocol,
     TypedDict,
     TypeVar,
     Union,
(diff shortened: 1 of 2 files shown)


cmd2/annotated.py
```diff
@@ -2136,8 +2138,18 @@ def handler(self_arg: Any, ns: Any) -> Any:
...
 
 
+_CommandParams = ParamSpec("_CommandParams")
+_CommandReturn = TypeVar("_CommandReturn")
+
+
+class _WithAnnotatedDecorator(Protocol):
+    """The signature-preserving decorator ``with_annotated(...)`` returns (generic per call)."""
+
+    def __call__(self, fn: "Callable[_CommandParams, _CommandReturn]", /) -> "Callable[_CommandParams, _CommandReturn]": ...
```
tleonhardt: Unnecessary string annotations for Callable can cause type variable binding issues.

Since `Callable`, `_CommandParams`, and `_CommandReturn` are already defined, there is no need to use string forward references here. Unquoted types ensure better compatibility and correct type variable binding across different type checkers (like `mypy` or `pyright`).

Recommend:
```Python
def __call__(self, fn: Callable[_CommandParams, _CommandReturn], /) -> Callable[_CommandParams, _CommandReturn]:
```

cmd2/annotated.py
```diff
@@ -2136,8 +2138,18 @@ def handler(self_arg: Any, ns: Any) -> Any:
     return handler, subcmd_name, parser_builder
 
 
+_CommandParams = ParamSpec("_CommandParams")
```
tleonhardt, line 2142: Recommend adding a brief bug fix blurb to the `CHANGELOG.md` file about fixing type hinting for this stuff.

neoniobium:
I check out this PR. It works for me. Thank you very much!

tleonhardt (approved):

PR #1698, opened 2026-06-25
feat: blocked argument

### LINKED ISSUE #1689: Template functionality for argument blocks using `with_annotaed` decorator
# Feature Request

To avoid redundant code, add a template mechanism for argument blocks that appear in multiple commands. Using the `with_argparse` decorator this is easy by reusing a function that modifies a parser; `with_annotated` currently has no simple equivalent.

## Possible solutions

- Use a TypedDict + Unpack to reuse a named block of annotations:
```python
class ArgumentBlock(TypedDict):
    # define reusable keys/values here
    ...

class App(Cmd):

    @with_annotated
    def do_a(..., **kwargs: Unpack[ArgumentBlock]) -> None:
        ...

    @with_annotated
    def do_b(..., **kwargs: Unpack[ArgumentBlock]) -> None:
        ...
```

- Allow attaching a predefined parser to the parser generated by `with_annotated`, and expose the resulting namespace inside the decorated method for easy access.

Both approaches reduce duplication; the second may be simpler to implement.

### LINKED ISSUE #1690: Is there an intended way to pass  modifed or added arguments from a parent parser to a one or more subparsers when using the `with_annotated` decorator?
# Question / Feature Request

Is there an intended way to modify, add, or pass arguments from a parent parser to a subparser when using the `with_annotated` decorator?

With the `with_argparse` approach this is straightforward because you have access to the full namespace.

Example:

```python
class App(Cmd):

    @with_annotated(subcommand_to=parent)
    def do_sub(...) -> None:
        ...  # use possibly modified parent_argument here

    @with_annotated(base_command=True)
    def do_parent(parent_argument: ...) -> None:
        ...  # possibly modify parent_argument here
```

I would like to use `parent_argument` (defined on the parent parser) in `do_sub` either as an argument *hidden* from the subparser or by accessing the full namespace of the parser.

Is `ns_provider` intended for this use case? If so I cannot see how.

KelvinChung2000:
close #1689 
close #1690

### DIFF
--- a/cmd2/annotated.py
+++ b/cmd2/annotated.py
@@ -145,6 +145,44 @@ so forward-referenced annotations still decorate -- meaning a misconfigured grou
 class is defined rather than on first command use.  The one group rule that needs the annotations
 (a required member in a mutually exclusive group) fires when the parser is built.
 
+A parameter annotated with a ``@dataclass`` that subclasses [`ArgumentBlock`][cmd2.annotated.ArgumentBlock] is a
+reusable *argument block*: each of the dataclass's ``init`` fields is expanded into a flat command-line
+argument (field name == argument name) at the parameter's position, and the parsed values are
+reconstructed into a dataclass instance passed to the command.  This lets several commands share one
+block without duplicating parameters, and a base block shared by inheritance is the reuse mechanism
+(a subclass of a block is itself a block):
+
+    @dataclass
+    class CommonArgs(cmd2.ArgumentBlock):
+        verbose: Annotated[bool, Option("-v", "--verbose")] = False
+        output: Annotated[Path | None, Option("--output")] = None
+
+
+    class MyApp(cmd2.Cmd):
+        @cmd2.with_annotated
+        def do_build(self, target: str, common: CommonArgs):
+            if common.verbose:
+                self.poutput(common.output)
+
+The [`ArgumentBlock`][cmd2.annotated.ArgumentBlock] trait -- not "is a dataclass" -- is the trigger, so a plain
+``@dataclass`` is never expanded and can still be used as an ordinary single argument value (e.g. with an
+``Argument(converter=...)``).  Each field carries the usual ``Annotated[T, Option(...)]`` /
+``Annotated[T, Argument(...)]`` metadata and behaves exactly as a top-level parameter of the same shape
+(type inference, completion, choices).  The dataclass is the single source of truth for defaults: a field
+with a default (``default`` or ``default_factory``) is emitted with ``argparse.SUPPRESS`` and filled by the
+dataclass constructor at reconstruction time, so a ``default_factory`` yields a fresh value per invocation
+(no shared-mutable default) and ``__post_init__`` runs.  A field with no default becomes a required
+argument.  A field whose type is itself a block is not expanded (no recursion) and raises the
+unsupported-type error.  Because fields expand flat, every field name must be unique across the command's
+own parameters and every other block's fields -- a collision (which would put two argparse actions on one
+namespace dest) raises ``TypeError`` when the parser is built.  Block fields cannot participate in
+``groups=`` / ``mutually_exclusive_groups=`` (those reference the function's own parameter names, which are
+validated without resolving type hints).  A block must be the *bare* annotation of a regular parameter:
+wrapping it in ``Annotated``/``Optional``/a union, or using it as ``*args`` / ``**kwargs``, raises
+``TypeError`` (to use a dataclass as a single value instead, make it a plain ``@dataclass`` with an
+``Argument(converter=...)``).  An ``ArgumentBlock`` subclass that is not a ``@dataclass`` has no fields and
+also raises ``TypeError``.
+
 Unsupported patterns (raise ``TypeError``):
 
 - a non-Optional type with a ``None`` default (e.g. ``name: str = None``); annotate it
@@ -229,8 +267,11 @@ from collections.abc import (
     Sequence,
 )
 from dataclasses import (
+    MISSING,
     dataclass,
     field,
+    fields,
+    is_dataclass,
 )
 from pathlib import Path
 from typing import (
@@ -238,9 +279,11 @@ from typing import (
     Any,
     ClassVar,
     Literal,
+    NamedTuple,
     ParamSpec,
     Protocol,
     TypedDict,
+    TypeGuard,
     TypeVar,
     Union,
     Unpack,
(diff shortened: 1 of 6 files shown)


cmd2/annotated.py
```diff
@@ -1940,14 +2052,243 @@ def _link_mutex_group_membership(
...
+    """
+    if not is_dataclass(dc_type):
+        raise TypeError(
+            f"ArgumentBlock subclass '{dc_type.__name__}' must be decorated with @dataclass so it has fields "
+            f"to expand into command-line arguments."
+        )
+    try:
+        field_hints = get_type_hints(dc_type, include_extras=True)
+    except (NameError, AttributeError, TypeError) as exc:
+        raise TypeError(
```
tleonhardt, line 2207: Please add a test to cover this `TypeError` case

cmd2/annotated.py
```diff
@@ -1940,14 +2052,243 @@ def _link_mutex_group_membership(
...
+    init_only = [name for name in list(inspect.signature(dc_type.__init__).parameters)[1:] if name not in init_field_names]
+    if init_only:
+        raise TypeError(
+            f"ArgumentBlock '{dc_type.__name__}' has InitVar field(s) {sorted(init_only)}, which @with_annotated "
+            f"cannot expand into command-line arguments; use a regular field instead."
+        )
+    expanded: list[_ArgparseArgument] = []
+    for f in fields(dc_type):
+        if not f.init:
+            continue  # field(init=False) is not a constructor argument, so it has no command-line value
```
tleonhardt, line 2224: Please add a test to cover this line

tleonhardt (changes requested):

cmd2/annotated.py
```diff
@@ -1940,14 +2052,243 @@ def _link_mutex_group_membership(
...
+    ``cmd2_base_args`` block: that parent stamps a presence marker into the shared namespace at parse time,
+    so its absence here means no ancestor declared the block.  Reconstructing anyway would hand back an
+    all-defaults instance and hide the misconfiguration, so raise instead.
+    """
+    for block_name, spec in blocks.items():
+        # The expanded field values always come out of func_kwargs -- they are not parameters of the command
+        # function (the block instance is), so leaving them in would crash the call with an unexpected keyword.
+        # Pop them up front, before the already-supplied/inherited checks, so a directly-supplied instance does
+        # not strand command-line field values in func_kwargs.
+        field_values = {name: func_kwargs.pop(name) for name in spec.field_names if name in func_kwargs}
```
tleonhardt: Programmatically supplied block instances can be accidentally passed to the dataclass constructor if a field shares the block's parameter name.

In `_reconstruct_dataclass_blocks`, if an `ArgumentBlock` has a field with the exact same name as the function parameter that accepts the block (e.g., `def do_x(self, opts: Opts)` where `Opts` has a field named `opts`), and a caller passes the block instance directly via `kwargs` (e.g., `do_x("cmd",
  opts=Opts(...))`), the block instance will be mistakenly popped as a field value and passed into the dataclass constructor, resulting in a `TypeError`. You must safely discard stray command-line field values without popping the provided block instance itself.

A change might look something like ...

```python
for block_name, spec in blocks.items():
    if block_name in func_kwargs:
        # A block instance is already present (e.g. a programmatic call passing it directly); use it as-is.
        # Pop any stray command-line field values so they don't crash the call with an unexpected keyword.
        for name in spec.field_names:
            if name != block_name:
                func_kwargs.pop(name, None)
        continue

    # The expanded field values always come out of func_kwargs -- they are not parameters of the command
    # function (the block instance is), so leaving them in would crash the call with an unexpected keyword.
    field_values = {name: func_kwargs.pop(name) for name in spec.field_names if name in func_kwargs}
    if spec.inherited and not getattr(ns, _base_args_marker_dest(spec.dc_type), False):
...
```

tleonhardt (changes requested):

cmd2/annotated.py
KelvinChung2000: I go with blocking instead of accepting. Since this sort of collision might have other side effects.

cmd2/annotated.py
```diff
@@ -1844,11 +1955,26 @@ def _const_mismatches_type(a: _ArgparseArgument) -> bool:
             f"annotate the type as '{_type_name(a.inner_type)} | None'."
         ),
     ),
+    (
+        # A block field's default must live on the dataclass field: a field default emits SUPPRESS (see
+        # _DEFAULT_RULES) so the dataclass constructor produces the value fresh on every call.
+        lambda a: a.is_block_field and not a.has_default and a.default is not _UNSET and a.default is not None,
```
tleonhardt: `a.default is not None` allows `Option(default=None)` to bypass the rule that block defaults must live on the dataclass.

By explicitly ignoring `None`, this rule allows a user to configure `Option(default=None)` on an
`ArgumentBlock` field that has no native dataclass default. This will silently parse as `None` and inject
`None` into the dataclass constructor. While `None `doesn't risk shared-mutable state, it completely circumvents the strict design intent that the dataclass is the single source of truth for defaults, and it could allow `None` to flow into a strictly typed field (e.g., `count: int`). Removing `and a.default is not None` ensures uniformly enforced, native dataclass default handling.

Recommend:
```python
lambda a: a.is_block_field and not a.has_default and a.default is not _UNSET,
```

cmd2/annotated.py
```diff
@@ -1940,14 +2089,276 @@ def _link_mutex_group_membership(
...
+            globalns=getattr(unwrapped, "__globals__", {}),
+            include_extras=True,
+        )
+    except (NameError, AttributeError, TypeError) as exc:
+        raise TypeError(
+            f"Failed to resolve type hints for {func.__qualname__}. Ensure all annotations use valid, importable types."
+        ) from exc
+
+
+def _is_argument_block(hint: Any, param: inspect.Parameter) -> TypeGuard[type]:
```
tleonhardt: Because this type guard explicitly uses `issubclass(hint, ArgumentBlock)`, we should be able to narrow the type guard to 

```python
def _is_argument_block(hint: Any, param: inspect.Parameter) -> TypeGuard[type[ArgumentBlock]]:
```

cmd2/annotated.py
```diff
@@ -1940,14 +2089,276 @@ def _link_mutex_group_membership(
...
+    which :func:`_expand_dataclass_block` enforces with a clear message.
+    """
+    return (
+        param.kind in (inspect.Parameter.POSITIONAL_OR_KEYWORD, inspect.Parameter.KEYWORD_ONLY)
+        and isinstance(hint, type)
+        and issubclass(hint, ArgumentBlock)
+    )
+
+
+def _require_magic_block(name: str, hint: Any, param: inspect.Parameter, func_qualname: str) -> type:
```
tleonhardt: Similarly, I believe the return type here can be narrowed:

```python
def _require_magic_block(name: str, hint: Any, param: inspect.Parameter, func_qualname: str) -> type[ArgumentBlock]:
```

cmd2/annotated.py
```diff
@@ -1940,14 +2089,276 @@ def _link_mutex_group_membership(
...
+                dest_override=_shared_field_dest(dc_type, f.name) if shared else None,
+            )
+        )
+    return expanded
+
+
+class _BlockSpec(NamedTuple):
+    """How to reconstruct one dataclass-block parameter from the parsed namespace at call time."""
+
+    dc_type: type
```
tleonhardt: I think the type for this can be narrowed to:

```python
dc_type: type[ArgumentBlock]
```

There are about 5 or so other places in this file where we have a `dc_type: type` as a function parameter or whatnot. We should be able to narrow the type hint in all cases to `type[ArgumentBlock`.

tleonhardt (changes requested):

tleonhardt (approved):

PR #1700, opened 2026-06-27
Refactored bottom toolbar and rprompt functionality.

### LINKED ISSUE #1699: Make use of refresh_interval parameter for prompt-toolkit sessions
In `prompt-toolkit`, you can control session and UI refresh rates by configuring the `refresh_interval` in your Application / Session constructor, which invalidates and redraws the UI automatically.

We should add this as a parameter to the `cmd2.Cmd.__init__` method with a default value.  The default value should probably match whatever the default value from `prompt-toolkit` is.

kmvanbrunt:
- Removed default implementation of get_bottom_toolbar() from Cmd class and moved it to the getting_started example.
- Renamed bottom_toolbar flag to enable_bottom_toolbar.
- Added enable_rprompt flag.

Also closes #1699

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -2006,8 +2006,8 @@ class Cmd:
 
         This method is called by prompt-toolkit while at the main prompt if ``enable_bottom_toolbar``
         was set to ``True`` during initialization. Because prompt-toolkit executes this callback
-        on every UI refresh (such as on every keypress or timed interval), keeping this function
-        highly optimized is critical to ensuring the CLI remains responsive.
+        on every UI refresh (such as on every keypress or at scheduled refresh intervals), keeping
+        this function highly optimized is critical to ensuring the CLI remains responsive.
 
         Override this if you want a bottom toolbar displaying contextual information useful for
         your application. This could be information like the application name, current state,
@@ -2022,8 +2022,8 @@ class Cmd:
 
         This method is called by prompt-toolkit while at the main prompt if ``enable_rprompt``
         was set to ``True`` during initialization. Because prompt-toolkit executes this callback
-        on every UI refresh (such as on every keypress or timed interval), keeping this function
-        highly optimized is critical to ensuring the CLI remains responsive.
+        on every UI refresh (such as on every keypress or at scheduled refresh intervals), keeping
+        this function highly optimized is critical to ensuring the CLI remains responsive.
 
         Override this if you want a right prompt displaying contextual information useful for
         your application. This could be information like the current Git branch, time, or current
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -370,6 +370,8 @@ class Cmd:
         auto_suggest: bool = True,
         complete_in_thread: bool = True,
         command_sets: Iterable[CommandSet[Any]] | None = None,
+        enable_bottom_toolbar: bool = False,
+        enable_rprompt: bool = False,
         include_ipy: bool = False,
         include_py: bool = False,
         intro: RenderableType = "",
@@ -410,6 +412,10 @@ class Cmd:
                              This allows CommandSets with custom constructor parameters to be
                              loaded.  This also allows the a set of CommandSets to be provided
                              when `auto_load_commands` is set to False
+        :param enable_bottom_toolbar: if ``True``, enables a bottom toolbar while at the main prompt.
+                                      Override ``get_bottom_toolbar()`` to define its content.
+        :param enable_rprompt: if ``True``, enables a right prompt while at the main prompt.
+                               Override ``get_rprompt()`` to define its content.
         :param include_ipy: should the "ipy" command be included for an embedded IPython shell
         :param include_py: should the "py" command be included for an embedded Python shell
         :param intro: introduction to display at startup
@@ -534,9 +540,14 @@ class Cmd:
         self._initialize_history(persistent_history_file)
 
         # Create the main PromptSession
-        self.complete_in_thread = complete_in_thread
-        self.refresh_interval = refresh_interval
-        self.main_session = self._create_main_session(auto_suggest, completekey)
+        self.main_session = self._create_main_session(
+            auto_suggest=auto_suggest,
+            complete_in_thread=complete_in_thread,
+            completekey=completekey,
+            enable_bottom_toolbar=enable_bottom_toolbar,
+            enable_rprompt=enable_rprompt,
+            refresh_interval=refresh_interval,
+        )
 
         # The session currently holding focus (either the main REPL or a command's
         # custom prompt). Completion and UI logic should reference this variable
@@ -727,7 +738,16 @@ class Cmd:
                 # No macro found or already processed. The statement is complete.
                 return False
 
-    def _create_main_session(self, auto_suggest: bool, completekey: str) -> PromptSession[str]:
+    def _create_main_session(
+        self,
+        *,
+        auto_suggest: bool,
+        complete_in_thread: bool,
+        completekey: str,
+        enable_bottom_toolbar: bool,
+        enable_rprompt: bool,
+        refresh_interval: float,
+    ) -> PromptSession[str]:
         """Create and return the main PromptSession for the application.
 
         Builds an interactive session if self.stdin and self.stdout are TTYs.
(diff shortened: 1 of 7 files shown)


examples/getting_started.py
```diff
@@ -87,11 +88,33 @@ def __init__(self) -> None:
             )
         )
 
-    def get_rprompt(self) -> str | FormattedText | None:
+    def get_bottom_toolbar(self) -> AnyFormattedText:
+        # Get the current time in ISO format with 0.01s precision
+        dt = datetime.datetime.now(datetime.timezone.utc).astimezone()
+        now = dt.strftime("%Y-%m-%dT%H:%M:%S.%f")[:-4] + dt.strftime("%z")
+        left_text = sys.argv[0]
```
tleonhardt, line 99: Since `sys`, is now being used by this class, the import should be moved to the top of the file instead of being within the `__main__` block below.

docs/features/prompt.md
```diff
@@ -65,22 +65,11 @@ terminal window while the application is idle and waiting for input.
...
-
-### Customizing Toolbar Content
-
-You can customize the content of the toolbar by overriding the [cmd2.Cmd.get_bottom_toolbar][]
-method. This method should return either a string or a list of `(style, text)` tuples for formatted
-text.
-
-```py
-    def get_bottom_toolbar(self) -> list[str | tuple[str, str]] | None:
+    def get_bottom_toolbar(self) -> AnyFormattedText:
```
tleonhardt, line 84: Missing import statement for `AnyFormattedText` in this documentation example.

tleonhardt (changes requested):
Other than the two very minor issues I called out, I don't have problems with the code changes per se. 

However, I have a big problem with the new default behavior from a UX perspective. Before this change, when a bottom toolbar wasn't requested, there was no visual display of a bottom bar. Now, it displays an empty gray bar which is ugly, distracting, and pointless.

Please fix it so the default behavior is like it was previously.

docs/features/prompt.md
kmvanbrunt: Done

examples/getting_started.py
kmvanbrunt: Done

docs/features/initialization.md
```diff
@@ -42,6 +41,8 @@ Here are instance attributes of `cmd2.Cmd` which developers might wish to overri
 - **disabled_commands**: commands that have been disabled from use. This is to support commands that are only available during specific states of the application. This dictionary's keys are the command names and its values are DisabledCommand objects.
 - **echo**: if `True`, each command the user issues will be repeated to the screen before it is executed. This is particularly useful when running scripts. This behavior does not occur when running a command at the prompt. (Default: `False`)
 - **editor**: text editor program to use with _edit_ command (e.g. `vim`)
+- **enable_bottom_toolbar**: if `True`, enables a bottom toolbar while at the main prompt. (Default: `False`)
```
tleonhardt: The newly added `enable_bottom_toolbar` and `enable_rprompt` (as well as `refresh_interval` below) are documented here as **instance attributes**, meaning they are instance members of the `cmd2.Cmd` instance that developers can override at runtime. However, in `cmd2.py`, they are strictly initializer arguments passed directly to `PromptSession` and are not stored on the `Cmd` instance. Attempting to override them at runtime will have no effect.

I think we should create a new section documenting `__init__` parameters which are not also instance attributes and move them there.

This also applies to `complete_in_thread`.

Alternatively, we could restore them to instance attributes stored in `Cmd`.

cmd2/cmd2.py
```diff
@@ -535,10 +540,14 @@ def __init__(
         self._initialize_history(persistent_history_file)
 
         # Create the main PromptSession
-        self.bottom_toolbar = bottom_toolbar
-        self.complete_in_thread = complete_in_thread
```
tleonhardt, line 539: Since `complete_in_thread`, `refresh_interval`, and `bottom_toolbar` were previously public instance attributes on` Cmd` (and documented as such), removing them from the object instance will break downstream code that attempts to access or modify them. This should be explicitly noted in the breaking changes section of `CHANGELOG.md`, e.g.:

```markdown
- Breaking Changes
    - ...
    - Removed `bottom_toolbar`, `complete_in_thread`, and `refresh_interval` as instance attributes of `Cmd`. They are now strictly `__init__` parameters.
```

Alternatively, we could restore them to being instance attributes.

tests/test_cmd2.py
```diff
@@ -4288,34 +4286,63 @@ def test_path_complete_users_windows(monkeypatch, base_app):
...
-    base_app.bottom_toolbar = True
-    monkeypatch.setattr(sys, "argv", ["myapp.py"])
-    toolbar = base_app.get_bottom_toolbar()
-    assert isinstance(toolbar, list)
-    assert toolbar[0] == ("ansigreen", "myapp.py")
-    assert toolbar[2][0] == "ansicyan"
+    # Test overridden
+    expected_text = "bottom toolbar text"
+    base_app.get_bottom_toolbar = lambda: expected_text
+    assert base_app.get_bottom_toolbar() == expected_text
```
tleonhardt: Overriding `base_app.get_bottom_toolbar` with a lambda and then asserting it returns the lambda's value only tests Python's ability to assign a function. Since the method no longer contains default logic, testing that the default implementation returns `None` is sufficient.

tests/test_cmd2.py
```diff
@@ -4288,34 +4286,63 @@ def test_path_complete_users_windows(monkeypatch, base_app):
...
 def test_get_rprompt(base_app):
     # Test default
     assert base_app.get_rprompt() is None
 
     # Test overridden
-    from prompt_toolkit.formatted_text import FormattedText
-
     expected_text = "rprompt text"
     base_app.get_rprompt = lambda: expected_text
     assert base_app.get_rprompt() == expected_text
```
tleonhardt: Similar to `test_get_bottom_toolbar`, overriding `get_rprompt` with a lambda and asserting its return value tests Python's assignment capabilities rather than actual cmd2 functionality.

tleonhardt (changes requested):

tests/test_cmd2.py
kmvanbrunt: Done

tests/test_cmd2.py
kmvanbrunt: Done

docs/features/initialization.md
kmvanbrunt: I think restoring them as instance attributes is confusing since changing them after `__init__()` runs has no effect. If a dev wants to alter prompt behavior later, they can call `self._create_main_session()` with different values for these options.

cmd2/cmd2.py
kmvanbrunt: I don't need to mention `refresh_interval` since it did not exist in `cmd2` 4.0.

kmvanbrunt: Done

docs/features/initialization.md
kmvanbrunt: I'm not sure what you mean by a new section in `initialization.md`. That file already contains a table documenting all `__init__` parameters.

tleonhardt: In retrospect, what we have is probably fine.

examples/getting_started.py
```diff
@@ -87,11 +92,33 @@ def __init__(self) -> None:
             )
         )
 
-    def get_rprompt(self) -> str | FormattedText | None:
+    def get_bottom_toolbar(self) -> AnyFormattedText:
```
tleonhardt, line 95: The documentation for `get_bottom_toolbar` correctly warns that this method is called on every UI refresh. However, the implementation in this example performs relatively heavy operations inside this callback and as such I'm worried we are creating somewhat of a bad example for others to follow.

`get_app().output.get_size()` and `datetime.datetime.now()` are called every single time the user presses a key. In a high-frequency typing scenario, this might introduce micro-stutter if the terminal resize logic or time fetching becomes a bottleneck.

I think this is acceptable as an example for now, but we should improve this in a future PR. We should probably have something like a background thread that updates variables with this information (protected by a lock). Then the `get_bottom_toolbar` should just format that info for display.

tleonhardt (approved):

examples/getting_started.py
tleonhardt: I created Issue #1702 to address this.

PR #1703, opened 2026-07-01
Refactor get_bottom_toolbar implementation in getting_started.py example

### LINKED ISSUE #1702: Refactor get_bottom_toolbar implementation for getting_started.py example
The documentation for `get_bottom_toolbar` correctly warns that this method is called on every UI refresh. However, the implementation in the `getting_started.py` example performs relatively heavy operations inside this callback and as such I'm worried we are creating somewhat of a bad example for others to follow.

`get_app().output.get_size()` and `datetime.datetime.now()` are called every single time the user presses a key. In a high-frequency typing scenario, this might introduce micro-stutter if the terminal resize logic or time fetching becomes a bottleneck.

We should probably have something like a background thread that updates variables with this information (protected by a lock). Then the `get_bottom_toolbar` callback should just format that info for display.

tleonhardt:
This refactors the implementation for the `get_bottom_toolbar` method in the `getting_started.py` example to use a background thread to do the more expensive operation of getting and formatting the current time and caching that in a local state variable so this isn't done with every key press.

This transitions the `getting_started.py` example from being a simple intro and starts turning it into something more representative of a rugged production-ready application.

Closes #1702 

Also:
- Made a couple unrelated changes to support using the [superpowers](https://github.com/obra/superpowers) agent skills framework so it doesn't commit the spec, plan, or progress Markdown files it generates while it does it's work.


examples/getting_started.py
```diff
@@ -15,11 +15,14 @@
 11) Shortcuts for commands
 12) Persistent bottom toolbar with realtime status updates
 13) Right prompt which displays contextual information
+14) Using a background thread to update the bottom toolbar in an efficient manner
```
kmvanbrunt: This comment could probably be more clear. It's really updating the content displayed by the bottom toolbar. And it does so outside of the UI loop to keep things responsive.

kmvanbrunt (changes requested):

examples/getting_started.py
tleonhardt: Clarified

kmvanbrunt (approved):

PR #1705, opened 2026-07-01
getting_started.py now demonstrates how to use common cmd2 deocrators

### LINKED ISSUE #1704: Refactor getting_started.py example to demonstrate best-practices for defining command arguments
The recommended way to define command arguments in `cmd2` is to use either the `@with_argparser` or `@with_annotated` decorator. 

The `getting_startedlpy` example is our best "Introduction to cmd2" example. As such, it should be fully displaying recommended best practices. We should refactor it to use both of these decorators in different commands.

tleonhardt:
`getting_started.py` now demonstrates how to use both the `@with_argparser` and `@with_annotated` decorators for defining command arguments.

The intent is to help make the `getting_started.py` example something of a one-stop shop to quickly orient new `cmd2` users to what we view as best practices. It was previously parsing raw `Statements` for arguments, which frankly I view as an antipattern.

Closes #1704


tleonhardt:
@kmvanbrunt I would appreciate any constructive feedback on how to improve the `@with_argparser` example here to make it more educational for new users.

@KelvinChung2000 I would appreciate any constructive feedback on how to improve the `@with_annotated` example here to make it more educational for new users.

examples/getting_started.py
```diff
@@ -156,18 +161,49 @@ def get_rprompt(self) -> AnyFormattedText:
...
+        if interactive:
+            self.poutput(
+                stylize(
+                    "Interactive mode enabled! (Simulated interactive setup)",
+                    style=Style(color=Color.YELLOW.value),
+                )
+            )
+
+    # do_echo parser
+    echo_parser = cmd2.Cmd2ArgumentParser(description="Multiline command that echoes input.")
```
kmvanbrunt: I prefer creating a command parser with a factory rather than making it a class member. A parser factory can be a `staticmethod`, `classmethod`, or even a function outside of the class. All commands in `cmd2.py` use a factory.

tleonhardt: I created a factory method for this

KelvinChung2000:
I recommend demonstrating an example with a positional `Path` type. 

```py
@cmd2.with_annotated
def do_cat(
    self,
    path: pathlib.Path,
    numbered: Annotated[
        bool, Option("-n", "--number", help_text="prefix each line with its number")
    ] = False,
) -> None:
    """Print a file's contents. `path` tab-completes filesystem paths automatically.

    Try:
        cat <TAB>              # path completes files/dirs -- no completer wired
        cat notes.txt
        cat notes.txt -n       # -n / --number, declared via Option metadata
        cat notes.txt --no-number
    """
    for index, line in enumerate(path.read_text().splitlines(), start=1):
        self.poutput(f"{index}: {line}" if numbered else line)
```

This demonstrates auto-completer inference and non-built-in type support while being simple and explicit. But this does not fit a "intro" command very well.

tleonhardt:
@KelvinChung2000 I took your advice and added a modified version of your `cat` command example for demonstrating use of the `@with_annotated` decorator. 

I also reverted the `do_intro` command logic to its previous state.

PR #1706, opened 2026-07-07
Fixed cmd2 bypassing NO_COLOR and allow_style when setting prompt-toolkit's color depth.

kmvanbrunt:


cmd2/cmd2.py
```diff
@@ -1443,6 +1443,7 @@ def allow_style(self) -> ru.AllowStyle:
     def allow_style(self, value: ru.AllowStyle) -> None:
         """Setter property needed to support do_set when it updates allow_style."""
         ru.ALLOW_STYLE = value
+        self.main_session.color_depth = pt_resolve_color_depth()
```
tleonhardt, line 1446: It is a common pattern for subclasses of `cmd2.Cmd` to configure properties like `self.allow_style` in their `__init__` before calling `super().__init__()`. Doing so here will crash the application because `main_session` has not been instantiated yet. Using `getattr` would makes this setter safe at any point in the lifecycle, i.e.:

```python
if getattr(self, "main_session", None) is not None:
    self.main_session.color_depth = pt_resolve_color_depth()
```

cmd2/pt_utils.py
```diff
@@ -75,6 +77,13 @@ def pt_filter_style(text: str | ANSI) -> str | ANSI:
     return text if isinstance(text, ANSI) else ANSI(text)
 
 
+def pt_resolve_color_depth() -> ColorDepth:
+    """Determine the prompt-toolkit ColorDepth based on NO_COLOR and ru.ALLOW_STYLE."""
+    if "NO_COLOR" in os.environ or ru.ALLOW_STYLE == ru.AllowStyle.NEVER:
```
tleonhardt: The `NO_COLOR` standard specifically states that colors should be prevented only when the variable is "present and not an empty string". The current implementation will incorrectly disable colors if NO_COLOR="" is passed. Using `os.environ.get()` correctly evaluates an empty string to False while remaining True for any other value, e.g.:

```python
if os.environ.get("NO_COLOR") or ru.ALLOW_STYLE == ru.AllowStyle.NEVER:
```

tleonhardt (changes requested):

cmd2/cmd2.py
kmvanbrunt: This is true for a few of our settables. They expect certain instance members to exist before than can safely be set.

1. `allow_style`
2. `always_prefix_settables`
3. `traceback_show_locals`
4. `traceback_width`

tleonhardt: We should either attempt to ruggedize their use like I was suggesting above or do a very good job of documenting that these are special and certain instance members need to exist before they can be safely set (likely by saying don't set them before calling parent class dunder init).

cmd2/pt_utils.py
kmvanbrunt: Done

tleonhardt (approved):

cmd2/cmd2.py
kmvanbrunt: I think it's best to leave the code alone. `__init__()` functions are meant to establish the baseline state for an instance. Setting a parent's instance variables before calling `super().__init__()` doesn't seem safe since those values  could be overwritten by `__init__()`.

PR #1710, opened 2026-07-09
Added ability to use Rich renderables as alert messages.

kmvanbrunt:


tleonhardt (approved):

tleonhardt:
I really like the way the `async_printing.py` example looks now.

PR #1718, opened 2026-07-16
Fix with_annotated decorator so using ArgumentBlock works with groups

### LINKED ISSUE #1714: [with_annotated]  Using `ArgumentBlock` does not works  with groups
# Description

`ArgumentBlock` typed arguments used with `with_annotated` do not work well with argument groups.

```python
@dataclass
class Block(ArgumentBlock):
    A: ...
    B: ...


class App(Cmd):
    @with_annotated(groups=(Group("Block", ...), ...))
    def do_with_block(self, block_argument: Block) -> None:
        ...
```

# Expected Behavior

The argument group should be created, and all arguments defined in `Block` should be listed in it (see the help message).

# Actual Behavior

An argument group is created, but it does not list any arguments. Defining `Group("A", "B", ...)` is rejected and causes an error.

# Possible Solution

I see at least three possible approaches:

1. List all arguments defined in the `ArgumentBlock` in the corresponding argument groups when using the block to create the group.
2. Allow arguments defined in the `ArgumentBlock` to be referenced when defining argument groups.
3. Add syntax to define default argument groups inside the `ArgumentBlock`.

tleonhardt:
## Cause
When an `ArgumentBlock` parameter is passed to `@with_annotated(groups=...)`, the decorator resolves the argument spec based on the function's parameter names (e.g. `block_argument`). However, when parsing was eventually performed, `block_argument` had been completely expanded into its respective argument fields (e.g. `A`, `B`) which didn't possess the `block_argument` name directly, causing them to fall through as ungrouped arguments rather than being mapped to the user-supplied argument groups.

## Fix
1. Modified `_ArgparseArgument` to store the name of the block parameter it was expanded from (`block_param_name`)
2. Updated `_expand_dataclass_block` to supply `block_param_name` when instantiating expanded arguments
3. Updated `build_parser_from_function` to properly fall back to looking up targets via the `block_param_name` if an exact name match isn't found in `target_for`
4. Updated `_link_mutex_group_membership` to map `block_param_names` to a list of its respective sub-fields, enabling `mutually_exclusive_groups` to also enforce constraints across all expanded block fields correctly
5. Added tests (`test_group_with_argument_block`, `test_mutex_group_with_argument_block`) to explicitly test this functionality in `tests/test_annotated.py`

Closes #1714

### DIFF
--- a/cmd2/annotated.py
+++ b/cmd2/annotated.py
@@ -175,9 +175,11 @@ dataclass constructor at reconstruction time, so a ``default_factory`` yields a 
 argument.  A field whose type is itself a block is not expanded (no recursion) and raises the
 unsupported-type error.  Because fields expand flat, every field name must be unique across the command's
 own parameters and every other block's fields -- a collision (which would put two argparse actions on one
-namespace dest) raises ``TypeError`` when the parser is built.  Block fields cannot participate in
-``groups=`` / ``mutually_exclusive_groups=`` (those reference the function's own parameter names, which are
-validated without resolving type hints).  A block must be the *bare* annotation of a regular parameter:
+namespace dest) raises ``TypeError`` when the parser is built.  Block fields can participate in
+``groups=`` / ``mutually_exclusive_groups=``: a group names arguments, and because a block expands to one
+argument per field, you name its *field* names.  Since resolving those field names needs the block's
+type hints, this membership is checked when the parser is built rather than at decoration time.  A block
+must be the *bare* annotation of a regular parameter:
 wrapping it in ``Annotated``/``Optional``/a union, or using it as ``*args`` / ``**kwargs``, raises
 ``TypeError`` (to use a dataclass as a single value instead, make it a plain ``@dataclass`` with an
 ``Argument(converter=...)``).  An ``ArgumentBlock`` subclass that is not a ``@dataclass`` has no fields and
--- a/cmd2/annotated.py
+++ b/cmd2/annotated.py
@@ -1126,6 +1126,7 @@ class _ArgparseArgument:
         kind: inspect._ParameterKind,
         is_base_command: bool,
         is_block_field: bool = False,
+        block_param_name: str | None = None,
         dest_override: str | None = None,
     ) -> None:
         # signature-derived inputs (never emitted):
@@ -1134,6 +1135,7 @@ class _ArgparseArgument:
         self.has_default = has_default
         self.param_default = param_default  # the function's own default, not the argparse `default` slot
         self.is_block_field = is_block_field
+        self.block_param_name = block_param_name
         self.dest_override = dest_override
         self.is_kw_only = is_kw_only
         self.is_variadic = is_variadic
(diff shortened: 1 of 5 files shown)


tleonhardt:
@KelvinChung2000 I'd greatly appreciate your review as this is my first attempt at modifying the code in `annotated.py`.

@neoniobium Since you reported the issue and understand it, I'd also very much appreciate your review to verify that this fixes the underlying problem.

cmd2/annotated.py
KelvinChung2000, line 2425:
```suggestion
            _reject_field_shadowing_block_param(inherited, name, func.__qualname__)
            if any(name in spec.members for spec in mutually_exclusive_groups or ()):
                raise TypeError("Cannot have BlockArgument for mutally exclusive groups as this is ambigious....")
```

Something similar, or can be extracted into a check function similar to `_reject_field_shadowing_block_param` for and have a more detailed error message

KelvinChung2000, line 2437:
```suggestion
        if _is_argument_block(block_hint, param):
            if any(name in spec.members for spec in mutually_exclusive_groups or ()):
                raise TypeError("Cannot have BlockArgument for mutally exclusive groups as this is ambigious....")
```

Similar to above

KelvinChung2000 (changes requested):
While this is working for normal groups, as they are just normal basiclly an "or" collection of all the entry. But applying the same logic to me does not really works on mutually exclusive group and I would suggest rejecting this use case. 

 I would prefer a more explict API, something like 

```py
@dataclass
class Block(ArgumentBlock):
    A: ...
    B: ...


class App(Cmd):
    @with_annotated(groups=(Group("A", "B"), ...))
    def do_with_block(self, block_argument: Block) -> None:
        ...
```
Which is considered a typo today.  But it will be clear in all cases. So the implementation for the above will be more like lifting/changing all the field checks from checking the parameter directly to, when it is an `ArgumentBlock`, checking whether it is part of the field instead. Since `ArgumentBlock` is a flat expansion in the parser, all the existing logic should be working already.

tleonhardt:
@KelvinChung2000 If you have a better idea for how to handle this, I'm happy to cancel this PR and let you put up another one. Or if you want, feel free to take over this branch and modify as you see fit.

neoniobium:
I checked out the PR and did some brief testing. The issue I raise is fully addressed with it. I happily do a technical review later. However I would prefer if it be decided weather this PR will be closed in favor a a different approach before I do that.

Edit: Ok, I see that @KelvinChung2000 commit. I assume this PR to be this to be the one that will address the issues.

neoniobium (dismissed):
I have tested the PR in private project. Works as intended. The implementation look clean.

You might also want to update the example? However, do not have a strong opinion on weather this is too much for an example especially since the current example is also rather long.

cmd2/annotated.py
tleonhardt: I think this comment that says `Block fields cannot participate in groups / mutually_exclusive_groups` needs to be updated.

tleonhardt:
PR LGTM.  I called out one place where I think the file header comment for `annotated.py` needs to be updated to ensure it is fully accurate with these changes.

PR #1727, opened 2026-08-05
Fixed bug where already sorted choices_provider results were being re-sorted.

kmvanbrunt:


cmd2/argparse_completer.py
```diff
@@ -785,13 +785,15 @@ def _complete_arg(
                     cmd_set,
                 )
                 all_choices = list(choices_provider(*args, **kwargs))
+                sort = False  # choices_provider results are already sorted, so don't re-sort them
```
tleonhardt, line 788: Hardcoding `sort = False` for all `choices_provider` results appears to assume that every custom choices provider returns a `Choices` object. However, if a developer's custom choices provider returns a raw `list` or `tuple` (such as strings or `CompletionItems`), it will bypass `Choices.__post_init__` sorting and will now skip sorting entirely because `sort = False` is passed to `basic_complete`. Furthermore, if raw strings are returned (as in the `modular_commandsets.py` example), `choice.text` in the subsequent list comprehension (`[choice for choice in all_choices if choice.text ...]`) will raise an `AttributeError: 'str' object has no attribute 'text'`, crashing tab completion.

Consider ruggedizing along the lines of:
```python
provider_results = choices_provider(*args, **kwargs)
from .completion import Choices
if isinstance(provider_results, Choices):
    all_choices = list(provider_results)
    sort = not provider_results.is_sorted
else:
    all_choices = [
        choice if isinstance(choice, CompletionItem) else CompletionItem(choice)
        for choice in provider_results
    ]
    sort = True
...
```

tleonhardt (changes requested):

cmd2/argparse_completer.py
kmvanbrunt: A `choices_provider` can only return a `Choices` object. I fixed the `modular_commandsets.py` example.

tleonhardt: Ok, if what I was worried about can't happen, all good

tleonhardt (approved):

PR #1732, opened 2026-08-21
Enabled Ctrl-Z suspension at the prompt

kmvanbrunt:


tleonhardt (approved):
Test locally on master and verified `Ctrl-Z` did not work and tested on this branch and verified it did.

PR #1733, opened 2026-08-22
Added support for bracketed paste

kmvanbrunt:
Added bracketed paste support so multiple pasted commands execute sequentially and multiline commands continue as expected.

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -78,6 +78,8 @@ from prompt_toolkit.formatted_text import ANSI, AnyFormattedText
 from prompt_toolkit.history import InMemoryHistory
 from prompt_toolkit.input import DummyInput, create_input
 from prompt_toolkit.key_binding import KeyBindings
+from prompt_toolkit.key_binding.key_processor import KeyPress, KeyPressEvent
+from prompt_toolkit.keys import Keys
 from prompt_toolkit.output import DummyOutput, create_output
 from prompt_toolkit.patch_stdout import patch_stdout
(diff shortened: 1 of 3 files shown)


cmd2/cmd2.py
```diff
@@ -740,6 +742,49 @@ def _should_continue_multiline(self) -> bool:
...
+            data = event.data.replace("\r\n", "\n").replace("\r", "\n")
+            if "\n" not in data:
+                event.current_buffer.insert_text(data)
+                return
+
+            key_presses = []
+            for i, line in enumerate(data.split("\n")):
+                if i > 0:
+                    key_presses.append(KeyPress(Keys.ControlM, "\r"))
+                key_presses.extend(KeyPress(ch, ch) for ch in line)
```
tleonhardt: Constructing and feeding a `KeyPress(ch, ch)` object for every single character in a pasted line is CPU and memory-intensive for large pasted commands. Additionally, feeding individual characters can trigger other active key bindings mapped to single characters (e.g., custom hotkeys or vi-mode commands) in the middle of a paste. 

Using `Keys.Any` with the entire line as its data allows prompt-toolkit's default wildcard handler to insert the whole string as a single, optimized operation. This bypasses character-by-character key binding evaluation entirely, which is safer and significantly more efficient.

```python
if line:
    key_presses.append(KeyPress(Keys.Any, line))
```

tleonhardt (changes requested):

cmd2/cmd2.py

tleonhardt (approved):

PR #1735, opened 2026-08-25
Fix broken example and improve some documentation

tleonhardt:
Fix broken example and improve some documentation

- Fix broken `cmd_as_argument.py` example whose `speak` command was crashing because `str` has no `copy` method since strings are immutable in Python.
- Update the `startup_commands.md` documentation to make it more discoverable what the recommended solution is for mixing commands at invocation with a custom argparse parser.
- Update the `os.md` documentation to make sure it is still accurate

### DIFF
--- a/examples/cmd_as_argument.py
+++ b/examples/cmd_as_argument.py
@@ -50,7 +50,7 @@ class CmdLineApp(cmd2.Cmd):
         words = []
         for w in args.words:
             word = w.strip()
-            if args.piglatin:
+            if args.piglatin and word:
                 word = f"{word[1:]}{word[0]}ay"
             if args.shout:
                 word = word.upper()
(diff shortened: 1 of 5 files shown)


examples/cmd_as_argument.py
```diff
@@ -1,13 +1,17 @@
 #!/usr/bin/env python
 """A sample application for cmd2.
 
-This example has additional code in main() that shows how to accept a command from
-the command line at invocation:
+This example reliese on the `allow_cli_args` init parameter being `True` by default which allows passing commands on the
```
kmvanbrunt: Fix spelling of **relies**.

kmvanbrunt (dismissed):

kmvanbrunt (approved):

PR #1737, opened 2026-08-26
Fix ty unresolved-attribute warnings and add mypy back

tleonhardt:
Add `mypy` back  in addition to `ty` for type checking. 

Also fixes `ty` unresolved-attribute warnings and stop globally ignoring them within `ty.toml`.

Major changes:
- Add `mypy` back so now both `ty` and `mypy` are used for type checking - each are better at some things
- Convert `BoundCommandFunc` and `UnboundCommandFunc` TypeAliases in `types.py` to Protocol classes for stricter type checking
- Added `_CommandFunc` type alias in annotated.py based on `BoundCommandFunc` and `UnboundCommandFunc`
- Used `getattr` and/or `cast()` to help resolve some type errors in cmd2.py

Minor changes:
- Added type ignore for `ty:unresolved-attribute` to a number of places we were already ignoring `attr-defined` for mypy (problem of different name for same type of check)

### DIFF
--- a/cmd2/annotated.py
+++ b/cmd2/annotated.py
@@ -284,13 +284,11 @@ from typing import (
     NamedTuple,
     ParamSpec,
     Protocol,
-    TypeAlias,
     TypedDict,
     TypeGuard,
     TypeVar,
     Union,
     Unpack,
-    cast,
     get_args,
     get_origin,
     get_type_hints,
@@ -311,11 +309,8 @@ from .decorators import _parse_positionals
 from .exceptions import Cmd2ArgparseError
 from .rich_utils import Cmd2HelpFormatter, HelpContent
 from .types import (
-    BoundCommandFunc,
-    CmdOrSet,
     CmdOrSetT,
     UnboundChoicesProvider,
-    UnboundCommandFunc,
     UnboundCompleter,
 )
 
@@ -323,7 +318,18 @@ from .types import (
 _NargsValue = int | str | tuple[int] | tuple[int, int] | tuple[int, float]
 
 
-_CommandFunc: TypeAlias = BoundCommandFunc | UnboundCommandFunc[CmdOrSet, Any]
+class _CommandFunc(Protocol):
+    """The command-function interface required by the annotation machinery.
+
+    The parser builders only depend on a command's name metadata and its
+    boolean command result; they never depend on whether the function is
+    bound or unbound, nor on the types of its parsed arguments.
+    """
+
+    __name__: str
+    __qualname__: str
+
+    def __call__(self, *args: Any, **kwargs: Any) -> bool | None: ...
 
 
 class Cmd2ParserKwargs(TypedDict, total=False):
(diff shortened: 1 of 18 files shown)


tleonhardt:
@KelvinChung2000 Could you please review the changes in `annotated.py`?  There this is mostly replacing every `Callable[..., Any]` type hint with a type alias: `_CommandFunc: TypeAlias = BoundCommandFunc | UnboundCommandFunc[CmdOrSetT, Any]` so it is more in line with type hints elsewhere in the codebase.

@kmvanbrunt Could you please review changes outside of `annotated.py`? For this there are two main changes:
1. Account for `ty` calling what `mypy` called `attr-defined`, `unresolved-attribute` instead.
2. Converted `BoundCommandFunc` and `UnboundCommandFunc` from simple type alias to Protocol classes which do a better job of type hinting that certain attributes have to exist (duck typing)

KelvinChung2000 (dismissed):

cmd2/types.py
```diff
@@ -65,13 +66,36 @@
...
-UnboundCommandFunc: TypeAlias = Callable[Concatenate[CmdOrSetT, P], bool | None]
+class UnboundCommandFunc(Protocol[CmdOrSetT, P]):
+    """Protocol for an unbound command function."""
+
+    __name__: str
+    __qualname__: str
+
+    def __call__(self, __self: CmdOrSetT, /, *args: P.args, **kwargs: P.kwargs) -> bool | None:
+        """Invoke the unbound command function with its command instance."""
+        ...
```
kmvanbrunt: You can remove this `...`

kmvanbrunt (changes requested):

cmd2/annotated.py
```diff
@@ -2957,7 +2964,7 @@ def with_annotated(
 
 
 def with_annotated(
-    func: Callable[..., Any] | None = None,
+    func: _CommandFunc | None = None,
```
kmvanbrunt: This doesn't match the type for `func` in the first overload of `with_annotated`.

kmvanbrunt (changes requested):

cmd2/annotated.py
tleonhardt: Can you say why it doesn't match?

kmvanbrunt: Look at the first overload on line 2940.

```Python
@overload
def with_annotated(func: Callable[_CommandParams, _CommandReturn]):
```

The type for `func` is more permissive here than the `_CommandFunc` accepted in the actual implementation.

`mypy` catches this error but I'm not sure why `ty` is fine with it.

cmd2/types.py
tleonhardt: Done

tleonhardt:
@KelvinChung2000 @kmvanbrunt Ok, I think this is ready for another review.

I ended up adding `mypy` back to our type checking in addition to `ty` because each of them are stricter about checking some things than the other.

cmd2/annotated.py
```diff
@@ -2902,35 +2908,52 @@ def _build_subcommand_handler(
...
+_DecoratorParams = ParamSpec("_DecoratorParams")
+
+
+class _AnnotatedCommand(Protocol[_CommandParams]):
+    """A callable command method with the metadata used by ``with_annotated``."""
+
+    __name__: str
+    __qualname__: str
+
+    def __call__(self, *args: _CommandParams.args, **kwargs: _CommandParams.kwargs) -> bool | None: ...
```
KelvinChung2000: Could move this to before `_build_subcommand_handler`, then `_build_subcommand_handler` will not need quotes on the type.

tleonhardt: Good suggestion.  Done.

PR #1745, opened 2026-09-05
Consolidated toolbar during command execution

### LINKED ISSUE #1743: Bottom bar should persist during command execution

tleonhardt:
## Summary

Today `enable_bottom_toolbar=True` only shows a toolbar while `prompt-toolkit` is waiting for input — it vanishes the moment a command starts running. This branch makes the toolbar persistent: it stays visible and keeps refreshing while commands execute, with command output scrolling above it.

Because a persistent toolbar owns the bottom of the terminal, anything else that wants the terminal has to coordinate with it. That drives the two other pieces of this branch: an embedded pager that shares the toolbar's application, and a suspension mechanism for the places that genuinely need the raw terminal (nested prompts, shell commands, Python/IPython shells, external pagers, and pipes to interactive processes).

### Persistent toolbar (`cmd2/command_toolbar.py`, new)

- `CommandToolbar` reuses `main_session.app` and its existing toolbar container rather than standing up a second `prompt-toolkit` application, so content callback, styling, and refresh interval are shared with the prompt.
- Commands still run on the main thread; the command loop drives the toolbar UI from a background thread.
- Stable `ToolbarStream` wrappers route Python-level and piped subprocess output through `prompt-toolkit`'s output proxy so writes land above the toolbar. Their identities stay consistent across cmd2 redirection and toolbar suspension.
- An `RLock` shared by the toolbar and its streams serializes writes against the proxy swap during pause/resume.
- Input stays coordinated: cursor position reports are handled, type-ahead is saved for the next prompt, `Ctrl-C` is forwarded to the process group (`os.killpg` on POSIX, `_thread.interrupt_main()` on Windows), and `Ctrl-Z` suspend-to-background is honored where supported.
- If the display thread dies on its own, `_app_exited()` detaches the streams so output goes back to the real terminal and the error is reported, instead of writing into a dead proxy.

### Embedded pager (`cmd2/pager.py`, new)

- `ppaged()` uses a built-in pager that shares the toolbar's application, so the toolbar keeps rendering while you scroll. Supports colors, wrapped lines, chopped lines with horizontal scrolling (including on Windows, which the external `more` never did), search (`/`, `n`/`N`), `g`/`G`, and `q`.
- New `use_builtin_pager` attribute, defaulting to `enable_bottom_toolbar`. Set it to `False` to keep using the configured external `pager`/`pager_chop`; an external pager takes over the terminal, so that path still suspends the toolbar.

### Suspension (`cmd2/cmd2.py`)

- New public `Cmd.suspend_bottom_toolbar()` context manager for application code that calls `input()`, drives its own terminal UI, or starts a subprocess that inherits the terminal.
- Applied internally via a `@suspend_toolbar` decorator on `_read_raw_input()`, `select()`, `do_shell()`, `_run_python()`, `do_ipy()`, and `_run_cmdfinalization_hooks()`.
- Pipes to interactive targets: the pipe process now inherits the real terminal through the new `pipe_target()` helper, and the toolbar suspends for the life of the pipe (tracked on `RedirectionSavedState.toolbar_suspension`, released in `_restore_output()`). Streams with no file descriptor still fall back to `PIPE`. Previously `alias | fzf`-style commands produced no output at all.
- The toolbar is started around `onecmd_plus_hooks()` from the interactive loop and around startup commands. It is not started for non-interactive input or for `onecmd_plus_hooks()` calls made outside the loop.
- A failing `toolbar.stop()` no longer disables the toolbar for the rest of the session — `_command_toolbar = None` moved into its own `finally`.

## Notes and trade-offs

- `get_bottom_toolbar()` now runs in a background UI thread during command execution. Applications that read state a command mutates should guard it with a lock. This is documented.
- The embedded pager is deliberately basic (scroll + search). Applications wanting full `less` behavior opt out with `use_builtin_pager = False`.
- One guarded dependency on `PromptSession`'s internal layout structure, verified against prompt-toolkit 3.0.52 and 3.0.53; it raises a clear `RuntimeError` if the structure moves.
- `Ctrl-\` (`SIGQUIT`) is inert while the toolbar is up because raw mode clears `ISIG`. This matches existing behavior at the main prompt, so it is documented rather than changed.

## Tests

- New `tests/test_command_toolbar.py` with 31 tests covering startup/shutdown, output routing, suspension paths, pipe handling, `Ctrl-C` forwarding, the pager, and the EOF shutdown path.
- Includes a fix for Windows CI: the fixture now binds the app session to the test pipe and recording output, so `patch_stdout()` cannot lazily construct a real `Win32Output` and raise `NoConsoleScreenBufferError` on runners with no console screen buffer.

## Docs

- `docs/features/prompt.md`: persistent behavior, refresh, the pager controls, `suspend_bottom_toolbar()`, and threading guidance.
- `docs/features/os.md` and `docs/features/initialization.md`: `use_builtin_pager` and the external-pager opt-out.
- `examples/getting_started.py`: new `work [seconds]` command demonstrating the toolbar clock updating while a command prints output.
- CHANGELOG entry under 4.3.0.

## NOTES

This is one possible architecture and approach to solving this problem.

I did a bunch of manual testing on both macOS and Linux and everything looks good there, but I need someone else to help test on Windows.

## TODO

- [x] Improve documentation
- [x] Investigate slow tests on Windows - they take about 7 times as long on Windows relative to Linux. Before this change Windows tests only took about 40% longer than Linux. The culprit appears to be slow Windows tests in the `test_command_toolbar.py` file.
- [x] Investigate stuck test on Windows when specifically using Python 3.11. One of the tests in `test_command_toolbar.py` gets stuck indefinitely
- [x] Manual testing on Windows - got a colleague to verify everything behaves as expected using the `getting_started.py` app on Windows from this branch

Closes #1743


tleonhardt:
This PR is a better attempt at making the bottom bar persistent. It works in nearly all situations, but involved some complexities like needing to introduce code to support a custom pager.  It works and has been validated to do what it is intended to do. However it continues to have an annoying issue with flicker of the bottom bar that has always existed.

While we could review it and merge it in, I'm holding off on that for now as I explore options for truly eliminating the flicker to provide a better user experience. As I've learned that is not an easy problem to solve.

PR #1748, opened 2026-09-06
DO NOT MERGE: continuous prompt-toolkit application - does not fix the toolbar flicker (findings recorded)

tleonhardt:
> **Status: does not achieve its goal. Do not merge.**
>
> This branch was built to remove the bottom toolbar flicker. Manual testing on macOS and Linux shows the flicker is unchanged, and measurement shows this branch is marginally **worse** on the reported trigger. It is kept open to record the findings, which redirect the work. CI is red on the free-threaded jobs; that is a real defect in this branch and is documented below rather than fixed, since the branch is parked.

## Scope

This PR targets `main`, so its diff contains the whole persistent bottom toolbar and embedded pager feature — the commits on `consolidated_toolbar` — plus 10 further commits that change how the toolbar's prompt-toolkit application is driven. Those 10, from `341a1eb5` onward, are the delta; everything before is `consolidated_toolbar` unchanged. `main` has since been merged in, which brought the right prompt fix (#1749) along with it.

## What this branch does

Keeps **one** prompt-toolkit `Application` running for the whole of `cmdloop()`, swapping `layout`/`key_bindings` on it instead of stopping and restarting between the prompt and each command — the technique `CommandToolbar.page()` already uses for the pager.

That part works: `tests/test_toolbar_flicker.py` asserts that reading a line neither stops nor restarts the application, and reverting the prompt routing makes it fail with `the application was restarted after the prompt`.

It just does not fix the flicker.

## What actually causes the flicker

The original diagnosis was that `PromptSession` excludes the bottom toolbar from every run's final frame via the `~is_done` term in its filter, so ending a run between the prompt and the command erases the toolbar. That is true, and this branch removes that boundary. **It is not what the user sees.**

A blink is a full-screen **erase followed by a repaint**. Counting `\x1b[J` (erase to end of screen) in the window right after pressing Enter on an empty line, driving `examples/getting_started.py` through a real pty — reproducible byte-for-byte across runs:

| | erase-down sequences per Enter |
| --- | --- |
| `consolidated_toolbar` | **3.0** |
| this branch | **3.8** |

This branch is worse. The regression is `_echo_accepted`: a continuous application never renders the `is_done` frame that used to commit the prompt line to scrollback, so `read_line()` writes the line itself through `StdoutProxy`. Every proxy write goes through `run_in_terminal`, which calls `renderer.erase()` and repaints. One erase was removed and another added.

The deeper point is that the erase and the repaint arrive in the **same pty write batch** — there is no time gap in our process to shrink. The terminal receives "wipe this region, now redraw it" and sometimes paints the intermediate state. That is why the blink is intermittent and terminal-dependent, and why reducing the number of application *runs* was never going to fix it. Printing anything above a live prompt-toolkit UI requires that erase; it is inherent to `run_in_terminal`.

Earlier measurements in this PR that suggested success counted how many times the *prompt line* was written (2x to 1x). That was real, but it measured the wrong thing.

### One further boundary, also insufficient

`_run_cmdfinalization_hooks` is decorated `@command_toolbar.suspend_toolbar` — it calls `termios.tcsetattr(..., self._initial_termios_settings)` to repair a terminal corrupted by binary output, which would clobber prompt-toolkit's raw mode. It costs a full stop/restart of the display on **every** command, on this branch and on `consolidated_toolbar` alike. Removing it takes pause/resume cycles per command to zero and drops erases from 6 to 4 per command, but not to zero, so it does not fix the blink either.

## The one confirmed win — extracted and merged

Manual testing found a genuine improvement: the right prompt used to be redrawn into the scrollback beside every accepted line; here it stays only on the live prompt line.

`PromptSession` renders the right prompt as a `Float` with **no `~is_done` filter** (`shortcuts/prompt.py:157-162`), unlike the bottom toolbar container, so it is included in the committed final frame. This branch fixed it only incidentally, because `_echo_accepted` writes `prompt + text` and never the right prompt — the same code responsible for the erase regression above.

That fix needed none of this architecture. It was extracted as a standalone change against `main` and merged in **#1749**.

### Prompt-mode Ctrl-C routing is unverified, and its test passes for the wrong reason

`test_read_line_ctrl_c_raises_without_stopping_the_app` sends `"partial\x03"` **before** `read_line()` switches the application into prompt mode. The keys are therefore handled with command-mode semantics, where the `interrupt` binding calls `os.killpg(os.getpgrp(), SIGINT)`. Spying on `_fail_line`, the method the prompt-mode Ctrl-C binding calls:

```
_fail_line calls: []
```

It is never called. The `KeyboardInterrupt` that makes the test pass comes from that real `SIGINT`, not from the binding added in `3ed28313`. **The claim that this branch routes prompt-mode interrupts without ending the run is therefore unverified.** Ctrl-D is genuinely exercised (`_fail_line calls (ctrl-d): ['EOFError']`), because `\x04` is swallowed by `save_key` and replayed into prompt mode rather than raising a signal.

### Free-threaded CI failures: a hang, not a crash

The 3.14t jobs on Linux and macOS fail after roughly 40-60 seconds with `The runner has received a shutdown signal` followed by `The operation was canceled`, and no failing test. `fail-fast` is `false`, so these are not collateral from another job, and `main`'s 3.14t jobs pass consistently on all three platforms.

Reproduced locally on a real free-threaded build (`uv sync --python 3.14t`, `sys._is_gil_enabled()` is `False`): the suite **hangs** in `test_read_line_ctrl_c_raises_without_stopping_the_app`, at a `Thread.join()`. Run in isolation it hangs inside `pytest.raises`, meaning `read_line()` never returns. The cause is the same as above — under the GIL the `SIGINT` happens to land while the main thread is inside `read_line()`, so it surfaces as the expected `KeyboardInterrupt`; without the GIL the timing shifts, the signal lands outside that window, and `read_line()` blocks forever. A hung suite with threads still running is consistent with the runner being killed.

Windows 3.14t passes, because the Windows `interrupt` binding uses `_thread.interrupt_main()` rather than `os.killpg`.

### The design gap behind both

In the real `cmdloop`, between a command finishing and `read_line()` entering prompt mode, the application is still in command mode. A Ctrl-C typed in that window sends `SIGINT` to the whole process group instead of just aborting the line. This branch widened that window, because prompt mode is now entered by an explicit layout swap rather than by starting a fresh application that already has the prompt's own bindings. Any future attempt at a continuous application has to close this window.

### Data races

`read_line()` has two unsynchronized cross-thread accesses, independent of the above: `self._line` is written by the command thread (`command_toolbar.py:457`) and read by the UI thread (`:430`, `:439`); `session.default_buffer.accept_handler` is set by the UI thread inside `enter()` (`:463`) and restored by the command thread in `finally` (`:493`). The second is the more dangerous — if a queued Enter is processed after the stock handler has been restored, that handler calls `app.exit()` and terminates the continuous application.

## Where the work should go instead

To stop the blink, the toolbar row has to stay **out of the erased region** — a terminal scroll region (DECSTBM) is how persistent status bars normally avoid exactly this. That is a change at the output layer, not the application layer, and is not what this branch does.


tleonhardt:
This branch and PR represent a failed attempt at preventing flicker in the bottom toolbar by having a single `prompt-toolkit` application running that we didn't start/stop, but where we dynamically swapped out layouts.  There were numerous problems and it wasn't workable.

PR #1751, opened 2026-09-07
Reserved-row toolbar: Stage 1 primitives, dependency contracts, and terminal qualification

tleonhardt:
Stage 1 qualification work for the reserved-row bottom toolbar. **Ready — Windows qualification is complete and Stage 1 is closed.** This existed to run the
Windows CI matrix and give the tester a known commit to check out; both are done. It
changes no runtime behaviour: nothing here is wired into toolbar rendering.

Targets the long-lived `reserved_row_toolbar` integration branch.

## Background

Pressing Enter makes the bottom toolbar blink, rated *objectionable* by a user on three
real-terminal runs. The cause is now understood: prompt-toolkit only rewrites rows that
differ from `Renderer._last_screen`, and `renderer.erase()` discards that baseline, forcing
a full repaint. Three paths trigger it per command. **The repaint that restores the toolbar
is the same event that erases it**, so removing the erases cannot work — proven by positive
control.

The design (specified separately) keeps the toolbar on a DECSTBM-reserved bottom row that
ordinary output never scrolls through, with the renderer's destructive operations bounded to
the region above it.

## What's here

**`cmd2/scroll_region.py` + tests** — DECSTBM sequences and a margin-bounded replacement for
`erase_down`. `ED` ignores the scroll margins and would destroy the reserved row; `DL` is
bounded by them and needs no knowledge of the cursor's row, which matters because `Output`
does not track one. 15 tests, 100% line coverage, five mutations each killing their test.

**`tests/test_prompt_toolkit_contracts.py`** — locks the non-public prompt-toolkit details
the design depends on, so a dependency upgrade fails loudly here instead of silently
breaking terminal rendering: the CPR available-height arithmetic, both destructive erase
calls going through the `Output` interface, and the renderer's diff baseline.

Four of these run only on Windows, and **CI is the only place they can run**:
`Windows10_Output` is a *registered virtual subclass* rather than a real one, so capability
checks must not rely on inheritance; geometry is delegated natively, so adapting
`get_size()` alone is insufficient; its inner VT output carries a zero-size stub; and legacy
`Win32Output.erase_down` is a separate implementation a VT replacement never reaches.

**`scripts/toolbar_inventory.py`** — temporary diagnostic for the Windows tester,
included so the tool and the code under test come from the same commit. It refuses to run
unless stdout and stdin are both terminals, because redirection selects `PlainTextOutput`
and would record the wrong backend. Its probe always restores margins and attributes. To be
removed once Windows qualification is done.

## Evidence so far

POSIX primitives were qualified through real tmux 3.7c at 12, 24 and 40 rows — 15 probes
including controls, all agreeing with expectation. Highlights: a region anchored at row 1
preserves scrollback and protects the reserved row; anchoring at row 2 orphans the first
line; a one-usable-row region is not honoured at all; `DL` at a nonzero column destroys the
line prefix, so column-zero normalization must be enforced rather than assumed.

## Qualification results — Stage 1 CLOSED

Every environment that gated Stage 1 has now been run and passes. Console modes were restored
exactly on both Windows runs.

| # | Environment | Backend | Geometry | Result |
| --- | --- | --- | --- | --- |
| — | iTerm2 3.6.11, bare (macOS 26.6.2) | `Vt100_Output` / `Vt100Input` | 66x230, region `1;65r` | **PASS** |
| W1 | Windows Terminal (Win 11 26200) | `Windows10_Output` / `Win32Input` | 21x92, region `1;20r` | **PASS** |
| W2 | Git Bash / mintty (`MSYSTEM=CLANGARM64`) | `Windows10_Output` / `Win32Input` | 25x156, region `1;24r` | **PASS** |

- [x] `TOOLBARMARKER` stayed on the bottom row for the whole scroll
- [x] It survived the bounded erase, with the region above it cleared
- [x] Scrollback complete and in order
- [x] Scrollback free of `TOOLBARMARKER`
- [x] Shell prompt behaves normally after exit
- [x] Cursor visible and correctly placed after exit
- [x] Console mode before and after match (`0x0007`, VT processing on, both Windows runs)

The bare iTerm2 run also closes a gap the POSIX evidence could not: Stage 1's primitives were
collected through tmux, and tmux owns its own scrollback and scroll regions, so those passes
qualified tmux rather than any host terminal.

### Split delegation, now observed rather than assumed

Both Windows runs show one object with three different owners:

```
delegate::erase_down  ->  prompt_toolkit.output.vt100.Vt100_Output
delegate::get_size    ->  prompt_toolkit.output.win32.Win32Output
delegate::flush       ->  prompt_toolkit.output.windows10.Windows10_Output
```

Geometry comes from the native Win32 side while rendering goes out over VT — the arrangement the
contract tests exist to pin. `rows_below_cursor_position` is implemented there (1 and 18) where
Vt100 raises `NotImplementedError`.

### Superseded first round

The earlier Windows attempts in this thread failed: no bar was ever visible and the scrollback
looked interleaved. That was the probe's own defect — it buffered the whole run and flushed once,
painting it as a single repaint, so nothing scrolled and the bar was never drawn where it could be
seen. Fixed by per-line flush plus a `--delay` pause. Those first artifacts qualify nothing.

## Known limitation: the toolbar is not pinned during scrollback

Seen on every terminal, POSIX and Windows: scrolling up moves the bar out of view rather than
holding it at the bottom. This is expected and not fixable within the design, and it is now
documented in `docs/features/prompt.md`.

An application cannot paint into the terminal's scrollback view — it is not notified of the scroll,
cannot query it, and has no escape sequence addressing it. A toolbar that stayed visible over
history would have to have been written into the scrollback buffer, which is exactly the hard
failure the "free of `TOOLBARMARKER`" check exists to catch. The property that makes this design
correct and the property being asked for are the same property in opposite directions. `less` and
`vim` pin a status line for a whole session only by using the alternate screen, where there is no
scrollback at all.

## What this still does not establish

- Nothing here is wired into toolbar rendering; there is no user-visible behaviour change.
- Backing buffer taller than the viewport was **not** exercised — `viewport_differs_from_buffer`
  was `False` on both Windows runs, so that hazard is untested rather than ruled out.
- Conhost (W3/W4) has not been re-run on the fixed probe. It is record-only; the compatibility
  fallback covers it.
- First placement over a populated terminal, and resize under a live application, need the
  integrated implementation and belong to later stages.

`scripts/toolbar_inventory.py` was to be removed once Windows qualification finished. It is kept
for now because the remaining stages still need real-terminal runs; it should be removed before
this feature reaches `main`.


tleonhardt:
@kmvanbrunt @bambu Can one of your please follow the instructions towards the bottom of the description under the `Windows checklist for the manual pass` section?  

The purpose of this branch and PR is to do some proof-of-concept testing for a new architectural approach where the bottom row (or 2 or whatever) is truly reserved just for the status bar and is completely separate from the normal app.

The concept appears viable for both Linux and Mac, but needs some manual testing validation on Windows. The basic thing to run is:
```sh
uv run python scripts/windows_toolbar_inventory.py
```

Then there is a checklist of things to look for. I sent you both an email with a status document explaining a little more.

tleonhardt:
Windows Terminal Run.

Here is what I was in the terminal to start:
```sh
PS C:\Users\toddleonhardt\src\python\cmd2> uv run .\scripts\windows_toolbar_inventory.py

=== Environment inventory ===
  captured_at                            2026-09-09T16:00:06.582220+00:00
  platform                               Windows-11-10.0.26200-SP0
  os_release                             11
  python                                 3.14.7 (tags/v3.14.7:823f032, Aug  5 2026, 11:30:42) [MSC v.1944 64 bit (ARM64)]
  python_executable                      C:\Users\toddleonhardt\src\python\cmd2\.venv\Scripts\python.exe
  prompt_toolkit                         3.0.53
  TERM                                   None
  WT_SESSION                             d875e8dd-78e2-4fb5-8636-7a6f3fafe6ef
  MSYSTEM                                None
  TERM_PROGRAM                           None
  output_class                           prompt_toolkit.output.windows10.Windows10_Output
  input_class                            prompt_toolkit.input.win32.Win32Input
  stdout_isatty                          True
  stdin_isatty                           True
  viewport_rows                          49
  viewport_columns                       90
  win32_buffer_size                      {'X': 91, 'Y': 49}
  win32_window                           {'Left': 0, 'Top': 0, 'Right': 90, 'Bottom': 48}
  win32_cursor                           {'X': 0, 'Y': 1}
  viewport_differs_from_buffer           False
  rows_below_cursor_position             48
  delegate::erase_down                   prompt_toolkit.output.vt100.Vt100_Output
  delegate::erase_screen                 prompt_toolkit.output.vt100.Vt100_Output
  delegate::get_size                     prompt_toolkit.output.win32.Win32Output
  delegate::get_rows_below_cursor_position prompt_toolkit.output.win32.Win32Output
  delegate::flush                        prompt_toolkit.output.windows10.Windows10_Output
  console_mode_before                    0x0007 (VIRTUAL_TERMINAL_PROCESSING=on, WRAP_AT_EOL=on)

=== DECSTBM probe (visual confirmation required) ===
  About to reserve the bottom row, scroll past it, then reset.
  WATCH THE BOTTOM ROW. Press Enter when ready.
```

After I pressed enter I saw some stuff and TOOLBARMARKER was at the bottom.  But at no point did I see a bottom bar at all and saw (with a lot of scrollback):
```sh
probe line 0001ell
probe line 0002icrosoft Corporation. All rights reserved.
probe line 0003
probe line 0004dleonhardt> cd .\src\python\cmd2\
probe line 0005dleonhardt\src\python\cmd2> uv run .\scripts\windows_toolbar_inventory.py
probe line 0006
probe line 0007 inventory ===
probe line 0008                          2026-09-09T16:05:07.412343+00:00
probe line 0009                          Windows-11-10.0.26200-SP0
probe line 0010                          11
probe line 0011                          3.14.7 (tags/v3.14.7:823f032, Aug  5 2026, 11:30:4
probe line 001264 bit (ARM64)]
probe line 0013able                      C:\Users\toddleonhardt\src\python\cmd2\.venv\Scrip
probe line 0014
probe line 0015t                         3.0.53
probe line 0016                          None
probe line 0017                          7ebb8bf0-daaa-4ac2-8891-0c284bfe590c
probe line 0018                          None
probe line 0019                          None
probe line 0020                          prompt_toolkit.output.windows10.Windows10_Output
probe line 0021                          prompt_toolkit.input.win32.Win32Input
probe line 0022                          True
probe line 0023                          True
probe line 0024                          49
probe line 0025mns                       90
probe line 0026size                      {'X': 91, 'Y': 49}
probe line 0027                          {'Left': 0, 'Top': 0, 'Right': 90, 'Bottom': 48}
probe line 0028                          {'X': 0, 'Y': 5}
probe line 0029ers_from_buffer           False
probe line 0030rsor_position             44
probe line 0031se_down                   prompt_toolkit.output.vt100.Vt100_Output
probe line 0032se_screen                 prompt_toolkit.output.vt100.Vt100_Output
probe line 0033_size                     prompt_toolkit.output.win32.Win32Output
probe line 0034_rows_below_cursor_position prompt_toolkit.output.win32.Win32Output
probe line 0035sh                        prompt_toolkit.output.windows10.Windows10_Output
probe line 0036before                    0x0007 (VIRTUAL_TERMINAL_PROCESSING=on, WRAP_AT_EO
probe line 0037
probe line 0038
probe line 0039be (visual confirmation required) ===
probe line 0040rve the bottom row, scroll past it, then reset.
probe line 0041TOM ROW. Press Enter when ready.
probe line 0042
probe line 0043
probe line 0044
probe line 0045
probe line 0046
probe line 0047
probe line 0048
probe line 0049
probe line 0050
probe line 0051
probe line 0052
probe line 0053
probe line 0054
probe line 0055
probe line 0056
probe line 0057
probe line 0058
probe line 0059
probe line 0060
probe line 0061
probe line 0062
probe line 0063
probe line 0064
probe line 0065
probe line 0066
probe line 0067
probe line 0068
probe line 0069
probe line 0070
probe line 0071
probe line 0072
probe line 0073
probe line 0074
probe line 0075
probe line 0076
probe line 0077
probe line 0078
probe line 0079
probe line 0080
probe line 0081
probe line 0082
probe line 0083
probe line 0084
probe line 0085
probe line 0086
probe line 0087
probe line 0088
probe line 0089
probe line 0090
probe line 0091
probe line 0092
probe line 0093
probe line 0094
probe line 0095
probe line 0096
probe line 0097
probe line 0098
probe line 0099
probe line 0100
=== Report ===1
  console mode before : 0x0007 (VIRTUAL_TERMINAL_PROCESSING=on, WRAP_AT_EOL=on)
  console mode after  : 0x0007 (VIRTUAL_TERMINAL_PROCESSING=on, WRAP_AT_EOL=on)
  modes match         : True
probe line 0105
  Answer in the result template:
    1. Did TOOLBARMARKER stay on the bottom row for the whole scroll?
    2. Is the scrollback complete (probe line 0001 upward) and in order?
    3. Does TOOLBARMARKER appear anywhere in the scrollback? (it must not)
    4. After this program exits, does the shell prompt behave normally?
probe line 0111
  machine-readable record written to: C:\Users\toddleonhardt\src\python\cmd2\scripts\windows_toolbar_inventory_TODDLEONHARCC3D.json
  Attach that file to the result template.
PS C:\Users\toddleonhardt\src\python\cmd2>
probe line 0116
probe line 0117
probe line 0118
probe line 0119
probe line 0120
probe line 0121
probe line 0122
probe line 0123
probe line 0124
probe line 0125
probe line 0126
probe line 0127
probe line 0128
probe line 0129
probe line 0130
probe line 0131
probe line 0132
probe line 0133
probe line 0134
probe line 0135
probe line 0136
probe line 0137
probe line 0138
probe line 0139
probe line 0140
probe line 0141
probe line 0142
probe line 0143
probe line 0144
probe line 0145
probe line 0146

TOOLBARMARKER
```
At no point did I ever see an actual bottom bar. Scrollback looked funky. The application had exited and dropped me back to the shell but in the middle of the screen. The shell prompt behaved normally afterwards and the cursor was visible after exit.

I don't understand how this was supposed to work, but I can't imagine this is what "right" looks like.

Here is that JSON output:
[windows_toolbar_inventory_TODDLEONHARCC3D.json](https://github.com/user-attachments/files/32016578/windows_toolbar_inventory_TODDLEONHARCC3D.json)

tleonhardt:
Git Bash shell.

Experience essentially identical to Windows terminal.

Here is the JSON output:

[windows_toolbar_inventory_TODDLEONHARCC3D.json](https://github.com/user-attachments/files/32016841/windows_toolbar_inventory_TODDLEONHARCC3D.json)

tleonhardt:
Conhost beheavior nearly identical to others except I was able to scroll where TOOLBARMARKER did not remain on the bottom.  For both Windows Terminal and Git Bash, TOOLBARMARKER always remained on the bottom.

JSON output attached
[windows_toolbar_inventory_TODDLEONHARCC3D.json](https://github.com/user-attachments/files/32016953/windows_toolbar_inventory_TODDLEONHARCC3D.json)

tleonhardt:
The recent commit fixed a couple bugs in the test script, generalized it to work on all platforms, and added a short pause inbetween each line of output to make it easier to observe the scrolling behavior.

tleonhardt:
Updated Windows Terminal Run. Everything looked fine.

[toolbar_inventory_TODDLEONHARCC3D.json](https://github.com/user-attachments/files/32018109/toolbar_inventory_TODDLEONHARCC3D.json)

tleonhardt:
Re-run of Git Bash terminal run.  Everything looked fine.

[toolbar_inventory_TODDLEONHARCC3D.json](https://github.com/user-attachments/files/32018212/toolbar_inventory_TODDLEONHARCC3D.json)

PR #1753, opened 2026-09-07
Write redirected and piped output as UTF-8, and stop the suite inheriting colour settings

tleonhardt:
Two independent fixes, both already reviewed and merged to `consolidated_toolbar` in #1752.
Ported here because `main` is the next patch release and both problems are present on it
today. The third commit from #1752 — a `CommandToolbar` race — is **not** included, because
that module does not exist on `main`.

## 1. Redirection and piping fail on current Windows  *(user-visible data loss)*

Command output is rendered by Rich and contains box-drawing characters, but redirection
targets and pipes are opened with the locale's encoding. On Windows that is the ANSI code
page, and **no `cp125x` code page can represent them**, so redirecting raises
`UnicodeEncodeError` and the user is left with an empty file plus advice to set
`PYTHONIOENCODING`.

Measured on `main` at `ca0ddbe7`, running `help > out.txt` under a non-UTF-8 default encoding:

```
file size: 1 bytes
contains help text: False
```

After this change the same probe writes 153 bytes containing the help text.

### Who this affects

This is **not** limited to old systems. A fully updated Windows 11 with default settings is
affected. The character that breaks is `U+2500` (`─`), and encodability by code page is:

| Windows ANSI code page | Encodes `U+2500`? |
| --- | --- |
| cp1252 US / Western European | **No** |
| cp1250 Central European | **No** |
| cp1251 Cyrillic | **No** |
| cp1253–1258 Greek, Turkish, Hebrew, Arabic, Baltic, Vietnamese | **No** |
| cp932 / cp936 / cp949 / cp950 CJK | Yes |

Only the CJK double-byte code pages escape, because they happen to include box drawing.

By interpreter, `sys.flags.utf8_mode` is `0` on Python 3.11, 3.12, 3.13 and 3.14, and `1`
on 3.15 — PEP 686 makes UTF-8 mode the default there. So every Python version this project
currently supports below 3.15 is affected.

The other escapes are Windows' *"Beta: Use Unicode UTF-8 for worldwide language support"*
option, which is off by default, and setting `PYTHONUTF8=1` or `-X utf8`.

In short: a user on Windows 11, US locale, Python 3.13, typing `help > out.txt` in any cmd2
app loses their output. That is the common case, not an edge case.

`main`'s own tests do not catch this — nothing there redirects Rich-heavy output and reads it
back, so the suite passes while the product is broken. This PR adds tests that do.

## 2. The suite inherits the caller's colour environment

Rich and cmd2 consult several environment variables when deciding whether to emit styling,
and the suite inherits them. Exporting any one makes large numbers of unrelated tests fail
depending on who runs the suite — measured on `main`:

| Variable | Tests failed |
| --- | ---: |
| `NO_COLOR=1` | 15 |
| `FORCE_COLOR=1` | 53 |
| `TTY_COMPATIBLE=1` | 53 |

The failures look like product regressions, which makes them expensive to diagnose; this
cost a reviewer and me a full round trip before the cause was found. An autouse fixture now
neutralizes them, and a guard test fails if any reaches a test again. Tests that exercise
these variables still set them explicitly, because a test's own monkeypatching runs after
the fixture.

## Verification

Full suite under every environment that previously broke it:

```
baseline           1863 passed, 2 skipped
NO_COLOR=1         1863 passed, 2 skipped
FORCE_COLOR=1      1863 passed, 2 skipped
TTY_COMPATIBLE=1   1863 passed, 2 skipped
non-UTF8 locale    1863 passed, 2 skipped
```

`make check`, `make test` and `make docs-test` all pass. Reverting either encoding change
fails the new tests.

The pipe test runs its filter through `sys.executable` rather than `cat`, matching the
convention in the existing pipe tests — `cmd.exe` has no `cat`, and this fix exists for
Windows. Note that CI would not have caught that: GitHub's Windows runners have Git for
Windows' utilities on `PATH`.

## Note for merging

These commits also exist on `consolidated_toolbar` via #1752, so the same change appears in
both histories. The content is identical, so the eventual merge resolves cleanly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 ## 4.2.4 (TBD)
 
 - Bug Fixes
+    - Fixed output redirection and piping failing on systems whose default encoding is not UTF-8,
+      such as a Windows console using a legacy code page. Command output is rendered by Rich and
+      routinely contains non-ASCII, so redirecting it raised `UnicodeEncodeError` and left an empty
+      file behind. Redirection targets and pipes now use UTF-8 explicitly
     - Fixed the right prompt being redrawn beside every accepted command line in the scrollback.
       prompt-toolkit includes it in the final frame of each prompt, which is the frame left on the
       terminal; it is now hidden there, as the bottom toolbar already was, and stays on the live
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,11 @@
 ## 4.2.4 (TBD)
 
 - Bug Fixes
-    - Fixed output redirection and piping failing on systems whose default encoding is not UTF-8,
-      such as a Windows console using a legacy code page. Command output is rendered by Rich and
-      routinely contains non-ASCII, so redirecting it raised `UnicodeEncodeError` and left an empty
-      file behind. Redirection targets and pipes now use UTF-8 explicitly
+    - Fixed output redirection and piping raising `UnicodeEncodeError` and leaving an empty file
+      behind. Command output is rendered by Rich and contains box-drawing characters, which no
+      `cp125x` Windows ANSI code page can represent, so redirecting or piping it failed. This
+      affects current Windows 11 with default settings on Python 3.11 through 3.14, not only older
+      systems. Redirection targets and pipes now use UTF-8 explicitly
     - Fixed the right prompt being redrawn beside every accepted command line in the scrollback.
       prompt-toolkit includes it in the final frame of each prompt, which is the frame left on the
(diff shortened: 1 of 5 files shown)


tleonhardt:
@kmvanbrunt @bambu This is a bug fix to merge to `main` for fixing some edge cases on older Windows systems where `utf-8` isn't the default encoding. This manually sets `utf-8` as the encoding when piping or redirecting to ensure any Rich text styling is supported. Everything LGTM. But if either of you have time for a quick review, great.

PR #1758, opened 2026-09-10
Stage 3: ordinary lifecycle and mode selection

tleonhardt:
# Flicker-free bottom bar overview and status 

This is a WIP proof-of-concept for a bottom bar which is truly flicker free. This is a LOT more work than I initially thought it would be because this is something which should be fixed within `prompt-toolkit`, but this is a [known bug](https://github.com/prompt-toolkit/python-prompt-toolkit/issues/547)  which was reported in 2017 and the sole author of `prompt-toolkit` thinks would be "too tricky to get right".

| Stage | Focus | Main deliverables | Completion criteria |
|---|---|---|---|
| **0** | Scope and design | Define supported environments, architecture, recovery rules, and implementation plan. | Scope and design reviewed; environment inventory prepared. |
| **1** | Terminal capabilities | Verify scrolling margins, bounded erases, cursor reports, and backend support. | Required automated and real-terminal probes pass. |
| **2a** | Geometry and output adapter | Reserve bottom rows, expose usable terminal dimensions, and manage acquisition and release. | Static toolbar survives scrolling and clearing without losing output. |
| **2b** | Painting and serialization | Add toolbar painter, terminal locking, renderer prepare/commit, and recovery. | Concurrent writes and rendering remain ordered; failure recovery preserves terminal state. |
| **3** | Main application integration | Connect mode selection, prompts, command output, input ownership, and lifecycle cleanup.| Ordinary commands preserve a stable toolbar, correct prompts, and complete output history. |
| **4** | Advanced interactions | Integrate nested prompts, pagers, external programs, resizing, and suspend/resume signals. | Ownership transitions and cleanup work across the interaction matrix. |
| **5** | Release qualification | Complete cross-platform manual testing, dependency qualification, documentation, and rollback guidance. | All advertised configurations pass; any default-mode promotion requires a separate decision. |

This PR implements Stage 3. Stages up through 2b were implemented previously on the `reserved_row_toolbar` integration branch for this feature. Stages 4 and 5 remain. This is the first stage at which the reserved toolbar is now integrated within cmd2 applications in a meaningful way that can be manually tested.  You can manually test this feature using:

```sh
uv run examples/getting_started.py
```

My goal at this point is a working proof-of-concept that clearly demonstrates a flicker-free toolbar is possible. Later, I plan to try to clean things up and simplify the code.

## What this adds

| Module | Role |
| --- | --- |
| `cmd2/toolbar_mode.py` | Mode validation and prerequisite qualification |
| `cmd2/reserved_toolbar.py` | Owns the reservation, bindings, painter and bridge for one loop |
| `cmd2/managed_output.py` | Writes command output inside the terminal transaction |

plus the integration in `cmd2.py`, `command_toolbar.py` and the bridge.

## The decisions worth reviewing

**`auto` falls back silently; `reserved` refuses to start.** A backend that looks close enough
is still a guess, and a wrong guess corrupts the screen the user is working in. But a caller
who asked for reserved rendering and silently got legacy has been given the behaviour they
ruled out, and would learn it from a flickering toolbar rather than an error.

**Two kinds of pause, and each site says which it means.** Stopping the renderer and the input
reader is one thing; handing the physical terminal to something else is another. Command
finalization needs only the first — the toolbar has to still be there at the next prompt — and
so does the main prompt, which is what the reservation exists for. Shell commands, editors,
embedded interpreters and external pagers get the rows back for the duration.

**Command output goes straight to the terminal, inside the transaction.** The stdout proxy
exists to put output above a toolbar that scrolls; a reserved toolbar does not, and proxying
anyway would put back exactly the flicker being removed. The bridge is told inside the same
transaction — the contract Stage 2b could not enforce from where it sat.

**Renders go through the bridge.** Prompt-toolkit rendered whenever its event loop decided to,
so a frame could interleave with command output or a paint. Renders, erases, clears and cursor
reports are intercepted on the renderer instance and restored on teardown.

## Gate evidence

The observer gate runs against a real cmd2 application now, not a harness that drives the
adapter directly.

| Rows | Commands | Disappearances | Bar present after | Committed history |
| --- | --- | --- | --- | --- |
| 12 | 22 | 0 | 22/22 | 20/20 |
| 24 | 34 | 0 | 34/34 | 32/32 |
| 40 | 50 | 0 | 50/50 | 48/48 |

The same application in legacy mode is the control: **34 disappearances across 34 commands, 33
of them erase-caused**, with the toolbar still present after every window and history complete.
Legacy remains functional and flickers; reserved does not. Same observer, same workload, only
the mode differs.

Mode selection verified from the raw captures: `reserved` and `auto` install `ESC[1;23r` and
reset exactly once; `legacy` installs no margins at all. The dynamic gate passes at all three
geometries, with 0 of 35 cursor reports coming from inside the band — the first live evidence
that the band-row validation is exercised by a real terminal.

Repeated loops, finalization failure and shutdown are covered at the `cmdloop()` level.

## Verification

Full suite 2514 passed / 6 skipped; `make check` and `make docs-test` clean. 100% line coverage
on the five new/rewritten modules, 99% on `command_toolbar.py` and `cmd2.py`. Ten review rounds
and twenty-two findings were resolved on this branch, each with a failing test first and each
fix mutation-checked.

Documentation, the changelog, the real-terminal matrix and any change to the default belong to
Stage 5.


tleonhardt:
@kmvanbrunt @bambu @theagilehacker 
I would appreciate help with manual testing from as many people and on as many operating systems, terminals, and versions of Python as possible. I have done some limited testing on Mac, Linux, and windows - everything works great.

If you run checkout the `stage3-lifecycle-integration` branch and run this:

```sh
uv run examples/getting_started.py
```

Then you will have an application using the proof-of-concept flicker-free bottom toolbar. If you want to understand the difference, checkout the `main` branch and run the same application and hit `<Enter>` a bunch of times.

kmvanbrunt:
@tleonhardt If the bottom toolbar contains a newline, the text after the newline does not display.

tleonhardt:
@kmvanbrunt Great find.  I'll work on a fix for that.  Currently it is just a single reserved line for now. Once I'm happy with this, I intend extend it where between 0 to 3 lines can be reserved. But if the text goes beyond the one line, it should strip anything extra in a known and documented way.

tleonhardt:
@kmvanbrunt Can you please pull the latest changes and test again in the case where the toolbar contains a newline?

kmvanbrunt:
@tleonhardt I see the `…` for truncated text.

tleonhardt:
@kmvanbrunt Yeah that `…` was my intention to be a visual indicator for truncated text. I was debating if we should strip newlines off the end first or not.  Are you happy with the way it works. Or do you have a suggestion for something better?

tleonhardt:
@kmvanbrunt The Windows bug has been fixed and that platform is now good for manual testing as well.

There is a known bug with terminal resizing that I'm working on.

bambu:
Not sure if its the due to the terminal resizing bug that you mentioned but the bottom toolbar does not refresh properly after suspending and restoring

tleonhardt:
@bambu Can you please elaborate a little on what you mean by "suspending and restoring"?

Are you talking about:
- Linux process backgrounding and re-foregrounding (Ctrl-Z, etc)?
- Tmux session/tab switching?
- VM suspend and restore?
- Something else?

bambu:
I meant suspending the process via `ctrl-z`/`fg`. I sent you a video of it as well.

tleonhardt:
@bambu Thanks.  I just verified that bug with manual testing.  The first pass at the terminal resize fix didn't catch it. That is basically another signal we'll need to make sure we catch probably so we can redraw.

tleonhardt:
@bambu The suspend/resume workflow support is planned to be added in Stage 4 for `Advanced interactions`.  This is Stage 3 for basic integration of the flicker-free reserved bottom bar.

tleonhardt:
I'm going to merge this and start on the next stage, Stage 4. I think manual testing should feel a lot better after that because nearly all of the edge cases should be taken care of.

PR #1761, opened 2026-09-12
Stage 4: integrate reserved toolbar handoffs, job control, and resize recovery

tleonhardt:
## Summary

Complete Stage 4 of the reserved-row toolbar work: transfer terminal and input ownership correctly across nested prompts, paging, external applications, suspend/resume, and resize. This PR targets `reserved_row_toolbar`, which already contains Stage 3; it retains the current one-row reservation and opt-in mode defaults.

### What changed

- **Managed nested input:** `read_input()`, `read_secret()`, and `select()` borrow the existing reservation after the command reader stops. Temporary applications receive their own bridge, retain the current toolbar content/styles, and restore original bindings when they finish, are interrupted, or fall back. This avoids a second input reader and preserves completion, secret masking, and typeahead.
- **Embedded pager ownership and loading:** release the reservation before the first pager layout measures available space, expose full physical geometry, and avoid subtracting the toolbar height twice when deciding whether output fits. Keep the existing main-screen toolbar visible while preparing a large pager's first frame; clear it when that frame commits or when intervening output/external ownership requires it. Restore the main-screen reservation after leaving the pager, including cleanup when exit-time erasure fails.
- **Nested and external handoffs:** make release/reacquisition idempotent across nested suspension and alternate-screen ownership. Reacquire from fresh geometry, invalidate stale prompt/output accounting, and restore compatibility bindings when acquisition fails. If both the guest and cleanup fail, preserve the guest's exception and retain the cleanup error for diagnostics.
- **POSIX job control:** release rows inside prompt-toolkit's cooked-mode/input-detached handoff before stopping the foreground job, then reacquire on resume. Suspending while a pager already owns the screen does not install a reservation inside its alternate buffer. Look up `SIGTSTP` conditionally so Windows type checking and unsupported-platform behavior remain valid.
- **Interactive pipelines:** terminal-attached POSIX pipelines now share cmd2's foreground process group. Previously `help -v | less`, Ctrl-Z, and `fg` could leave `less` running while cmd2 was stopped, allowing it to steal shell input. Avoid forwarding SIGINT back into cmd2's own group; retain isolated groups for nonterminal pipelines and the existing Windows process-group behavior.
- **Minimum-height recovery:** include the reservation generation in the painter's cached-band identity. A 3 → 2 → 3 resize can erase the old cells and reacquire identical coordinates; the first new paint must redraw the entire band rather than reuse that stale cache. Unchanged content remains a no-op within a stable generation.
- **Examples and regression coverage:** enable `py` in the getting-started example and add an exception-raising script for manual testing. Add nested-input, pager, cleanup, polling, partial-output, and real-PTY job-control regressions.

### Review focus

The implementation retains the existing alternating main/command application lifecycle and serialized-output architecture. Review ownership restoration and failure ordering, especially nested suspension, alternate-screen exit, native Windows geometry, and partial output. The `ChoiceInput._create_application()` integration and polling tests depend on prompt-toolkit internals; qualification remains at exactly 3.0.53. Broader dependency/backend qualification follows in Stage 8.

## Roadmap

The additional investigations requested after Stage 4 move the former Stage 5 release qualification to Stage 8.

| Stage | Work | Status / exit criterion |
| --- | --- | --- |
| 3 | Ordinary prompt/command lifecycle, mode selection, serialization, stable toolbar and history | Completed and merged into `reserved_row_toolbar` via #1758 |
| 4 | Nested prompts, pager/external handoffs, job control, resizing and cleanup | This PR; integration regressions and real-terminal transitions verified, with PR CI/review and final evidence closure still required |
| 5 | Spike: is the embedded pager still useful and worth maintaining? | Compare capabilities, startup latency, portability and maintenance cost; review a keep/simplify/replace/remove recommendation before changing behavior |
| 6 | Toolbar package organization, test naming, and simplification | Investigate an internal `cmd2/toolbar/` package; align focused tests with production modules; keep cross-module integration suites; separate mechanical moves from justified behavior-preserving refactors |
| 7 | Configurable 0–3 reserved toolbar lines | 0 disables; 1 preserves current behavior; 2/3 use explicit newlines for subsequent reserved rows. Clip each line at terminal width with visible truncation; verify geometry, fallback and every ownership transition. Review mode consolidation separately rather than assuming removal of legacy/forced controls |
| 8 | Final integrated release qualification | Qualify the final code across required terminals/backends and dependency versions; record history/flicker evidence, compatibility and rollback. Default promotion remains a separate maintainer decision |

## Validation

At head `37645b59` on Mac:

- `make test`: **2,667 passed, 6 skipped**.
- Ruff lint/format checks on the pulled fix, mypy, documentation build, and diff whitespace checks passed. Required full checks also passed before the preceding Stage 4 commits.
- An in-memory mutation removing the generation from the painter cache makes both new painter regressions fail.
- Pipeline regression: a real controlling PTY and interactive shell exercise repeated Ctrl-Z/fg, shell input, resizing, quit/interrupt, toolbar restoration and subsequent cmd2 input. The preceding commit passed 20 PTY cases across 10 runs; removing either process-group fix fails its regression.
- Earlier Stage 4 acceptance/dynamic harness captures passed at 12/24/40 rows. These are earlier snapshots, not a claim that the final PR head has already been recaptured.

Maintainer-reported manual results: nested prompts on Mac and Windows; embedded paging and external tool/interpreter recovery; Mac job-control edge cases; clean shell return; and the exact pipeline sequence in Mac iTerm2. The Windows 2 → 3-row issue was subsequently reported fixed and pulled in `37645b59`; reviewers should repeat the exact boundary check below and attach the native result. Mac review of that commit found no blocking issues. PR CI is additional evidence, not a substitute for native terminal verification.

## Manual reviewer checklist

Run from the repository root with `uv run python examples/stage4_manual.py`. `stage4_manual.py` is a super-set of the `getting_started.py` application which adds commands for testing nested prompts via `read_input`, `read_secret`, and `select`. The example uses AUTO with a live timestamp. Record the commit, OS, terminal/version, shell, Python and prompt-toolkit versions, actual selected backend/mode, and measured viewport dimensions. AUTO can fall back: a visible native toolbar alone does not prove reserved rendering. `scripts/toolbar_inventory.py` can help identify the environment; inspect its usage first because it includes interactive probes.

Perform applicable checks on Mac/Linux and Windows Terminal. POSIX Ctrl-Z/fg is not expected in PowerShell. External TUIs intentionally hide the reserved toolbar while they own the terminal; it must return when they exit. Exact physical row counts matter, especially at the minimum height.

| Check | How to exercise it | Expected result |
| --- | --- | --- |
| Ordinary prompts and commands | Press Enter repeatedly; run `echo hello` and `work 5`; edit input and use completion/history | Stable reserved band, live timestamp, usable prompt, no duplicated/lost input or output |
| Managed text input | In the optional test app below run `nested`; use Tab and Up, resize, then accept; repeat and press Ctrl-C or EOF | Band remains reserved; completion/history work; only the active prompt consumes input; main prompt recovers |
| Secret input and selection | Run `secret` with dummy text only; run `choose` and use arrows/Enter; repeat with cancellation and resize | Secret remains masked; selection works; toolbar and main input recover without stale bindings |
| Typeahead | At `nested`, quickly enter `alpha` followed by a newline and `echo after` followed by a newline | Accepted input and following command are each consumed once by the correct owner |
| Large embedded pager | `cat cmd2/cmd2.py`; use Space/PgDn, b/PgUp, `/` search, `n`, resize and `q` | No prolonged blank-toolbar interval while loading; pager uses full geometry with its native bar; main reserved band and prompt return on exit |
| Short pager output | `cat examples/scripts/raise_exception.py` in a normal-height terminal | Short output prints normally without an unnecessary full-screen pager |
| External pager | On POSIX run `help -v \| less`; resize inside it, navigate and quit with `q` | External pager owns the full terminal; toolbar returns at current dimensions with usable input and intact output |
| External editor | Use `edit` with a disposable file and a configured editor; resize and exit | Full terminal available to the editor; reservation restored afterward |
| Embedded Python/IPython and scripts | Run `py`/`ipy`, print text and exit normally; run a normal `run_pyscript` and `run_pyscript examples/scripts/raise_exception.py` | External interpreter ownership is intentional; success and exception paths restore toolbar/input. Record unavailable optional IPython explicitly |
| Nested public suspension | Optional app: `handoff`, then `handoff fail`; resize at the guest input and complete it | No early reacquisition at an inner exit; toolbar restored after outer exit, including failure |
| POSIX prompt/command suspend | At an idle or partially typed prompt, and during `work 10`, press Ctrl-Z; run a shell command, resize, `fg`; repeat | Shell has full scrolling/input while stopped; resume restores correct geometry and preserves editable input/output |
| POSIX pager suspend | Repeat Ctrl-Z, shell command, resize and `fg` inside the embedded pager; navigate and quit afterward | No reservation installed into a guest screen; pager remains usable and restores main-screen toolbar on exit |
| POSIX pipeline ownership regression | `help -v \| less`, Ctrl-Z; type a complete shell command such as `printf 'shell owns input\n'`; `fg`; `q`; then `echo after` in cmd2 | Shell receives all keystrokes; less and cmd2 stop/resume together; reserved bar and command input return. Repeat the cycle |
| Exact minimum-height recovery | Idle main prompt: measure and resize 24 → 2 → 3 → 4 → 24; repeat with partially typed input. Also start at 2 or 3 rows and grow | At 2 rows reservation releases; at 3 there are 2 usable rows plus a fully painted toolbar on row 3; 4+ remains correct. No Enter/Ctrl-L needed to repair it |
| Quiet and partial-output resize | Optional app: `quiet`, then `partial`; resize repeatedly through 2/3/4 rows during the pause | Automatic resize detection; full bar returns when eligible; `PARTIALEND` remains intact and output never lands in the toolbar band |
| Width and handoff recovery | Narrow/widen at main/nested prompts and in a pager; repeat the height-boundary sequence after returning from an external tool | Current width/viewport used, no stale band, prompt overlap, missing input or permanent fallback |
| Compatibility and disabled mode | In a disposable copy of the example change constructor mode AUTO to LEGACY, then OFF; repeat basic commands and paging. Test RESERVED on a qualified backend | Legacy remains usable; OFF has no toolbar; RESERVED really exercises the new path. Unsupported forced prerequisites report an error rather than pretending to qualify |
| History and final shell cleanup | After the transitions, scroll through history; quit, type a shell command and generate more than one screen of output; relaunch | No toolbar leakage or unexplained lost/duplicated command output; normal shell echo/cursor and full-height scrolling; fresh loop behaves normally |

### Optional commands for managed-input and handoff checks

Run `uv run python examples/stage4_manual.py`. These commands execute inside the normal command lifecycle. Calling `read_input()` from the external `py` shell would instead exercise an already-suspended handoff.

```python
import time
from getting_started import BasicApp


class ManualApp(BasicApp):
    def do_nested(self, arg):
        value = self.read_input(
            "Value> ", choices=["alpha", "beta"], history=["alpha", "beta"]
        )
        self.poutput(repr(value))

    def do_secret(self, arg):
        self.read_secret("Dummy secret> ")
        self.poutput("Secret accepted; value not displayed")

    def do_choose(self, arg):
        self.poutput(self.select(["alpha", "beta"], "Choose> "))

    def do_quiet(self, arg):
        time.sleep(10)

    def do_partial(self, arg):
        self.stdout.write("PARTIAL")
        self.stdout.flush()
        time.sleep(10)
        self.stdout.write("END\n")
        self.stdout.flush()

    def do_handoff(self, arg):
        with self.suspend_bottom_toolbar():
            with self.suspend_bottom_toolbar():
                input("Guest owns terminal; resize, then Enter> ")
            if arg.strip() == "fail":
                raise RuntimeError("Intentional guest failure")


if __name__ == "__main__":
    ManualApp().cmdloop()
```

Fault-injection cases such as failing pager erasure, failed reacquisition and reservation abandonment are covered by automated tests; reviewers should inspect those outcomes rather than make production callbacks fail ad hoc. Record PASS/FAIL/not available per manual row, with terminal details and screenshots/diagnostic traces for failures. The broader release terminal matrix and configurable multiple reserved rows remain later-stage work.

### DIFF
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -86,6 +86,7 @@ from prompt_toolkit.layout.containers import ConditionalContainer, FloatContaine
 from prompt_toolkit.output import DummyOutput, create_output
 from prompt_toolkit.patch_stdout import patch_stdout
 from prompt_toolkit.shortcuts import CompleteStyle, PromptSession, choice, set_title
+from prompt_toolkit.shortcuts.choice_input import ChoiceInput
 from prompt_toolkit.styles import DynamicStyle
 from rich.console import (
     Group,
@@ -2175,7 +2176,8 @@ class Cmd:
 
         Use this context manager around application-specific calls to ``input()``, other
         terminal UIs, or subprocesses that inherit the terminal. cmd2 automatically suspends
-        its toolbar for its own input prompts, external pagers, and shell commands.
+        its toolbar for external pagers and shell commands. Managed input prompts borrow
+        the reservation without releasing the rows.
 
         In reserved mode this also gives the reserved rows back, because a program that
         inherits the terminal knows nothing about a scroll region and would find its output
@@ -3790,9 +3792,9 @@ class Cmd:
         The command display is stopped either way, but only some prompts give the terminal
         away with it. The main prompt is the one the reservation exists for: it renders
         through the reserved output, and the toolbar has to still be there while the user is
-        typing -- that is what "stable across ordinary commands" means. Any other session is
-        an application prompt cmd2 has not bound to the reservation, so it gets the terminal
-        to itself, rows included.
+        typing. Managed nested prompts on the same terminal borrow that reservation with
+        their own bridge after the command input reader stops. Other terminals and prompts
+        inside an external handoff retain the exclusive-terminal path.
 
         :param prompt: the prompt text or a callable that returns the prompt.
         :param session: the PromptSession instance to use for reading.
@@ -3800,11 +3802,13 @@ class Cmd:
         :return: the stripped input string.
         :raises EOFError: if the input stream is closed or the user signals EOF (e.g., Ctrl+D)
         """
-        owns_the_reservation = session is self.main_session
+        reserved = self._reserved_toolbar
+        owns_the_reservation = session is self.main_session or (reserved is not None and reserved.can_manage(session))
         with self._quiesce_bottom_toolbar() if owns_the_reservation else self.suspend_bottom_toolbar():
-            reserved = self._reserved_toolbar
             if owns_the_reservation and reserved is not None and reserved.bridge is not None:
                 reserved.bridge.finish_command_output()
+                with reserved.prompt_session(session):
+                    return self._read_raw_input_now(prompt, session, **prompt_kwargs)
             return self._read_raw_input_now(prompt, session, **prompt_kwargs)
 
     def _read_raw_input_now(
@@ -3932,7 +3936,7 @@ class Cmd:
 
         temp_session: PromptSession[str] = PromptSession(
             auto_suggest=self.main_session.auto_suggest,
-            bottom_toolbar=self.get_bottom_toolbar if self.main_session.bottom_toolbar is not None else None,
+            bottom_toolbar=self.main_session.bottom_toolbar,
             color_depth=self.main_session.color_depth,
             complete_style=self.main_session.complete_style,
             complete_in_thread=self.main_session.complete_in_thread,
@@ -3961,7 +3965,7 @@ class Cmd:
         :raises Exception: any other exceptions raised by prompt()
         """
         temp_session: PromptSession[str] = PromptSession(
-            bottom_toolbar=self.get_bottom_toolbar if self.main_session.bottom_toolbar is not None else None,
+            bottom_toolbar=self.main_session.bottom_toolbar,
             color_depth=self.main_session.color_depth,
             enable_suspend=self.main_session.enable_suspend,
             input=self.main_session.input,
@@ -4937,7 +4941,7 @@ class Cmd:
         self.last_result = True
         return True
 
-    @command_toolbar.suspend_toolbar
+    @command_toolbar.quiesce_toolbar
     def select(self, opts: str | Iterable[str] | Iterable[tuple[Any, str | None]], prompt: str = "Your choice? ") -> Any:
         """Present a menu to the user.
 
@@ -4972,7 +4976,22 @@ class Cmd:
             try:
                 while True:
                     with create_app_session(input=self.main_session.input, output=self.main_session.output):
-                        result = choice(message=prompt, options=fulloptions)
+                        reserved = self._reserved_toolbar
+                        if reserved is not None and reserved.can_manage(self.main_session):
+                            if reserved.bridge is not None:
+                                reserved.bridge.finish_command_output()
+                            # ChoiceInput.prompt() constructs and runs in one step. Binding
+                            # its application first gives the initial frame reserved geometry.
+                            selection = ChoiceInput(
+                                message=prompt,
+                                options=fulloptions,
+                                style=self.main_session.style,
+                                enable_suspend=self.main_session.enable_suspend,
+                            )._create_application()
+                            with reserved.prompt_application(selection):
+                                result = selection.run()
+                        else:
+                            result = choice(message=prompt, options=fulloptions)
                     if result is not None:
                         return result
             except KeyboardInterrupt:
--- a/cmd2/cmd2.py
+++ b/cmd2/cmd2.py
@@ -2183,7 +2183,11 @@ class Cmd:
         inherits the terminal knows nothing about a scroll region and would find its output
         confined to rows it never asked for. The rows are taken again afterwards.
         """
-        with self._quiesce_bottom_toolbar():
+        reader = self._cur_pipe_proc_reader
+        with (
+            reader.borrow_terminal() if reader is not None else contextlib.nullcontext(),
+            self._quiesce_bottom_toolbar(),
+        ):
             reserved = self._reserved_toolbar
 
             if reserved is None:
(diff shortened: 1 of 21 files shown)


tleonhardt:
@kmvanbrunt @bambu The flicker-free bottom bar is now in a more mature state where it is more completely integrated with `cmd2`.  Compared to the version yesterday evening, `<Ctrl>+Z` process suspesion and `fg` resume work properly, terminal resizing should be robust, it works with nested prompts and various other edge cases bugs have been fixed.

I'll keep reviewing and testing. But if others have time for manual testing on other platforms use:
```sh
uv run examples/getting_started.py
```

The bottom bar should now be fully persistent unless you invoke a shell application which is a TUI and takes over the whole screen and it should remain flicker-free. Though if for some reason you shrink the terminal vertical height to a pathological level of only 2 rows, it will revert to a default `prompt-toolkit` toolbar; though when you grow the size it will return to the reserved toolbar.

Please report any bugs you find along with the following:
- Operating system
- Termainl program
- Python version
- Steps to reproduce

NOTE: This is not the "done" state. I'm anticipating a good deal of cleanup along with adding support for multiple reserved rows.

